PR #3676, opened 2024-03-05
Update DCT limit_to filters

pavlo-mk:
[AB#190204](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/190204): Data Collecting Type not available


backend/hct_mis_api/apps/core/validators.py
```diff
@@ -329,9 +329,9 @@ def validate_data_collecting_type(cls, *args: Any, **kwargs: Any) -> Optional[No
         program = kwargs.get("program")
         business_area = kwargs.get("business_area") or getattr(program, "business_area", None)
 
-        # validate program BA and DCT.limit_to
-        if data_collecting_type and business_area:
-            if business_area not in data_collecting_type.limit_to.all():
+        # validate program BA and DCT.available_for
+        if data_collecting_type and business_area and data_collecting_type.available_for.all().count() > 0:
```
MarekBiczysko, line 333: data_collecting_type.available_for.exists()

pkujawa (commented):
Actually, with the naming it should stay the same: `limit_to`.
Since you are changing that if `limit_to` is empty -> it should be applied to all BAs (because we are not limiting).
The issue was that Iwith the previous logic, it should be named `available for` - because if that was empty, we would treat it as if DCT was not available for any BA.

pavlo-mk:
@saxix what you think about Paulina's comment?

johniak (approved):

PR #3677, opened 2024-03-05
Program permissions migration

pavlo-mk:


wozniakpl (approved):

pkujawa (approved):

pkujawa (changes requested):
I think this migration already is applied in 0043_migration?

johniak (approved):

PR #3700, opened 2024-03-13
[193048] Import people via api

patryk-dabrowski:


backend/hct_mis_api/api/endpoints/rdi/upload_people.py
```diff
@@ -0,0 +1,169 @@
...
+    ImportedHousehold,
+    ImportedIndividual,
+    RegistrationDataImportDatahub,
+)
+
+
+class PeopleSerializer(serializers.ModelSerializer):
+    first_registration_date = serializers.DateTimeField(default=timezone.now)
+    last_registration_date = serializers.DateTimeField(default=timezone.now)
+    household = serializers.ReadOnlyField()
```
saxix, line 38: household can be removed. It is an internal design

backend/hct_mis_api/apps/household/migrations/0172_migration.py
```diff
@@ -0,0 +1,23 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('household', '0171_migration'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='household',
+            name='social_worker',
```
saxix, line 15: do not use `social_worker`, let add a type field with value "STANDARD" and "SINGLE".
Default value is "STANDARD" use "SINGLE" for any "fake" HH created.

backend/hct_mis_api/apps/household/migrations/0172_migration.py
```diff
@@ -0,0 +1,23 @@
...
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AddField(
```
saxix, line 18: this is not needed, we can get it from linked HH

saxix (changes requested):

backend/hct_mis_api/api/endpoints/rdi/upload_people.py
patryk-dabrowski: removed

backend/hct_mis_api/apps/household/migrations/0172_migration.py
patryk-dabrowski: changed

pavlo-mk (approved):

PR #3716, opened 2024-03-20
[188755] Payment Plan Managerial View

pkujawa:
[AB#188755](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/188755)


backend/hct_mis_api/apps/payment/api/serializers.py
```diff
@@ -0,0 +1,58 @@
...
+            "is_follow_up",
+            "follow_ups",
+            "program",
+            "program_id",
+            "last_approval_process_date",
+            "last_approval_process_by",
+        )
+
+    def get_last_approval_process_by(self, obj: PaymentPlan) -> Optional[str]:
+        return str(obj.last_approval_process_by) if obj.last_approval_process_by else None
```
KonradMarzec1991, line 52: Might be also: return getattr(obj, "last_approval_process_by", None)

backend/hct_mis_api/apps/payment/api/views.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @action(
+        detail=False,
+        methods=["post"],
+        url_path="bulk-action",
+        serializer_class=PaymentPlanBulkActionSerializer,
+    )
+    def bulk_action(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        serializer = self.get_serializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
```
KonradMarzec1991, line 89: Just serializer.is_valid(raise_exception=True) is enough

backend/hct_mis_api/apps/payment/api/views.py
```diff
@@ -0,0 +1,152 @@
...
+        old_payment_plan = copy_model_object(payment_plan)
+        if old_payment_plan.imported_file:
+            old_payment_plan.imported_file = copy_model_object(payment_plan.imported_file)
+        if old_payment_plan.export_file_entitlement:
+            old_payment_plan.export_file_entitlement = copy_model_object(payment_plan.export_file_entitlement)
+        if old_payment_plan.export_file_per_fsp:
+            old_payment_plan.export_file_per_fsp = copy_model_object(payment_plan.export_file_per_fsp)
+
+        payment_plan = PaymentPlanService(payment_plan).execute_update_status_action(
+            input_data=input_data, user=request.user  # type: ignore
```
KonradMarzec1991, line 135: Why it is ignored?

backend/tests/apps/payments/test_payment_plan_views.py
```diff
@@ -0,0 +1,318 @@
...
+    ApprovalFactory,
+    ApprovalProcessFactory,
+    PaymentPlanFactory,
+)
+from hct_mis_api.apps.payment.models import Approval, PaymentPlan
+from hct_mis_api.apps.program.fixtures import ProgramFactory
+
+
+class PaymentPlanTestMixin:
+    def set_up(self, api_client: Callable, afghanistan: BusinessAreaFactory, id_to_base64: Callable) -> None:
```
KonradMarzec1991, line 26: Why not override setUp?

frontend/src/api/api.ts
```diff
@@ -0,0 +1,43 @@
...
+    const query = new URLSearchParams(params).toString();
+    const response = await fetch(`${this.baseURL}${url}?${query}`, {
+      headers: this.headers,
+    });
+
+    if (!response.ok) {
+      throw new Error(`Error fetching data from ${url}`);
+    }
+
+    return response.json();
```
KonradMarzec1991, line 16: Fetch method returns a promise, so you have to wait for the JSON - return await response.json();

backend/tests/apps/payments/test_payment_plan_views.py
pkujawa: There is no setUp, I am using pytest

backend/hct_mis_api/apps/payment/api/views.py
pkujawa: Because `request.user` returns `Union[AbstractBaseUser, AnonymousUser]` so either I ignore typing or I would have to change typing, considering also that `AnonymousUser` will never reach this code so either add `if` just for typing or have confusing typing with `AnonymousUser` inside

backend/hct_mis_api/apps/payment/api/serializers.py
pkujawa: Attribute `last_approval_process_by` will always be there, I need to take `str` and check if `last_approval_process_by` is not `None`

backend/tests/fixtures/api_client.py
```diff
@@ -0,0 +1,23 @@
+from typing import Callable, Optional
+
+import pytest
+from rest_framework.test import APIClient
+
+from hct_mis_api.apps.account.fixtures import UserFactory
+
+
+@pytest.fixture()
+def unauthenticated_api_client() -> APIClient:
```
KonradMarzec1991, line 10: This name is misleading - it just api client. Where it is used?

backend/hct_mis_api/apps/payment/api/serializers.py
KonradMarzec1991: If that's the case - it is better for “Ask forgiveness, not permission”

backend/tests/apps/payments/test_payment_plan_views.py
KonradMarzec1991: Old habits die hard

backend/tests/fixtures/common.py
```diff
@@ -0,0 +1,30 @@
+import base64
+from typing import Any, Callable
+
+import pytest
+
+from hct_mis_api.apps.account.fixtures import BusinessAreaFactory
+
+
+@pytest.fixture()
+def afghanistan(db: Any) -> BusinessAreaFactory:
```
KonradMarzec1991, line 10: Where db is used?

pkujawa: It is needed to access the database
[documentation](https://pytest-django.readthedocs.io/en/latest/helpers.html#db)

backend/tests/fixtures/api_client.py
pkujawa: I used `api_client` name for authenticated client, since it will be used in majority of cases and better for this to be `api_client` than `authenticated_api_client`

backend/hct_mis_api/apps/payment/api/views.py
```diff
@@ -0,0 +1,152 @@
...
+            new_object=payment_plan,
+        )
+
+    def _get_action_permission(self, action_name: str) -> Optional[str]:
+        action_to_permissions_map = {
+            PaymentPlan.Action.APPROVE.name: Permissions.PM_ACCEPTANCE_PROCESS_APPROVE.name,
+            PaymentPlan.Action.AUTHORIZE.name: Permissions.PM_ACCEPTANCE_PROCESS_AUTHORIZE.name,
+            PaymentPlan.Action.REVIEW.name: Permissions.PM_ACCEPTANCE_PROCESS_FINANCIAL_REVIEW.name,
+        }
+        return action_to_permissions_map.get(action_name)
```
MarekBiczysko, line 152: return action_to_permissions_map.get(action_name, None)

or limit choices in BulkActionSerializer only to app/auth/rev

backend/hct_mis_api/apps/payment/models.py
```diff
@@ -1016,6 +1016,43 @@ def get_program(self) -> "Program":
         # TODO will update after add feature with 'program_cycle' and migrate all data
         return self.program_cycle.program if self.program_cycle else self.program
 
+    def _get_last_approval_process_data(self) -> dict:
+        approval_process = hasattr(self, "approval_process") and self.approval_process.first()
+        if approval_process:
+            if self.status == PaymentPlan.Status.IN_APPROVAL:
+                return {
+                    "modified_date": approval_process.sent_for_approval_date,
```
MarekBiczysko, line 1024: minor thing, but this is a good place for dataclass usage instead of redefining same dict in all places 

{ "modified_date": x,"modified_by": x }


also access attribute would be not hardcoded

 return self._get_last_approval_process_data().modified_date
 
 instead of
  return self._get_last_approval_process_data()["modified_date"]

backend/tests/fixtures/common.py
KonradMarzec1991: ok, clear

KonradMarzec1991 (approved):
BE ok

frontend/src/api/api.ts
mmaciekk: yeah but.
The response.json() method itself returns a Promise that resolves with the result of parsing the body text as JSON. Therefore, it's not necessary to use await again when returning response.json(). The await keyword is already used before fetch to wait for the Promise that fetch returns. When response.json() is returned, it will return a Promise that will be awaited where the get or post method is called.

KonradMarzec1991 (approved):

patryk-dabrowski (approved):

mmaciekk (approved):
Good job

patryk-dabrowski (approved):

mmaciekk (approved):
very good code

PR #3722, opened 2024-03-21
[192988] Fix permissions

patryk-dabrowski:
[AB#192988](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/192988)

### DIFF
--- a/backend/hct_mis_api/apps/account/models.py
+++ b/backend/hct_mis_api/apps/account/models.py
@@ -99,11 +99,14 @@ class PartnerPermission:
         return instance
 
     def set_roles(self, business_area_id: str, roles: List[str]) -> None:
+        business_area_id = str(business_area_id)
         permissions = self._permissions.get(business_area_id, BusinessAreaPartnerPermission(business_area_id))
         permissions.roles = roles
         self._permissions[business_area_id] = permissions
 
     def set_program_areas(self, business_area_id: str, program_id: str, areas_ids: List[str]) -> None:
+        business_area_id = str(business_area_id)
+        program_id = str(program_id)
         permissions = self._permissions.get(business_area_id, BusinessAreaPartnerPermission(business_area_id))
         permissions.programs[program_id] = areas_ids
         self._permissions[business_area_id] = permissions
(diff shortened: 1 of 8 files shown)


pavlo-mk (approved):

wozniakpl (approved):

backend/hct_mis_api/apps/account/models.py
```diff
@@ -379,30 +382,25 @@ def cached_has_user_roles_for_business_area_and_permission(
...
             permission=permission,
         )
 
         if self.partner.is_unicef:
-            return has_program_access and (
-                has_user_roles or permission in DEFAULT_PERMISSIONS_LIST_FOR_IS_UNICEF_PARTNER
-            )
-        else:
-            return has_program_access and (has_user_roles or has_partner_roles)
+            return has_user_roles
```
pkujawa, line 391:
```
if self.partner.is_unicef:
    has_program_access = True
    has_partner_roles = permission in DEFAULT_PERMISSIONS_LIST_FOR_IS_UNICEF_PARTNER
else:
    has_program_access = str(program_id) in self.get_partner_programs_areas_dict(business_area_id=business_area.pk)
    has_partner_roles = self.cached_has_partner_roles_for_business_area_and_permission(
            business_area=business_area,
            permission=permission,
        )
```
?
Permission can still originate from partner UNICEF

patryk-dabrowski: UNICEF partner still has access to all programs, but user permissions determine to which business area and which features, in previous implementations in some cases, UNICEF partner had access to all business areas

pkujawa: But is it supposed to no longer grant any permissions now? Because right now users under UNICEF partner will not inherit any permissions.
Shouldn't UNICEF partner grant the permissions from the default list for the BAs that user has access to? So if user has access to Afg, but with role that grants only some 1 basic permission, he should have all permissions from UNICEF partner inside this BA, I think.

PR #3750, opened 2024-03-29
Selenium - RDI happy path

szymon-kellton:


pavlo-mk (dismissed):

backend/selenium_tests/page_object/programme_population/households.py
```diff
@@ -0,0 +1,125 @@
+from page_object.base_components import BaseComponents
+from selenium.webdriver.remote.webelement import WebElement
+
+
+class Households(BaseComponents):
+    mainContent = 'div[data-cy="main-content"]'
```
KonradMarzec1991, line 6: Many of these attributes and methods can be separated into more general class. Then that class can be used for inheritance

KonradMarzec1991 (changes requested):

pavlo-mk (approved):

KonradMarzec1991 (approved):

PR #3752, opened 2024-04-02
Introduce GHA in staging

wozniakpl:


wozniakpl:
cherry pick can be tricky due to differences in branches

PR #3758, opened 2024-04-03
Upgrade elasticsearch to 8.13.0

KonradMarzec1991:
For `birth_date`, those similarity parameters never worked: 
```
// These parameters were previously *always* parsed by TypeParsers#parseField(), even if they
// made no sense; if we've got here, that means that they're not declared on a current mapper,
// and so we emit a deprecation warning rather than failing a previously working mapping.
private static final Set<String> DEPRECATED_PARAMS = Set.of("store", "meta", "index", "doc_values", "index_options", "similarity");
```
For `boost` on `id`, since es8:
```
Index-time boosts have been deprecated since the 5x line, but it was still possible to declare field-specific boosts in the mappings. This is now removed completely. Indexes built in 7x that contain mapping boosts will emit warnings, and the boosts will have no effect in 8.0. New indexes will not permit boosts to be set in their mappings at all.
```

We add boost in query in `_prepare_query_dict`.


wozniakpl (dismissed):

pavlo-mk (approved):

patryk-dabrowski (approved):

johniak:
-- Conflicts
-- Waiting For @wozniakpl to come back

PR #3794, opened 2024-04-17
[197346] Caching, use cache for payment managerial view

pkujawa:
[AB#197346](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/197346)

- [x] TODO: FE - use ETAG


johniak:
Please verify your code coverage

pavlo-mk (approved):

johniak:
Waiting for discussion with Stefano

patryk-dabrowski (dismissed):

backend/hct_mis_api/apps/payment/api/caches.py
```diff
@@ -0,0 +1,33 @@
...
+from hct_mis_api.api.caches import ModelVersionKeyBitMixin
+from hct_mis_api.apps.core.models import BusinessArea
+
+
+class PaymentPlanProgramVersionsKeyBit(ModelVersionKeyBitMixin, KeyBitBase):
+    def get_data(
+        self, params: Any, view_instance: Any, view_method: Any, request: Any, args: tuple, kwargs: dict
+    ) -> str:
+        business_area = BusinessArea.objects.get(slug=kwargs.get("business_area"))
+        program_ids = request.user.partner.get_program_ids_for_business_area(str(business_area.id))
```
saxix, line 16: I think it is too much - let use a generic program_list:version

backend/hct_mis_api/apps/payment/api/caches.py
```diff
@@ -0,0 +1,33 @@
...
+            program_id = str(program_id)
+            program_ids_with_versions[program_id] = self.get_value_for_key(
+                f"payment_plan_version_for_program_{program_id}"
+            )
+        return str(program_ids_with_versions)
+
+
+class PaymentPlanKeyConstructor(KeyConstructor):
+    unique_method_id = bits.UniqueMethodIdKeyBit()
+    format = bits.FormatKeyBit()
```
saxix, line 29: I suppose we only returns json. Is this needed ?

backend/hct_mis_api/apps/payment/signals.py
```diff
@@ -0,0 +1,22 @@
...
+) -> None:
+    if instance.status in [
+        PaymentPlan.Status.IN_APPROVAL,
+        PaymentPlan.Status.IN_AUTHORIZATION,
+        PaymentPlan.Status.IN_REVIEW,
+        PaymentPlan.Status.ACCEPTED,
+    ]:
+        program_id = instance.program.id
+        current_version = cache.get(f"payment_plan_version_for_program_{program_id}", 0)
+        cache.set(f"payment_plan_version_for_program_{program_id}", value=int(current_version) + 1, timeout=None)
```
saxix, line 22: this is wrong. we must use INCR verb here

backend/tests/apps/api/test_cache.py
```diff
@@ -0,0 +1,50 @@
...
+    def test_etag_decorator_etag_not_exists(self) -> None:
+        response = self.dummy_view(self.DummyRequest({}), (1, 2, 3), {})
+        assert response.status_code == status.HTTP_200_OK
+        assert response.headers["ETAG"] == "etag"
+
+    def test_etag_decorator_old_etag(self) -> None:
+        response = self.dummy_view(self.DummyRequest({"If-None-Match": "old"}), (1, 2, 3), {})
+        assert response.status_code == status.HTTP_200_OK
+        assert response.headers["ETAG"] == "etag"
+
```
saxix, line 41: we should add some test to be sure business logic is not triggered

frontend/src/api/api.ts
```diff
@@ -3,17 +3,34 @@ export const api = {
...
+  cache: new Map(),
+
   async get(url: string, params: Record<string, any> = {}) {
     const query = new URLSearchParams(params).toString();
+
     const response = await fetch(`${this.baseURL}${url}?${query}`, {
       headers: this.headers,
     });
 
+    if (response.status === 304) {
```
saxix, line 15: this is not clear. why do we need this?

saxix (changes requested):

frontend/src/api/api.ts
pkujawa: While you are right that the `304` response is handled by the protocol, the issue is that the request is retried because of the `await response.json()` below in the code, so the return here is needed

johniak (dismissed):

backend/hct_mis_api/apps/payment/signals.py
```diff
@@ -0,0 +1,26 @@
...
+
+from django.core.cache import cache
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from hct_mis_api.api.caches import get_or_create_cache_key
+from hct_mis_api.apps.payment.models import PaymentPlan
+
+
+@receiver(post_save, sender=PaymentPlan)
```
saxix, line 11: should we monitor PaymentRecord too ?

saxix (changes requested):

backend/hct_mis_api/apps/payment/signals.py
pkujawa: I don't think so? As this is a version specifically for `management_payment_plans_list`, we need the status and basic info for Payment Plans for the list view. I don't think any changes to PaymentRecords impact this

johniak (approved):

saxix (approved):

PR #3847, opened 2024-05-09
[196847] No validate end date for existing programmes

patryk-dabrowski:
[AB#196847](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/196847)


mmaciekk (approved):

pkujawa (changes requested):
According to the requirements (specified more in the comments) - the validation should be omitted only if editing FINISHED program

patryk-dabrowski:
> According to the requirements (specified more in the comments) - the validation should be omitted only if editing FINISHED program

@pkujawa What do you mean by FINISHED program? we cannot edit programs with FINISHED status at all

pkujawa (approved):

PR #3884, opened 2024-05-20
[HOTFIX][201273] Fix Grievance creation flow for All Programs, fix existing Grievances

pkujawa:
[AB#201273](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/201273)

### DIFF
--- a/backend/hct_mis_api/one_time_scripts/assign_program_to_grievance_tickets.py
+++ b/backend/hct_mis_api/one_time_scripts/assign_program_to_grievance_tickets.py
@@ -2,12 +2,13 @@ import logging
 
 from django.db.models import Case, F, Q, Value, When
 
+from hct_mis_api.apps.accountability.models import Feedback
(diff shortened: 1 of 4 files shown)


domdinicola:
@pkujawa could you pls explain what happened here? seems a regression

pkujawa:
@domdinicola 
It's just pushed so that @mmaciekk can add his frontend changes to the same branch.
I am preparing a one-time-script to fix all the grievances that do not have a Program assigned but they should (based on household/individual), when they were created in the All Programs flow.
And I asked Maciej to fix the frontend flow (probably a regression) so that the Program is filled based on chosen HH/IND and is editable if empty.

pavlo-mk (approved):

wozniakpl (approved):

pavlo-mk (approved):

PR #3889, opened 2024-05-21
Upgrade elasticsearch to 8.13.0

KonradMarzec1991:
For `birth_date`, those similarity parameters never worked:
```
// These parameters were previously *always* parsed by TypeParsers#parseField(), even if they
// made no sense; if we've got here, that means that they're not declared on a current mapper,
// and so we emit a deprecation warning rather than failing a previously working mapping.
private static final Set<String> DEPRECATED_PARAMS = Set.of("store", "meta", "index", "doc_values", "index_options", "similarity");
```

For `boost` on `id`, since es8:
```
Index-time boosts have been deprecated since the 5x line, but it was still possible to declare field-specific boosts in the mappings. This is now removed completely. Indexes built in 7x that contain mapping boosts will emit warnings, and the boosts will have no effect in 8.0. New indexes will not permit boosts to be set in their mappings at all.
```

We add boost in query in `_prepare_query_dict`.


wozniakpl:
Let's not merge it until I don't check the es update on some eph deployment

patryk-dabrowski (approved):

pavlo-mk (approved):

johniak (approved):

wozniakpl:
We'll be merging that in 2.6, so can go on hold now.

backend/pyproject.toml
```diff
@@ -190,7 +191,7 @@ dependencies = [
     "drf-jwt<2.0.0,>=1.19.2",
     "drf-spectacular[sidecar]<1.0,>=0.27",
     "drf-extensions==0.7.1",
-    "elasticsearch-dsl<8.0.0,>=7.0.0",
+    "elasticsearch-dsl==8.13.0",
```
domdinicola, line 194: do we want to have <9.0 or <8.14? @KonradMarzec1991

backend/pyproject.toml
```diff
@@ -160,7 +161,7 @@ dependencies = [
     "django-cors-headers<4.0.0,>=3.13.0",
     "django-countries<8.0.0,>=7.3.2",
     "django-csp<4.0,>=3.7",
-    "django-elasticsearch-dsl<8.0.0,>=7.2.2",
+    "django-elasticsearch-dsl==8.0.0",
```
domdinicola, line 164: do we want to have <9.0 or <8.1? @KonradMarzec1991

backend/pyproject.toml
KonradMarzec1991: Prefer <8.14

backend/pyproject.toml
KonradMarzec1991: The same like above, 8.1

patryk-dabrowski (approved):

domdinicola:
AB#207778

domdinicola:
@wozniakpl can you coordinate to fix conflicts?

domdinicola:
AB#197101

domdinicola:
@patryk-dabrowski to fix conflicts urgent

patryk-dabrowski:
Conflict resolved

domdinicola (approved):

PR #4020, opened 2024-07-08
202443_Payment_Module_Configure_Mobile_money_delivery_mechanisms_to_send_payment_records_to_WU_through_Payment_Gateway

MarekBiczysko:
[AB#202443](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/202443)


domdinicola:
@MarekBiczysko together with this we need to add a flex field at database level, correct?

MarekBiczysko:
> @MarekBiczysko together with this we need to add a flex field at database level, correct?

I believe this should be done by an end user via django admin?

pavlo-mk (approved):

backend/hct_mis_api/apps/payment/services/payment_gateway.py
```diff
@@ -115,20 +115,14 @@ def get_extra_data(self, obj: Payment) -> Dict:
         return {}
 
     def get_payload(self, obj: Payment) -> Dict:
```
patryk-dabrowski, line 117: you can use `TypedDict` here to create some kind of contract

patryk-dabrowski (approved):

PR #4030, opened 2024-07-10
cypress remove

johniak:


szymon-kellton (approved):

wozniakpl (commented):
`deployment/docker-compose.cy.yml`, `cy` entry in entrypoint and `CYPRESS_TESTING` env should be removed too

domdinicola:
@johniak  this will stay for now. we discussed internally with @ntrncic @saxix

PR #4053, opened 2024-07-16
[208274] Amend button to remove RDI in admin

pkujawa:
[AB#208274](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/208274)

### DIFF
--- a/backend/hct_mis_api/apps/registration_data/admin.py
+++ b/backend/hct_mis_api/apps/registration_data/admin.py
@@ -23,15 +23,12 @@ from hct_mis_api.apps.grievance.models import GrievanceTicket
 from hct_mis_api.apps.household.celery_tasks import enroll_households_to_program_task
 from hct_mis_api.apps.household.documents import get_individual_doc
 from hct_mis_api.apps.household.forms import MassEnrollForm
-from hct_mis_api.apps.household.models import Individual
+from hct_mis_api.apps.household.models import Individual, PendingIndividual
 from hct_mis_api.apps.payment.models import PaymentRecord
 from hct_mis_api.apps.registration_data.models import RegistrationDataImport
-from hct_mis_api.apps.registration_datahub import models as datahub_models
 from hct_mis_api.apps.registration_datahub.celery_tasks import (
     merge_registration_data_import_task,
 )
-from hct_mis_api.apps.registration_datahub.documents import get_imported_individual_doc
-from hct_mis_api.apps.targeting.models import HouseholdSelection
 from hct_mis_api.apps.utils.admin import HOPEModelAdminBase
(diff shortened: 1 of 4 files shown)


backend/hct_mis_api/apps/registration_data/admin.py
```diff
@@ -217,28 +207,18 @@ def generate_query_for_all_grievances_tickets(rdi: RegistrationDataImport) -> Q:
...
                     )
-                    rdi_datahub.delete()
                     GrievanceTicket.objects.filter(
                         RegistrationDataImportAdmin.generate_query_for_all_grievances_tickets(rdi)
                     ).filter(business_area=rdi.business_area).delete()
                     rdi.delete()
                     # remove elastic search records linked to individuals
                     business_area_slug = rdi.business_area.slug
-                    remove_elasticsearch_documents_by_matching_ids(
-                        datahub_individuals_ids, get_imported_individual_doc(business_area_slug)
```
domdinicola, line 240: shouldn't it be like "pending_individuals_ids, get_individual_doc(business_area_slug) "?? @pkujawa

pkujawa: After Datahub Removal, there are no separate objects for `before` and `after` merging RDI. These are just Household/Individuals already after import and merging updates their status to MERGED. So for the MERGED RDI there are no Pending objects anymore.

patryk-dabrowski (approved):

domdinicola:
@pkujawa part were "remove the check `is_not_used_in_targeting` and list the TP that will be removed in cascade in the warning message."

pkujawa:
@domdinicola There was no such a requirement before :) But no TP will be removed by a cascade, only HouseholdSelections. I could list TPs that the households were a part of and now they are removed from it. But TPs are not removed

domdinicola:
@pkujawa  mmm but TPs need to have an Household Selection, no?

pkujawa:
@domdinicola It's a ManyToMany between TP and Household, through the `HouseholdSelection`. `HouseholdSelection` is a single Household-TP connection. TP can be without any households

domdinicola:
@pkujawa let's just show some counts of HS and how many Households and Individuals will be removed.

backend/hct_mis_api/apps/registration_data/admin.py
domdinicola: @pkujawa also would be good to amend the description to explain what will be happening (to TP, household, individuals and _tickets_)

domdinicola (approved):

johniak (approved):

PR #4054, opened 2024-07-17
Selenium: People - happy path

szymon-kellton:


backend/selenium_tests/people/test_people.py
```diff
@@ -26,7 +33,7 @@ def social_worker_program() -> Program:
 
 
 @pytest.fixture
-def add_people(social_worker_program: Program) -> None:
+def add_people(social_worker_program: Program) -> []:
```
wozniakpl, line 36: https://docs.pytest.org/en/7.1.x/how-to/fixtures.html

how about we yield instead of return from that fixture
also, `-> []` will not be the case here.

backend/selenium_tests/people/test_people.py
```diff
@@ -42,6 +49,40 @@ def add_people(social_worker_program: Program) -> None:
         )
         individual = individuals[0]
         create_individual_document(individual)
+    return [individual, household]
+
+
+@pytest.fixture
+def add_people_with_payment_record(add_people: []) -> PaymentRecord:
```
wozniakpl, line 56: how about a default arg of None instead of a mutable empty list.

wozniakpl (changes requested):

backend/selenium_tests/people/test_people.py
```diff
@@ -42,6 +51,40 @@ def add_people(social_worker_program: Program) -> None:
...
+        parent=cash_plan,
+        target_population=target_population,
+        entitlement_quantity="21.36",
+        delivered_quantity="21.36",
+        currency="PLN",
+        status=GenericPayment.STATUS_DISTRIBUTION_SUCCESS,
+    )
+    add_people[1].total_cash_received_usd = "21.36"
+    add_people[1].save()
+    return payment_record
```
wozniakpl, line 87: consider yielding here.

wozniakpl (approved):

pavlo-mk (approved):

PR #4061, opened 2024-07-18
fix python required version

johniak:


MarekBiczysko (approved):

johniak:
> @johniak just curious why?

@domdinicola Because django-admin-sync fails to install with pdm
https://github.com/unicef/hct-mis/actions/runs/9988435667/job/27606230215#step:5:1218

domdinicola:
we are not using the cache @wozniakpl @johniak ?

wozniakpl:
> we are not using the cache @wozniakpl @johniak ?

we are using cache.

domdinicola:
so why we're having an issue with requirements when it was passing in staging?

wozniakpl:
> so why we're having an issue with requirements when it was passing in staging?

most likely, some transitive dependencies have been updated recently and aren't compatible with python 3.12.

domdinicola:
so the image has not been frozen

wozniakpl:
Or do you mean some versioned base image that contains only built dependencies? If so, we would create a base image with fixed dependencies, tag it (e.g., some-base-image:1.0.0), push it to our Docker registry, and update our project Dockerfile to use this base image. This would indeed help stabilize our builds and avoid issues with dependency updates, but it would require extra maintenance to update and version the base image regularly.

PR #4070, opened 2024-07-19
HOTFIX // KOBO Skip validation for pictures

pavlo-mk:
[AB#209023](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/209023): Pull pictures from Kobo flag is ignored

https://github.com/user-attachments/assets/86fa87b3-26af-44ba-b2d5-eb648f6ced72


backend/hct_mis_api/apps/registration_datahub/validators.py
```diff
@@ -1352,7 +1352,7 @@ def _get_field_type_error(self, field: str, value: Any, attachments: list) -> Un
 
             complex_types: Dict[str, Callable] = {
                 "GEOPOINT": self.geopoint_validator,
-                "IMAGE": self.image_validator,
+                # "IMAGE": self.image_validator,  # comment for now as hotfix AB#209023 for skip image validation
```
domdinicola, line 1355: @pavlo-mk this should depend of the form flag.
Pls also change default to check for the flag

pavlo-mk: updated, now it will based on `KoboImportData. pull_pictures` (bool)

PR #4140, opened 2024-08-16
 DEDUP ENGINE - All Individuals should have RDI

pavlo-mk:
[AB#211397](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/211397): DEDUP ENGINE - All Individuals should have RDI


backend/hct_mis_api/apps/program/utils.py
```diff
@@ -59,6 +59,7 @@ def copy_program_object(copy_from_program_id: str, program_data: dict, user: Use
         end_date=None,
         created_by=user,
     )
+    # create default rdi
```
MarekBiczysko, line 62: ?

MarekBiczysko (approved):

patryk-dabrowski (approved):

backend/hct_mis_api/apps/program/utils.py
pavlo-mk: removed just old comment XD

backend/hct_mis_api/apps/household/forms.py
```diff
@@ -220,11 +220,16 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             queryset=Program.objects.filter(business_area_id=self.business_area_id, status=Program.ACTIVE),
             label="Select a program to enroll households to",
         )
+        self.fields["rdi"] = forms.ModelChoiceField(
+            queryset=RegistrationDataImport.objects.filter(business_area_id=self.business_area_id),
+            label="Select a RDI to enroll households to",
+        )
```
pkujawa, line 226: I don't think you should select RDI - I think we want to just create a new RDI for the enrolled records now. But I think @MarekBiczysko was already working on that?

pkujawa: I don't think it would be a good approach to add records to already existing RDI. I think we would need a new RDI

pavlo-mk: updated

johniak (approved):

PR #4291, opened 2024-10-04
[HOTFIX] Fix Document status RDI program population

pavlo-mk:
[AB#217083](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/217083): Ukraine- Merge error with no detials

### DIFF
--- a/backend/hct_mis_api/apps/program/utils.py
+++ b/backend/hct_mis_api/apps/program/utils.py
@@ -264,6 +264,7 @@ class CopyProgramPopulation:
             document.individual = individual_representation
             document.program_id = individual_representation.program_id
             document.rdi_merge_status = rdi_merge_status
+            document.status = Document.STATUS_PENDING
             documents_list.append(document)
         return documents_list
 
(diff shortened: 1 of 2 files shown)


backend/hct_mis_api/apps/program/utils.py
```diff
@@ -264,6 +264,7 @@ def copy_document_per_individual(
             document.individual = individual_representation
             document.program_id = individual_representation.program_id
             document.rdi_merge_status = rdi_merge_status
+            document.status = Document.STATUS_PENDING
```
pkujawa, line 267: I don't think this is needed, `rdi_merge_status` is passed as PENDING from when calling the `CopyProgramPopulation`

backend/hct_mis_api/apps/registration_datahub/tests/test_program_population_to_pending_objects.py
```diff
@@ -371,6 +371,10 @@ def test_create_pending_objects_from_objects(self) -> None:
             pending_document.individual,
             head_of_household_pending_individual,
         )
+        self.assertEqual(
+            pending_document.status,
```
pkujawa, line 375: pending_document is already `Document.pending_objects.first()` so it was already in this status before the fix and this check doesn't seem to bring anything new

backend/hct_mis_api/apps/program/utils.py
pkujawa: Also this needs to stay in an argument of a function, `rdi_merge_status` should be passed here

pkujawa (changes requested):

pkujawa (approved):

domdinicola (approved):

PR #4326, opened 2024-10-14
Script Clean/Remove pre GPF data

pavlo-mk:
[AB#204353](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/204353): Clean/Remove pre GPF data


pkujawa (approved):

MarekBiczysko (approved):

src/hct_mis_api/one_time_scripts/remove_migrated_data_is_original.py
```diff
@@ -6,15 +6,26 @@ def remove_migrated_data_is_original() -> None:
...
-            count, _ = queryset.delete()
-            print(f"Deleted {count} {model.__name__} objects.\n")
+            print(
+                f"*** {model.__name__} All objects: {queryset_all.count()}. "
+                f"Removing objects with 'is_original=True': {queryset_is_original.count()}"
+            )
+
+            try:
+                count, _ = queryset_is_original.delete(soft=False)
+            #  TypeError: QuerySet.delete() got an unexpected keyword argument 'soft'
```
pkujawa, line 27: It will fail for all, because you need to iterate over them; it is not possible to call `soft=` on a queryset

MarekBiczysko: just call qs.delete() in all cases

MarekBiczysko: you can just use delete(), beacuse all of above managers all_objects/default_for_migrations/original_and_repr_objects are not soft deletable, right?

pavlo-mk: yes @MarekBiczysko you are right! that's why test passed

pkujawa: But for the ones using it, you are taking `all_objects` so it actually works in this case :)

pavlo-mk: so for `EntitlementCard` it's `original_and_repr_objects = models.Manager()` :ok_hand:
`Feedback` >> `original_and_repr_objects = OriginalAndRepresentationsManager() > models.Manager` :ok_hand:
`Message` >> `original_and_repr_objects = OriginalAndRepresentationsManager() > models.Manager` :ok_hand:
`HouseholdSelection` >> `original_and_repr_objects = models.Manager()` :ok_hand:

pavlo-mk: hope everything have to be okay 😄

PR #4336, opened 2024-10-16
PaymentPlan add new status DRAFT

pavlo-mk:
[AB#215790](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/215790): Adjust Payment Module to filter out the Payment Plan which are draft (old Target Population)


domdinicola:
@pavlo-mk I think this should go in the long term branch, no?
Unless it's just adding a filter then we can have in develop

pavlo-mk:
will go to long term branch 🥇 📐

pavlo-mk:
will move it in long term branch

PR #4349, opened 2024-10-17
add libraries

domdinicola:


johniak (approved):

pyproject.toml
```diff
@@ -138,7 +138,7 @@ distribution = true
 
 [project]
 name = "hope"
-version = "2.12.0"
+version = "3.0.0"
```
pavlo-mk, line 141: 2.b.c ?? or now it will be v 3

pavlo-mk (approved):

src/hct_mis_api/config/settings.py
```diff
@@ -500,7 +503,7 @@ def masker(key: str, value: Any, config: Dict, request: HttpRequest) -> Any:
 from hct_mis_api.config.fragments.social_auth import *  # noqa: F403, F401, E402
 from hct_mis_api.config.fragments.storages import *  # noqa: F403, F401, E402
 
-LIBRARY_PATHS: bool = env("LIBRARY_PATHS", default=False)
```
wozniakpl, line 503: I'm not sure if removing this particular default will benefit us. It's used only for local selenium testing.

Now on every deployed instance, we'll have to explicitly set this to false.

src/hct_mis_api/config/settings.py
```diff
@@ -466,7 +469,7 @@ def masker(key: str, value: Any, config: Dict, request: HttpRequest) -> Any:
     }
 }
 
-CYPRESS_TESTING = env("CYPRESS_TESTING", default="no") == "yes"
```
wozniakpl, line 469: This var should be removed

src/hct_mis_api/config/settings.py
```diff
@@ -290,19 +293,19 @@ def extend_list_avoid_repeats(list_to_extend: List, extend_with: List) -> None:
...
 
-CACHE_ENABLED = env("CACHE_ENABLED", default=True)
+CACHE_ENABLED = env("CACHE_ENABLED")
 
 CACHES: Dict[str, Any]
 if CACHE_ENABLED:
     CACHES = {
         "default": {
             "BACKEND": "django_redis.cache.RedisCache",
-            "LOCATION": env("CACHE_LOCATION", default=f"redis://{REDIS_INSTANCE}/1"),
```
wozniakpl, line 305: We can remove the default value for REDIS_INSTANCE but we shouldn't remove this here. There may be a service where we depend on setting up only the REDIS_INSTANCE and now it will require explicitly setting the CACHE_LOCATION too.

src/hct_mis_api/config/settings.py
wozniakpl: Oh, nevermind. The default is set in the `env.py`.
```
"LIBRARY_PATHS": (bool, False),
```

Well then this still will work.
But I wouldn't say moving the default there makes it more readable when viewing the code :P

src/hct_mis_api/config/fragments/storages.py
```diff
@@ -1,24 +1,31 @@
 from hct_mis_api.config.env import env
 
-AZURE_ACCOUNT_NAME = env("STORAGE_AZURE_ACCOUNT_NAME", default="")
-AZURE_ACCOUNT_KEY = env("STORAGE_AZURE_ACCOUNT_KEY", default="")
+STORAGES = {
```
wozniakpl, line 3: This config seems weird
`_MEDIA` goes under key `staticfiles` but the coment indicates it's related to statics
`_DEFAULT` goes under key `default` but the comment indicates it being the media part

PR #4386, opened 2024-10-30
individal admin tweaks for wallet

domdinicola:


src/hct_mis_api/apps/household/admin/individual.py
```diff
@@ -50,6 +51,18 @@
 logger = logging.getLogger(__name__)
 
 
+class IndividualDeliveryMechanismDataInline(admin.TabularInline):
+    model = DeliveryMechanismData
+    extra = 0
+    fields = ("delivery_mechanism_choice", "delivery_mechanism", "data", "is_valid")
```
MarekBiczysko, line 57: delivery_mechanism_choice field is legacy and will be dropped, no need to include it here

src/hct_mis_api/apps/household/admin/individual.py
```diff
@@ -50,6 +51,18 @@
...
+    readonly_fields = ("delivery_mechanism_choice", "delivery_mechanism", "data", "is_valid")
```
MarekBiczysko, line 58: delivery_mechanism_choice field is legacy and will be dropped, no need to include it here

MarekBiczysko (changes requested):

MarekBiczysko (approved):

PR #4409, opened 2024-11-05
[WIP] Permissions new rework

pkujawa:


src/hct_mis_api/apps/account/fixtures/initial.json
```diff
@@ -0,0 +1,38 @@
...
+    "level": 1,
+    "lft": 2,
+    "rght": 3,
+    "tree_id": 1,
+    "is_un": true,
+    "parent": 1
+  }
+},
+{
+  "model": "account.role",
```
pavlo-mk, line 28: do we still need this fixtures account.role "Role with all permissions"?

pavlo-mk (approved):
🤞 🔥 🔥 🔥

src/hct_mis_api/apps/account/fixtures/initial.json
pavlo-mk: I think so because we need role will all perms 😉

pkujawa: Yes, we need it in the initial because it is assigned to UNICEF subpartners when creating new BAs so it needs to be already created.

PR #4477, opened 2024-11-27
TP fix filtering if answer is NO

pavlo-mk:
Filter by collector's fields:

  **if answer _YES_ > Delivery mechanism data (Wallet) must be VALID and have FIELD in the data json
  if answer _NO_ > exclude all collectors with valid Delivery mechanism data (Wallet) that has the field**
  
  Example: 
  `api/unicorn/payment/deliverymechanismdata`
Screenshot 2024-11-27 at 10 10 08 AM

### DIFF
--- a/src/hct_mis_api/apps/targeting/models.py
+++ b/src/hct_mis_api/apps/targeting/models.py
@@ -607,16 +607,17 @@ class TargetingCollectorBlockRuleFilter(TimeStampedUUIDModel, TargetingCriteriaF
 
         collectors_ind_query = Individual.objects.filter(
             pk__in=list(collector_primary_qs),
-            delivery_mechanisms_data__is_valid=True,
         )
         # If argument is Yes
         if argument:
             individuals_with_field_query = collectors_ind_query.filter(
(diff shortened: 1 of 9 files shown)


src/hct_mis_api/apps/targeting/models.py
```diff
@@ -607,16 +607,17 @@ def get_query(self) -> Q:
 
         collectors_ind_query = Individual.objects.filter(
             pk__in=list(collector_primary_qs),
-            delivery_mechanisms_data__is_valid=True,
         )
         # If argument is Yes
         if argument:
             individuals_with_field_query = collectors_ind_query.filter(
```
domdinicola, line 613: are Yes and No strings? bool("No") is True @pavlo-mk

pavlo-mk: it's a good question let me double check with @mmaciekk

pavlo-mk: updated this one

pkujawa (approved):

johniak (approved):

PR #4484, opened 2024-11-29
224671 timestamp

domdinicola:
[AB#224671](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/224671)


src/hct_mis_api/contrib/aurora/migrations/0003_migration.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 3.2.25 on 2024-11-29 13:55
+
```
pavlo-mk, line 2: you also have to add this migration into Jan's migrations script `excluded_migrations `

https://github.com/unicef/hope/blob/33609dee2abfa8fa9ca5534502890a0e538dc5b9/src/hct_mis_api/migrations_script/main.py#L48

domdinicola: 👍🏻

johniak (approved):

PR #4513, opened 2024-12-09
RDI Program Population filter by ids

pavlo-mk:
[AB#224070](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/224070): RDI - Modify Import form Programme Population to allow selection of a criteria or specific IDs
[AB#226214](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/226214): FE // RDI - Modify Import form Programme Population to allow selection of a criteria or specific IDs

- [x] Backend
- [x] Frontend

Screenshot 2025-01-14 at 2 08 35 PM
Screenshot 2025-01-14 at 2 10 14 PM


johniak (approved):

src/hct_mis_api/apps/registration_datahub/utils.py
```diff
@@ -72,3 +77,54 @@ def calculate_hash_for_kobo_submission(submission: Dict) -> str:
...
+    program = get_object_or_404(Program, pk=import_to_program_id)
+
+    # filter by rdi.import_from_ids HH or Ins ids based on Program.DCT
+    list_of_ids = [item.strip() for item in import_from_ids.split(",")] if import_from_ids else []
+    if list_of_ids:
+        individual_ids_q = (
+            Q(unicef_id__in=list_of_ids)
+            if program.is_social_worker_program
+            else Q(household__unicef_id__in=list_of_ids)
+        )
```
pkujawa, line 94: we need to also include individuals that have roles in this household.
So if a user provides a list of household IDs, we need to import individuals that are in these households OR have a role in any of these households.

tests/unit/apps/registration_datahub/test_registration_data_import_program_population_mutation.py
```diff
@@ -258,3 +259,81 @@ def test_registration_data_import_create_cannot_check_against_sanction_list(
...
+        )
+
+        rdi = RegistrationDataImport.objects.filter(name="New Import of Data Ind ids").first()
+        self.assertEqual(rdi.status, RegistrationDataImport.IMPORT_SCHEDULED)
+        self.assertEqual(rdi.data_source, RegistrationDataImport.PROGRAM_POPULATION)
+        self.assertEqual(rdi.program.data_collecting_type.type, DataCollectingType.Type.SOCIAL)
+        self.assertEqual(rdi.import_from_ids, ind_ids)
+        self.assertEqual(rdi.number_of_individuals, 2)
+        self.assertEqual(rdi.number_of_households, 2)
+        self.assertEqual(rdi.program_id, self.import_to_program.id)
```
pkujawa, line 339: Could you also create a test with importing by list of households while collector for one of the households is external and belongs to a household that is not listed for import?
Program population import has quite some edge cases so it would be cool to have this case here as well 💯

pavlo-mk: updated

src/hct_mis_api/apps/registration_datahub/utils.py
pavlo-mk: updated

pkujawa (approved):

PR #4529, opened 2024-12-19
Unique constraint for unicef_id in program

pkujawa:


pavlo-mk:
@johniak do we still have to add new migration into `excluded_migrations` migrations_script/main.py?

pavlo-mk (approved):

johniak:
> @johniak do we still have to add new migration into `excluded_migrations` migrations_script/main.py?

no

johniak (approved):

PR #4543, opened 2025-01-08
Hotfix New currency 

pavlo-mk:
[AB#229044](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/229044): Update Sierra Leone Currency


src/hct_mis_api/apps/core/currencies.py
```diff
@@ -125,6 +125,7 @@
 SGD = "SGD"
 SHP = "SHP"
 SLL = "SLL"
```
pkujawa, line 127: `SLL` can be removed, right?

pavlo-mk: thank you

PR #4576, opened 2025-01-18
Global dashboard

srugano:


src/hct_mis_api/apps/dashboard/services.py
```diff
@@ -201,3 +204,138 @@ def refresh_data(cls, business_area_slug: str) -> ReturnDict:
...
+                "individuals": 0,
+                "households": 0,
+                "pwd_counts": 0,
+                "reconciled": 0,
+                "finished_payment_plans": 0,
+                "total_payment_plans": 0,
+            }
+        )
+
+        for item in list(payments_aggregated):
```
pavlo-mk, line 285: iterator will get data in smaller chunks 
```suggestion
        for item in list(payments_aggregated.iterator()):
```

pavlo-mk (approved):

src/hct_mis_api/apps/dashboard/serializers.py
```diff
@@ -1,23 +1,30 @@
...
-    admin1 = serializers.CharField()
     fsp = serializers.CharField()
     delivery_types = serializers.CharField()
     pwd_counts = serializers.IntegerField()
     reconciled = serializers.IntegerField()
     finished_payment_plans = serializers.IntegerField()
     total_payment_plans = serializers.IntegerField()
+
+
+class DashboardHouseholdSerializer(DashboardBaseSerializer):
```
domdinicola, line 22: why don't we have only one serializer?

src/hct_mis_api/apps/dashboard/views.py
```diff
@@ -39,16 +42,28 @@ def get(self, request: Any, business_area_slug: str) -> Response:
         Retrieve dashboard data for a given business area from Redis cache.
         If data is not cached or needs updating, refresh it.
         """
+        is_global = business_area_slug.lower() == "global"
         business_area = get_object_or_404(BusinessArea, slug=business_area_slug)
+        data_cache: Type[DashboardDataCache] = DashboardGlobalDataCache if is_global else DashboardDataCache
+        if is_global:
+            data_cache = DashboardGlobalDataCache  # noqa
```
domdinicola, line 49: twice same if

src/hct_mis_api/apps/dashboard/services.py
```diff
@@ -201,3 +204,138 @@ def refresh_data(cls, business_area_slug: str) -> ReturnDict:
...
+                "currency",
+                "status",
+                "year",
+                "programs",
+                "sectors",
+                "fsp",
+                "delivery_types",
+            )
+            .annotate(
+                total_usd=Sum(
```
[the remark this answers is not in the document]
srugano: Fixed

src/hct_mis_api/apps/dashboard/serializers.py
domdinicola: it is ok... we don't need to use all the fields, but we can save 2 class creations

srugano: Sure. I will use the required argument.

domdinicola: no need... aren't these just to read info?

src/hct_mis_api/apps/dashboard/views.py
srugano: Fixed

domdinicola (approved):

PR #4595, opened 2025-01-27
drop migration model

domdinicola:


pavlo-mk (approved):

johniak (approved):
Nice, I'm really glad we can finally remove this part of code

PR #4661, opened 2025-02-28
REST - Household

pkujawa:


src/hct_mis_api/apps/household/api/views.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # apply admin area limits if partner has restrictions
+        filter_q = Q()
+        area_limits = self.request.user.partner.get_area_limits_for_program(program.id)
+        if area_limits.exists():
+            areas_null = Q(admin_area__isnull=True)
+            areas_query = Q(Q(admin1__in=area_limits) | Q(admin2__in=area_limits) | Q(admin3__in=area_limits))
+            filter_q = Q(areas_null | Q(areas_query))
+        return (
+            super().get_queryset().filter(filter_q).select_related("head_of_household", "program", "admin1", "admin2")
```
domdinicola, line 83: what about admin3? and other models referenced in the serializers? program and so on?

src/hct_mis_api/apps/household/api/mixins.py
```diff
@@ -0,0 +1,22 @@
+from typing import Any
+
+from rest_framework.exceptions import PermissionDenied
+
+
+class CreatorOrOwnerPermissionMixin:
```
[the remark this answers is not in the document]
pkujawa: it's never used separately. CreatorOrOwner is logic specific for household and individual permissions

src/hct_mis_api/apps/household/api/views.py
pkujawa: program is listed here. I added fields that are needed in both list and retrieve. I decided to ignore the ones just for the detail since it made the code more complex than needed in my opinion

PR #4671, opened 2025-03-05
Update django form

pavlo-mk:
[AB#237011](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/237011): Withdraw household from list - programme should have an autocomplete widget
Mar-11-2025 15-49-31

image


pkujawa:
As discussed before creating the ticket, program name is not globally unique. I don't think we can have autocomplete with just names there in this case @domdinicola

pkujawa:
Like I mentioned before, we would need business area linked selection before the program or a different autocomplete. @domdinicola

domdinicola:
@pavlo-mk can't we have it in the same page?
like here https://stg-hope.unitst.org/api/unicorn/household/household/ when you select the business area

PR #4787, opened 2025-04-28
Backmerge master 28 04

MarekBiczysko:


tests/selenium/targeting/test_targeting.py
```diff
@@ -640,7 +640,6 @@ def test_create_targeting_with_pdu_string_criteria(
         assert len(pageTargetingDetails.getHouseholdTableRows()) == 1
         assert pageTargetingDetails.getHouseholdTableCell(1, 1).text == individual1.household.unicef_id
 
-    @pytest.mark.xfail(reason="UNSTABLE")
```
pavlo-mk, line 643: not sure if it passed but we can try... this test is so unstable and painful

pavlo-mk (approved):

PR #4824, opened 2025-05-15
fix search AB#252345

johniak:


src/hct_mis_api/apps/household/filters.py
```diff
@@ -360,12 +360,18 @@ def _search_es(self, qs: QuerySet[Individual], value: str) -> QuerySet[Individua
 
     def _get_elasticsearch_query_for_individuals(self, search: str) -> Dict:
         business_area = self.data["business_area"]
+        encoded_program_id = self.data.get("program")
+        filters = [{"term": {"business_area": business_area}}]
+        if encoded_program_id is not None:
+            program_id = decode_id_string(encoded_program_id)
+            filters.append({"term": {"program_id": program_id}})
```
pkujawa, line 367: Should we do the same for `_get_elasticsearch_query_for_households`?

johniak: yes we should!

PR #4832, opened 2025-05-21
REST GRV mutations

pavlo-mk:
[AB#253112](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/253112): REST Grievance mutations


tests/unit/apps/grievance/test_grievance_views.py
```diff
@@ -0,0 +1,1689 @@
...
+from hct_mis_api.apps.payment.models import PaymentVerification, PaymentVerificationPlan
+from hct_mis_api.apps.program.fixtures import ProgramFactory
+from hct_mis_api.apps.program.models import Program
+from hct_mis_api.apps.utils.elasticsearch_utils import rebuild_search_index
+from hct_mis_api.apps.utils.models import MergeStatusModel
+
+pytestmark = pytest.mark.django_db()
+
+
+class TestGrievanceTicketCreate:
```
pkujawa, line 75: Can you split create and update tickets into separate files? The list, global list and details also have separate as they were too big to hold in one file. (Or rename this file so it doesn't say `views` as it doesn't contain all of them)

tests/unit/apps/grievance/test_grievance_views.py
```diff
@@ -0,0 +1,1689 @@
...
+            "issue_type": 16,
+            "category": 2,
+            "consent": True,
+            "language": "PL",
+            "program": str(self.program.pk),
+            "documentation": [
+                # {
+                #     # "file": SimpleUploadedFile(name="123.pdf", content=b"abc", content_type="application/pdf"),
+                #     "name": "123.pdf"
+                # }
```
pkujawa, line 226: #comment

tests/unit/apps/grievance/test_grievance_views.py
```diff
@@ -0,0 +1,1689 @@
...
+                            "marital_status": SINGLE,
+                            "estimated_birth_date": False,
+                            "relationship": RELATIONSHIP_UNKNOWN,
+                            "role": ROLE_NO_ROLE,
+                            "documents": [
+                                {
+                                    "key": IDENTIFICATION_TYPE_TO_KEY_MAPPING[IDENTIFICATION_TYPE_NATIONAL_ID],
+                                    "country": "POL",
+                                    "number": "123-123-UX-321",
+                                    # "photo": SimpleUploadedFile(name="test_file.pdf", content=b"abc", content_type="application/pdf"),
```
pkujawa, line 247: #comment

tests/unit/apps/grievance/test_grievance_views.py
```diff
@@ -0,0 +1,1689 @@
...
+            "language": "Polish, English",
+            "consent": True,
+            "extras": {},
+        }
+        assert GrievanceTicket.objects.all().count() == 0
+        assert TicketComplaintDetails.objects.all().count() == 0
+        response = self.api_client.post(self.list_url, data, format="json")
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+        assert "Feedback tickets are not allowed to be created through this mutation." in response.json()
+
```
pkujawa, line 444: Could you add a test for create grievance where user doesn't have perm in this program?
And one where he doesn't have access to the area? 
For the second case, grant permission correctly and use `AdminAreaLimitedTo` to limit to different area than the ticket's. Should be `404` Response as result (see `test_grievance_ticket_detail_access_based_on_permissions` for the extended version if not clear)

src/hct_mis_api/apps/account/permissions.py
```diff
@@ -370,20 +370,17 @@ def check_permissions(user: Any, permissions: Iterable[Permissions], **kwargs: A
...
 ) -> None:
     scope = program or business_area
     if not user.is_authenticated or not (
-        check_permissions(
-            user, [general_permission], business_area=business_area, program=program.slug if program else None  # type: ignore
-        )
-        # TODO: does it okay? user.has_perm()
+        user.has_perm(general_permission, scope)
         or (is_creator and user.has_perm(creator_permission, scope))
         or (is_owner and user.has_perm(owner_permission, scope))
```
pkujawa, line 385: ah right, since we need to add .value everywhere, it would be even better to do here
```
        user.has_perm(general_permission.value, scope)
        or (is_creator and user.has_perm(creator_permission.value, scope))
        or (is_owner and user.has_perm(owner_permission.value, scope))
```
        
as we always need a value here

PR #4894, opened 2025-06-17
cleaning HH & IND fields

domdinicola:


src/hct_mis_api/apps/household/migrations/0020_migration.py
```diff
@@ -0,0 +1,62 @@
+from django.db import migrations
+
+BLANK = ""
+
+def replace_blank_with_null(apps, schema_editor):
+    Household = apps.get_model("household", "Household")
+    Individual = apps.get_model("household", "Individual")
+
+    for instance in Household.objects.all():
```
pavlo-mk, line 9: maybe it's better to use `.iterator(chunk_size=1000)` here because qs will be so huge

domdinicola:
@pavlo-mk thanks for the feedback, can you adjust the PR? I agree with you, we will have this in a script.

tests/selenium/grievance/grievance_tickets/test_grievance_tickets.py
```diff
@@ -703,7 +703,7 @@ def test_grievance_tickets_create_new_ticket_Data_Change_Add_Individual_All_Fiel
         pageGrievanceNewTicket.getSelectIndividualdataSeeingdisability().click()
         pageGrievanceNewTicket.select_listbox_element("Some difficulty")
         pageGrievanceNewTicket.getSelectIndividualdataPhysicaldisability().click()
-        pageGrievanceNewTicket.select_listbox_element("None")
+        pageGrievanceNewTicket.select_listbox_element("Cannot do at all")
```
domdinicola, line 706: shall we remove the line? since this should be nullable? @srugano

srugano: Indeed we can remove it.

tests/selenium/programme_population/test_individuals.py
```diff
@@ -111,12 +111,12 @@ def test_smoke_page_individuals_details(
         assert "Head of household (self)" in pageIndividualsDetails.getLabelRelationshipToHoh().text
         assert "-" in pageIndividualsDetails.getLabelPreferredLanguage().text
         assert "HH-00-0000.1380 -Primary collector" in pageIndividualsDetails.getLabelLinkedHouseholds().text
-        assert "None" in pageIndividualsDetails.getLabelObservedDisabilities().text
-        assert "None" in pageIndividualsDetails.getLabelSeeingDisabilitySeverity().text
-        assert "None" in pageIndividualsDetails.getLabelHearingDisabilitySeverity().text
-        assert "None" in pageIndividualsDetails.getLabelPhysicalDisabilitySeverity().text
-        assert "None" in pageIndividualsDetails.getLabelRememberingOrConcentratingDisabilitySeverity().text
-        assert "None" in pageIndividualsDetails.getLabelCommunicatingDisabilitySeverity().text
+        assert "" in pageIndividualsDetails.getLabelObservedDisabilities().text
```
MarekBiczysko, line 114: this assertion won't work

>>> assert "" in "Dom" -> True

MarekBiczysko (changes requested):

tests/selenium/programme_population/test_individuals.py
pkujawa: Also I think it's more readable if we show the `None` on the UI when it's empty, not an empty text for sure.
But I think frontend should already handle that since it displays `-` for other cases like this. Probably it displays `-` now if you check locally or assert in this test.

pkujawa:
The `None` option is also needed in choices for filtering - for example for creating message when we filter by Residence Status, we need to have `None` as an option as well.
Same for all these parameters when creating Targeting.
Right now they will be missing.
Maybe that was the reason to have NONE in the options in the first place.

(Also the `None` selection needs to be checked if works properly, some filters might be ChoiceFilters now or if Targeting handles that properly after it is made available)

PR #4906, opened 2025-06-24
Feature/no role

domdinicola:


pavlo-mk (commented):
and coverage

tests/unit/apps/grievance/test_grievance_data_change_mutations.py
```diff
@@ -256,7 +255,6 @@ def test_grievance_create_individual_data_change(self, _: Any, permissions: List
                                 "maritalStatus": SINGLE,
                                 "estimatedBirthDate": False,
                                 "relationship": RELATIONSHIP_UNKNOWN,
-                                "role": ROLE_NO_ROLE,
```
pkujawa, line 259: Same as in this PR https://github.com/unicef/hope/pull/4894 - we need to make sure that this option is still displayed in the dropdown. We need the `None` option in the dropdown to be able to create a grievance ticket for changing a role for individual `primary/alternate -> No role`. (So the option to remove the role from the individual) Either by adding the None option to the role choices for frontend or adding it on frontend already. And please make sure that this kind of ticket is still creatable and removes the role correctly.

src/hct_mis_api/apps/household/models.py
```diff
@@ -1310,8 +1305,7 @@ def age(self) -> int:
 
     @property
     def role(self) -> str:
-        role = self.households_and_roles.first()
-        return role.role if role is not None else ROLE_NO_ROLE
+        return self.households_and_roles.filter(role__in=[ROLE_PRIMARY, ROLE_ALTERNATE]).first()
```
pkujawa, line 1308: You need to retrieve the `.role` from the object. 
```
        role = self.households_and_roles.first()
        return role.role if role else None
```
The same as before just different default if individual does not have any `IndividualRoleInHousehold`

src/hct_mis_api/apps/household/models.py
```diff
@@ -1445,7 +1439,7 @@ def recalculate_data(self, save: bool = True) -> Tuple[Any, List[str]]:
         return self, update_fields
 
     def count_all_roles(self) -> int:
-        return self.households_and_roles.exclude(role=ROLE_NO_ROLE).count()
+        return self.households_and_roles.filter(role__in=[ROLE_PRIMARY, ROLE_ALTERNATE]).count()
```
pkujawa, line 1442: I think just `return self.households_and_roles.count()` since now if the IndividualRoleInHousehold obj exists, it will only have one of these two options.`

pkujawa:
Also in 
`def handle_role` we can change
` if role in (ROLE_PRIMARY, ROLE_ALTERNATE) and household:` -> `if role and household`:
 and change typing since we probably will take `None` now as the change for `no role` - this flow needs to be adjusted with frontend.

pavlo-mk:
one question about grievance... do we cover every think there if we haven't yet None?

pkujawa:
> one question about grievance... do we cover every think there if we haven't yet None?

We need to make sure that together with this PR we are able to remove existing roles without the option NO_ROLE. `handle_role` will then receive `None` for this case and will remove the existing role. But this needs to be handled in the flow.

PR #4910, opened 2025-06-25
Feature/financial institution lookup endpoint

patsatsia:


src/hct_mis_api/api/endpoints/serializers.py
```diff
@@ -25,3 +25,8 @@ class Meta:
             "valid_until",
             "updated_at",
         )
+
+
+class FinancialInstitutionListSerializer(serializers.Serializer):
+    id = serializers.IntegerField()
+    name = serializers.CharField()
```
domdinicola, line 32: let's add type swift_code and country iso_code3 @patsatsia

PR #4938, opened 2025-07-08
PP - Assign fund commitment - Display FC assigned and its details upon assignation at Release step

pavlo-mk:
[AB#254338](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/254338): Payment - Grey out 'Assign fund commitment' button when FC is assigned to PP
[AB#254343](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/254343): Payment - Display FC assigned and its details upon assignation at Release step
funds_c_updates

Accepted status:
image


MarekBiczysko:
_Payment - Grey out 'Assign fund commitment' button when FC is assigned to PP_

This means that the user cannot change FC assignments once they are set (in case of mistake etc). 
Ticket says:

`After I assign the FC, the assign button remains available, and that could confuse the user whether FC is assigned or not yet. `

This is not true because it's clearly visible in the table below whether FC was assigned or not.

pavlo-mk:
FC can be reassigned and removed. I just added one more check and remove ability assign the same FC

mmaciekk (approved):

PR #4943, opened 2025-07-10
Bug 244721: Grievances - Unable to close NA ticket

pavlo-mk:
[AB#244721](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/244721): Grievances - Unable to close NA ticket

image

### DIFF
--- a/src/hct_mis_api/apps/household/models.py
+++ b/src/hct_mis_api/apps/household/models.py
@@ -1387,10 +1387,10 @@ class Individual(
             except IntegrityError:
                 error_message = f"{self.unicef_id}: Valid Document already exists: {doc.document_number}."
                 raise Exception(error_message)
-            self.accounts.update(active=True)
-            self.duplicate = False
-            self.duplicate_date = timezone.now()
-            self.save()
+        self.accounts.update(active=True)
+        self.duplicate = False
+        self.duplicate_date = timezone.now()
+        self.save()
 
     def set_relationship_confirmed_flag(self, confirmed: bool) -> None:
         self.relationship_confirmed = confirmed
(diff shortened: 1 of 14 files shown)


src/hct_mis_api/apps/household/models.py
```diff
@@ -1378,11 +1378,28 @@ def mark_as_duplicate(self, original_individual: Optional["Individual"] = None)
...
-        self.duplicate_date = timezone.now()
-        self.save()
+        try:
+            self.documents.update(status=Document.STATUS_VALID)
+            self.accounts.update(active=True)
+            self.duplicate = False
+            self.duplicate_date = timezone.now()
+            self.save()
+        # AB#244721
+        except Exception as e:
```
MarekBiczysko, line 1388: IntegrityError

src/hct_mis_api/apps/household/models.py
```diff
@@ -1378,11 +1378,28 @@ def mark_as_duplicate(self, original_individual: Optional["Individual"] = None)
         self.save()
 
     def mark_as_distinct(self) -> None:
-        self.documents.update(status=Document.STATUS_VALID)
-        self.accounts.update(active=True)
-        self.duplicate = False
-        self.duplicate_date = timezone.now()
-        self.save()
+        try:
```
MarekBiczysko, line 1381: To be honest I don't like this solution :)
Maybe we can iterate over each document and update it one by one, then we don;t have to parse exception message etc.

This way also allows to sepreate the accounts logic

MarekBiczysko (changes requested):

pkujawa (approved):

MarekBiczysko (approved):

PR #4950, opened 2025-07-11
Grievance Tickets: Move Role to a specific section

pavlo-mk:
[AB#261631](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/261631): Grievance Tickets: Move Role to a specific section

image
image


pkujawa:
I don't think the removal option is working. The delete button next to the role line should also remove the Role (like for other separate sections - documents etc.). This feature is also the step before removing the None option for the Role (so the remove button should handle this as well now)

pkujawa:
Also, by deafult the Roles are cleared now if noone changes them in the ticket.By deafult, the Roles should stay the same

pkujawa (changes requested):
Also, by deafult the Roles are cleared now if noone changes them in the ticket.By deafult, the Roles should stay the same
image

pkujawa:
And from backend perspective, I guess `role_reassign_data` is not needed on TicketIndividualDataUpdateDetails anymore?
Also we call `reassign_roles_on_update_service`  and `handle_role` on the individual data update which is not valid anymore.
(There is also more role-related logic in the` individual_data_update_service.py`)

PR #4959, opened 2025-07-18
fix Decimal NaN

pavlo-mk:
[AB#264659](https://dev.azure.com/unicef/ICTD-HCT-MIS/_workitems/edit/264659): Syria: Scrolling verification records displays unexpected error
image


mmaciekk (commented):
Maybe "Received amount needs to be a number"

domdinicola (approved):

PR #4968, opened 2025-07-22
REST // Grievance Tickets: Move Role to a specific section

pavlo-mk:
[AB#261631](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/261631): Grievance Tickets: Move Role to a specific section


pkujawa:
Last thing 🏁 : - there is an issue with error display but maybe it is a general issue with REST error in the `extras` (for example if duplicate role is selected - frontend does not show any message)

pkujawa (approved):

PR #5020, opened 2025-08-01
Refactoring rdi endpoints

datamik:


src/hct_mis_api/api/endpoints/rdi/lax.py
```diff
@@ -0,0 +1,317 @@
...
+    PendingAccount,
+)
+from hct_mis_api.apps.periodic_data_update.utils import populate_pdu_with_null_values
+from hct_mis_api.apps.registration_data.models import RegistrationDataImport
+
+if TYPE_CHECKING:
+    from hct_mis_api.apps.core.models import BusinessArea
+
+
+class DocumentSerializer(serializers.ModelSerializer):
```
domdinicola, line 40: can't we reuse existing serializers for document and accounts

src/hct_mis_api/api/endpoints/rdi/lax.py
```diff
@@ -0,0 +1,317 @@
...
+
+                for document_data in documents_data:
+                    document_data["photo"] = self.get_photo(document_data.pop("image", None))
+                    self.save_document(individual, document_data)
+
+                for account_data in accounts_data:
+                    self.save_account(individual, account_data)
+
+                individual_id_mapping[individual.individual_id] = individual.unicef_id
+                errs.append({"pk": individual.pk})
```
johniak, line 168: Why are we adding the individual to errs after successful creation?

johniak: Oh I see this is not only errors it is also results, pleas fix the name

src/hct_mis_api/api/endpoints/rdi/lax.py
```diff
@@ -0,0 +1,317 @@
...
+        total_individuals = 0
+        total_errors = 0
+        total_accepted = 0
+        errs = []
+
+        serializer_context = {
+            "rdi": self.selected_rdi,
+        }
+
+        for individual_raw_data in request.data:
```
johniak, line 151: Maybe we should think about some bulk creation, not sending multiple saves?

src/hct_mis_api/api/endpoints/rdi/lax.py
```diff
@@ -0,0 +1,317 @@
+from functools import cached_property
```
johniak, line 1: I’m not sure this name fully reflects our case. While ‘lax’ suggests a lack of strictness, we actually have validations here, so the name might not be very descriptive. Maybe we could use something that would be easier to understand.

src/hct_mis_api/api/endpoints/rdi/lax.py
```diff
@@ -0,0 +1,317 @@
...
+    first_registration_date = serializers.DateTimeField(default=timezone.now)
+    last_registration_date = serializers.DateTimeField(default=timezone.now)
+    household = serializers.ReadOnlyField()
+    observed_disability = serializers.CharField(allow_blank=True, required=False)
+    marital_status = serializers.CharField(allow_blank=True, required=False)
+    documents = DocumentSerializer(many=True, required=False)
+    birth_date = serializers.DateField(validators=[BirthDateValidator()])
+    accounts = AccountSerializer(many=True, required=False)
+    photo = serializers.CharField(allow_blank=True, required=False)
+    individual_id = serializers.CharField(required=True)
```
johniak, line 79: Since individual_id has no uniqueness constraint, without validation at the creation stage we risk allowing multiple records with the same individual_id to be submitted to the endpoint.

src/hct_mis_api/api/endpoints/rdi/lax.py
datamik: I believe there is no clean way to do it in bulk, as each individual requires additional processing beyond creation.

src/hct_mis_api/api/endpoints/rdi/lax.py
datamik: The purpose of these endpoints are to replace existing lax endpoint. 
@domdinicola Could you help as here?

src/hct_mis_api/api/endpoints/rdi/lax.py
datamik: We can just save individual without this field. The purpose of this field is to map ids from external system to unicef_ids

src/hct_mis_api/api/endpoints/rdi/lax.py
datamik: @johniak @domdinicola 
Bulk_create does not work well with FileField, as it bypasses the save() method, which is required for proper file handling. Additionally, unicef_id is generated at the database level and must be fetched after creation; this logic is currently implemented in the save() method. We can process individuals and documents that contain images, separately, but it seems to be more complex implementation. Are you aware of any better way of doing it?

Batch processing could be implemented in households endpoint, since it does not involve file uploads and does not rely on unicef_id

PR #5034, opened 2025-08-11
HOTFIX // FSP XLS Template - include account details

pavlo-mk:
[AB#259377](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/259377): FSP XLS Template - include account details

image


pavlo-mk:
working on small refactoring...

PR #5037, opened 2025-08-11
more ruff

domdinicola:
[AB#266686](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/266686)
[AB#266690](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/266690)

### DIFF
--- a/tests/unit/apps/vision/test_models.py
+++ b/tests/unit/apps/vision/test_models.py
@@ -18,26 +18,26 @@ class TestFundsCommitmentDBTrigger(TestCase):
         cls.user = UserFactory()
 
     def test_trigger_creates_rows(self) -> None:
-        self.assertEqual(FundsCommitmentGroup.objects.count(), 0)
-        self.assertEqual(FundsCommitmentItem.objects.count(), 0)
+        assert FundsCommitmentGroup.objects.count() == 0
+        assert FundsCommitmentItem.objects.count() == 0
 
         FundsCommitmentFactory(funds_commitment_number="123")
 
-        self.assertEqual(FundsCommitmentGroup.objects.count(), 1)
-        self.assertEqual(FundsCommitmentItem.objects.count(), 1)
+        assert FundsCommitmentGroup.objects.count() == 1
+        assert FundsCommitmentItem.objects.count() == 1
 
         FundsCommitmentFactory(funds_commitment_number="123")
 
-        self.assertEqual(FundsCommitmentGroup.objects.count(), 1)
-        self.assertEqual(FundsCommitmentItem.objects.count(), 2)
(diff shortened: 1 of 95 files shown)


tests/unit/apps/registration_datahub/test_rdi_merge.py
```diff
@@ -261,7 +261,7 @@ def test_merge_rdi_and_recalculation(self) -> None:
         individual_with_invalid_phone_data = Individual.objects.filter(given_name="Jenna").first()
 
         self.assertEqual(individual_with_valid_phone_data.phone_no_valid, True)
-        self.assertEqual(individual_with_valid_phone_data.phone_no_alternative_valid, True)
+        assert individual_with_valid_phone_data.phone_no_alternative_valid is True
 
         self.assertEqual(individual_with_invalid_phone_data.phone_no_valid, False)
         self.assertEqual(individual_with_invalid_phone_data.phone_no_alternative_valid, False)
```
pkujawa, line 267: It's just 1 change here around all the `self.assertEqual`

tests/unit/apps/registration_datahub/test_rdi_people_create.py
```diff
@@ -123,7 +123,7 @@ def test_execute(self) -> None:
         self.assertEqual(alternate_role.individual.full_name, "Collector ForJanIndex_3")
 
         worker_individuals = PendingIndividual.objects.filter(relationship="NON_BENEFICIARY")
-        self.assertEqual(worker_individuals.count(), 2)
+        assert worker_individuals.count() == 2
 
         self.assertEqual(PendingAccount.objects.count(), 3)
         dmd1 = PendingAccount.objects.get(individual__full_name="Collector ForJanIndex_3")
```
pkujawa, line 129: same here, whole file of `self.assertEqual`. If we want to change it, let's change the whole files.
But are we sure we want to change for `assert` when tests inherit from `TestCase`?

tests/unit/apps/vision/test_models.py
```diff
@@ -40,4 +40,4 @@ def test_trigger_creates_rows(self) -> None:
         self.assertEqual(fcg.funds_commitment_items.count(), 2)
 
         fcg = FundsCommitmentGroup.objects.get(funds_commitment_number="345")
-        self.assertEqual(fcg.funds_commitment_items.count(), 1)
+        assert fcg.funds_commitment_items.count() == 1
```
pkujawa, line 43: -||- there is a lot of files like that, where just 1 assert changes

PR #5107, opened 2025-09-02
Alter mixin ordering for "BusinessAreaViewSet" class

patsatsia:
Alter respective test cases

### DIFF
--- a/src/hope/apps/core/api/views.py
+++ b/src/hope/apps/core/api/views.py
@@ -51,8 +51,8 @@ class BusinessAreaViewSet(
     CountActionMixin,
     RetrieveModelMixin,
     ListModelMixin,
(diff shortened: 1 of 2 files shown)


pkujawa (commented):
BTW, I looked into the code history and this test did not test anything till the Mixin was added now.
The test was added before the REST refactor, where the BusinessAreaViewSet inherited from `HOPEAPIView` which handled the Grant permissions.
After REST refactor, the Grants were no longer handled and this test actually worked only because of the 
`self.client.force_authenticate(self.user)`
which just checked if user is authenticated, even if no Grants were granted.
So now it works without `force_authentication`, as it should. It just did not handle the Grant at all after REST refactor and without the Mixin.
@patsatsia @domdinicola

PR #5145, opened 2025-09-10
Better check permissions

srugano:


pavlo-mk (dismissed):

src/hope/apps/dashboard/views.py
```diff
@@ -49,7 +49,14 @@ def get(self, request: Any, business_area_slug: str) -> Response:
         has_permission = False
 
         if is_global:
-            if request.user.is_superuser or request.user.has_perm(Permissions.DASHBOARD_VIEW_COUNTRY.value):
+            if (
+                request.user.is_superuser
```
johniak, line 53: I think we should use has perm form user, this is not checking every case

PR #5229, opened 2025-10-06
Dev fix migration `status_rank`

pavlo-mk:
https://monitoring.hope.unicef.org/organizations/hope/issues/16510/events/939e66b89f214c36b84db15b833f180b/

`cannot insert a non-DEFAULT value into column "status_rank" DETAIL: Column "status_rank" is a generated column.`


mmaciekk (approved):

MarekBiczysko (approved):

johniak (changes requested):
migrations are bad, whjen you call makemigrations it creates a new one

PR #5799, opened 2026-03-10
303284: Add deduplication engine reference PK to Individual model and…

arsen-vs:
… related services

- Introduced a new field `deduplication_engine_reference_pk` in the Individual model to facilitate communication with the biometric deduplication engine.
- Updated the CreateLaxIndividuals endpoint to include this reference in the validated data.
- Enhanced the BiometricDeduplicationService to utilize the new reference PK for individual identification during deduplication processes.
- Added tests to ensure the correct usage of the deduplication reference PK in various service methods and endpoints.

### DIFF
--- a/src/hope/api/endpoints/rdi/lax.py
+++ b/src/hope/api/endpoints/rdi/lax.py
@@ -348,6 +348,7 @@ class CreateLaxIndividuals(CreateLaxBaseView, PhotoMixin):
             self._programme_code,
         )
         validated_data = dict(serializer.validated_data)
+        validated_data["deduplication_engine_reference_pk"] = external_individual_id
         validated_data["flex_fields"] = populate_pdu_with_null_values(
             self._rdi_program, validated_data.get("flex_fields")
         )
--- a/src/hope/api/endpoints/rdi/lax.py
+++ b/src/hope/api/endpoints/rdi/lax.py
@@ -45,6 +45,7 @@ from hope.models import (
     FinancialInstitution,
     FlexibleAttribute,
     Grant,
+    Individual,
     IndividualRoleInHousehold,
     PendingAccount,
     PendingDocument,
@@ -136,7 +137,7 @@ class IndividualSerializer(serializers.ModelSerializer):
     accounts = AccountLaxSerializer(many=True, required=False)
     photo = serializers.CharField(allow_blank=True, required=False)
     disability_certificate_picture = serializers.CharField(allow_null=True, allow_blank=True, required=False)
-    individual_id = serializers.CharField(required=True)
+    individual_id = serializers.CharField(required=True, max_length=255, trim_whitespace=True)
     disability = DisabilityChoiceField(choices=DISABILITY_CHOICES, required=False, allow_blank=True)
     sex = serializers.ChoiceField(SEX_CHOICE, allow_blank=False, default=NOT_COLLECTED)
 
@@ -157,7 +158,26 @@ class IndividualSerializer(serializers.ModelSerializer):
             "program",
             "rdi_merge_status",
             "is_removed",
+            "deduplication_engine_reference_pk",
         ]
+
+    def validate_individual_id(self, value: str) -> str:
+        normalized_value = value.strip()
+        if not normalized_value:
+            raise serializers.ValidationError("This field may not be blank.")
+
+        program = self.context.get("program")
+        if (
+            program
+            and Individual.all_objects.filter(
+                program=program,
+                is_removed=False,
+                deduplication_engine_reference_pk=normalized_value,
+            ).exists()
+        ):
+            raise serializers.ValidationError("An individual with this individual_id already exists in this program.")
+
+        return normalized_value
 
 
 class HandleFlexFieldsMixin:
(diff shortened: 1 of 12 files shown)


src/hope/apps/registration_data/services/biometric_deduplication.py
MarekBiczysko: this function should now take individuals [Queryset] as an argument to avoid unnecessary queries

MarekBiczysko (changes requested):

src/hope/apps/registration_data/services/biometric_deduplication.py
```diff
@@ -338,19 +354,27 @@ def fetch_biometric_deduplication_results_and_process(
...
+                }
+
+                data = self.get_deduplication_set_results(program, list(reference_to_individual_id))
                 similarity_pairs = [
                     SimilarityPair(
                         score=item["score"],
                         status_code=item["status_code"],
-                        first=item["first"]["reference_pk"] or None,
-                        second=item["second"]["reference_pk"] or None,
+                        first=self._resolve_individual_id_from_reference(
```
johniak, line 390: This implementation has a bug.

`fetch_biometric_deduplication_results_and_process` should save all duplicate pairs across the whole program, not only pairs from individuals inside the current RDI scope.

The bug is visible in this scenario:

1. RDI with 1 HH and 1 individual with face "A" is created. `deduplication_engine_reference_pk = "EXT-JAN"`.
2. RDI is sent to the deduplication engine (no duplicates yet).
3. RDI is merged.
4. A new RDI is created with 1 HH and 1 individual with the same face "A". `deduplication_engine_reference_pk = "EXT-JAN2"`.
5. The new RDI is sent to the deduplication engine.
6. Results are fetched (`fetch_biometric_deduplication_results_and_process`).
7. Expected: 1 pair. Actual: 0 pairs.
   The pair should be between "EXT-JAN" (already merged) and "EXT-JAN2".
8. No pair is saved.

Why this happens:

`reference_to_individual_id` is built only from individuals in RDIs that are in review and dedup in progress, not from the whole program.
So in this case it contains only `{"EXT-JAN2": <uuid>}`.

The dedup result item contains:
`first.reference_pk = "EXT-JAN"`, `second.reference_pk = "EXT-JAN2"`.

When creating `SimilarityPair`, `_resolve_individual_id_from_reference` is called with `reference_to_individual_id`.
This produces something like:
```
SimilarityPair(
  ...,
  first="EXT-JAN",   # fallback (not resolved)
  second=<uuid>,     # resolved
)
```

`SimilarityPair.first/second` are later used as FK IDs to `Individual` when saving `DeduplicationEngineSimilarityPair`.
Because one side remains an external reference, the pair is skipped.


Besides fixing this bug, we should also add a test for this exact scenario.

src/hope/api/endpoints/rdi/lax.py
```diff
@@ -334,6 +334,7 @@ def _prepare_individual(
             self.selected_rdi.program.programme_code,
         )
         validated_data = dict(serializer.validated_data)
+        validated_data["deduplication_engine_reference_pk"] = external_individual_id
```
johniak, line 371: We should add stronger input validation for individual_id in the LAX flow.

  Please consider checking:

  1. max_length=255 at serializer level (currently only DB enforces this).
  2. normalization (strip) before save.
  3. reject blank/whitespace-only values.
  4. detect duplicates of individual_id within the same request payload.
  5. validate uniqueness in program with a clear API validation error (instead of only DB IntegrityError).
  6. add tests for all cases above (too long value, blank value, duplicated value in one payload, duplicated value in program).

src/hope/models/individual.py
```diff
@@ -590,6 +597,11 @@ class Meta:
                 condition=Q(is_removed=False) & Q(originating_id__isnull=False),
                 name="originating_id_ind_unique_constraint",
             ),
+            UniqueConstraint(
+                fields=["program", "deduplication_engine_reference_pk"],
+                condition=Q(is_removed=False) & Q(deduplication_engine_reference_pk__isnull=False),
```
johniak: The new uniqueness condition checks IS NOT NULL but not non-empty string. If "" is ever saved, only one blank value per program is allowed.

tests/unit/apps/registration_data/test_biometric_deduplication_service.py
```diff
@@ -149,6 +150,35 @@ def test_upload_individuals_success(
     assert rdi.deduplication_engine_status == RegistrationDataImport.DEDUP_ENGINE_UPLOADED
```
johniak, line 152: As I mentioned in previous comment we have a test gap. New test covers mapping EXT-* refs for pending individuals only. Missing test for the critical mixed case where one side is current pending and the other side is an already merged individual identified by external dedup ref. That case is where current logic is most likely to regress.

src/hope/api/endpoints/rdi/lax.py
johniak, line 146: Because we rely on exclude, which is technically fragile, we can currently send both individual_id and deduplication_engine_reference_pk. In practice, however, we always save individual_id into deduplication_engine_reference_pk. This does not raise an error, but the behavior is confusing. Swagger will also expose this field as a possible input.

johniak (changes requested):

PR #5804, opened 2026-03-12
Feat/reinstate mypy

johniak:
## Summary
  - Add mypy to CI pipeline (alongside lint) with `--no-incremental` flag
  - Add type annotations, `cast()` calls, return types, and `TYPE_CHECKING` imports across 30+ source files
  - Add `# type: ignore[error-code]` where develop's existing code is type-unsafe
  - Achieve 100% patch coverage with targeted tests

  ## Notes on `# type: ignore` usage
  Some lines use `# type: ignore` instead of proper type fixes. These are cases where fixing the types would require runtime logic changes (adding None guards, type conversions, etc.),
  which would obscure the scope of this PR and make it harder to track down regressions. They can be addressed incrementally in follow-up PRs with proper testing.


src/hope/admin/document.py
```diff
@@ -75,7 +75,7 @@ def save_model(self, request: HttpRequest, obj: "Document", form: Any, change: b
         if "cleared" in form.changed_data and obj.individual.relationship == FOSTER_CHILD:
             cleared = form.cleaned_data["cleared"]
             obj.individual.set_relationship_confirmed_flag(cleared)
-            obj.cleared_by = request.user
+            obj.cleared_by = cast("User", request.user)
```
domdinicola, line 78: why cast is needed?

src/hope/admin/account_filters.py
```diff
@@ -19,7 +19,9 @@ class BusinessAreaFilter(SimpleListFilter):
     template = "adminfilters/combobox.html"
 
     def lookups(self, request: HttpRequest, model_admin: "ModelAdmin[Any]") -> list:
-        return BusinessArea.objects.filter(role_assignments__user__isnull=False).values_list("id", "name").distinct()
+        return list(
```
domdinicola, line 22: why this is needed? should not be more efficient to return a QS?

src/hope/admin/business_area.py
```diff
@@ -68,11 +68,11 @@ class BusinessofficeFilter(SimpleListFilter):
     title = "Business Ofiice"
     parameter_name = "bo"
 
-    def lookups(self, request: HttpRequest, model_admin: "ModelAdmin") -> list[tuple[int, str]]:
-        return [(1, "Is a Business Office"), (2, "Is a Business Area")]
+    def lookups(self, request: HttpRequest, model_admin: "ModelAdmin") -> list[tuple[str, str]]:
+        return [("1", "Is a Business Office"), ("2", "Is a Business Area")]
```
domdinicola, line 72: why this change?

src/hope/admin/fsp.py
```diff
@@ -50,15 +51,17 @@ def total_selected_columns(self, obj: Any) -> str:
...
         change: bool,
     ) -> None:
+        if not isinstance(obj, FinancialServiceProviderXlsxTemplate):
+            raise TypeError(f"Expected FinancialServiceProviderXlsxTemplate, got {type(obj).__name__}")
         for required_field in ["payment_id", "delivered_quantity"]:
             if required_field not in obj.columns:
                 raise ValidationError(f"'{required_field}' must be present in columns")
         if not change:
-            obj.created_by = request.user
+            obj.created_by = cast("User", request.user)
```
domdinicola, line 62: why?

src/hope/admin/geo.py
```diff
@@ -157,10 +157,12 @@ def get_queryset(self, request: "HttpRequest") -> "QuerySet":
...
 
 class AreaTypeFilter(RelatedFieldListFilter):
-    def field_choices(self, field: Any, request: "HttpRequest", model_admin: ModelAdmin) -> list[tuple[str, str]]:
+    def field_choices(
+        self, field: "RelatedField", request: "HttpRequest", model_admin: ModelAdmin
+    ) -> "list[tuple[str, str | _StrOrPromise]]":
         if "area_type__country__exact" not in request.GET:
             return []
-        return AreaType.objects.filter(country=request.GET["area_type__country__exact"]).values_list("id", "name")
+        return list(AreaType.objects.filter(country=request.GET["area_type__country__exact"]).values_list("id", "name"))
```
domdinicola, line 165: mmm isn't this more inefficient?

src/hope/admin/grievance.py
```diff
@@ -87,7 +89,8 @@ class GrievanceTicketAdmin(LinkedObjectsMixin, HOPEModelAdminBase):
     show_full_result_count = False
 
     def get_queryset(self, request: HttpRequest) -> QuerySet["GrievanceTicket"]:
-        qs = (
+        qs = cast(
```
domdinicola, line 92: cast is needed?

src/hope/admin/household.py
```diff
@@ -127,11 +127,11 @@ def mass_withdraw(self, request: HttpRequest, qs: QuerySet) -> TemplateResponse
                     for hh in qs.filter(withdrawn=False):
                         service = self._toggle_withdraw_status(
                             request,
-                            hh,
+                            cast("Household", hh),
```
domdinicola, line 132: as above

src/hope/admin/household.py
```diff
@@ -166,11 +166,11 @@ def mass_unwithdraw(self, request: HttpRequest, qs: QuerySet) -> TemplateRespons
                     for hh in qs.filter(withdrawn=True):
                         service = self._toggle_withdraw_status(
                             request,
-                            hh,
+                            cast("Household", hh),
```
domdinicola, line 171: 1

src/hope/admin/individual.py
```diff
@@ -354,7 +354,7 @@ class IndividualCollectionAdmin(admin.ModelAdmin):
     def get_queryset(self, request: HttpRequest) -> QuerySet:
         return super().get_queryset(request).annotate(representations_count=Count("individuals"))
 
-    def number_of_representations(self, obj):
+    def number_of_representations(self, obj: Any) -> int:
```
domdinicola, line 359: shall we remove ANy?

src/hope/admin/kobo_template.py
```diff
@@ -135,7 +137,7 @@ def add_view(
             try:
                 # Load workbook from uploaded file
                 xls_file.seek(0)
-                wb = load_workbook(filename=xls_file, data_only=True)
+                wb = load_workbook(filename=cast("BytesIO", xls_file), data_only=True)
```
domdinicola, line 140: 1

src/hope/admin/registration_data.py
```diff
@@ -208,7 +208,7 @@ def delete_rdi(self, request: HttpRequest, pk: UUID) -> Any:  # TODO: typing
                     self._delete_rdi(rdi)
                     self.message_user(request, "RDI Deleted")
                     LogEntry.objects.log_action(
-                        user_id=request.user.pk,
+                        user_id=cast("int", request.user.pk),
```
domdinicola, line 211: 1

src/hope/admin/registration_data.py
```diff
@@ -292,7 +292,7 @@ def delete_merged_rdi(self, request: HttpRequest, pk: UUID) -> HttpResponse | No
                     self._delete_merged_rdi(rdi)
                     self.message_user(request, "RDI Deleted")
                     LogEntry.objects.log_action(
-                        user_id=request.user.pk,
+                        user_id=cast("int", request.user.pk),
```
domdinicola, line 295: 1

src/hope/admin/rule.py
```diff
@@ -227,7 +227,7 @@ def delete_view(
     ) -> HttpResponse | HttpResponse:
         return super().delete_view(request, object_id, extra_context)
 
-    def render_delete_form(self, request: HttpRequest, context: dict) -> Form:
+    def render_delete_form(self, request: HttpRequest, context: dict) -> HttpResponse:
```
domdinicola, line 230: big change

src/hope/admin/rule.py
```diff
@@ -248,7 +248,7 @@ def process_file(self, request: HttpRequest, pk: UUID) -> HttpResponse:
             state_opts=RuleCommit._meta,
         )
         if request.method == "POST":
-            rule: Rule | None = self.get_object(request, str(pk))
+            rule: Rule | None = cast("Rule | None", self.get_object(request, str(pk)))
```
domdinicola, line 251: 1

src/hope/admin/user.py
```diff
@@ -315,7 +315,7 @@ def get_queryset(self, request: HttpRequest) -> QuerySet:
             )
         )
 
-    def get_readonly_fields(self, request: HttpRequest, obj: Any | None = ...) -> Sequence[str]:
+    def get_readonly_fields(self, request: HttpRequest, obj: Any | None = ...) -> Any:
```
domdinicola, line 318: we shou;d remove Anys...

src/hope/admin/user.py
```diff
@@ -372,6 +372,7 @@ def add_business_area_role(self, request: HttpRequest, queryset: QuerySet) -> Ht
                 with atomic():
                     users, added, removed = 0, 0, 0
                     for u in queryset.all():
+                        u = cast("User", u)
```
domdinicola, line 375: 1

src/hope/admin/utils.py
```diff
@@ -132,8 +132,8 @@ class BusinessAreaForCollectionsListFilter(admin.SimpleListFilter):
     parameter_name = "business_area__exact"
     template = "adminfilters/combobox.html"
 
-    def lookups(self, request: HttpRequest, model_admin: ModelAdmin) -> QuerySet:
-        return BusinessArea.objects.all().values_list("id", "name")
+    def lookups(self, request: HttpRequest, model_admin: ModelAdmin) -> list[tuple[Any, str]]:
```
domdinicola, line 138: 2= inefficent?

src/hope/admin/xlsupdate.py
```diff
@@ -71,7 +71,7 @@ def _stage2(self, request: HttpRequest) -> TemplateResponse:
         )
         xlsx_update_file.save()
         try:
-            updater = IndividualXlsxUpdate(xlsx_update_file)
+            updater = IndividualXlsxUpdate(cast("IO[Any]", xlsx_update_file))
```
domdinicola: 1

src/hope/admin/xlsupdate.py
```diff
@@ -130,7 +130,7 @@ def _render_stage3_report(
     def _stage4(self, request: HttpRequest) -> TemplateResponse | HttpResponseRedirect:
         xlsx_update_file_id = request.POST.get("xlsx_update_file")
         xlsx_update_file = XlsxUpdateFile.objects.get(pk=xlsx_update_file_id)
-        updater = IndividualXlsxUpdate(xlsx_update_file)
+        updater = IndividualXlsxUpdate(cast("IO[Any]", xlsx_update_file))
```
domdinicola: 1

src/hope/api/endpoints/lookups/base.py
```diff
@@ -55,50 +55,50 @@ class FinancialInstitutionAPIView(HOPEAPIView, ListAPIView):
 
 
 class ResidenceStatus(HOPEAPIView):
-    def get(self, request: "Request", **kwargs) -> Response:
+    def get(self, request: "Request", **kwargs: Any) -> Response:
```
domdinicola: we can fix these Any here

src/hope/api/endpoints/rdi/common.py
```diff
@@ -1,17 +1,19 @@
+from typing import Any
+
 from rest_framework import serializers
 
 from hope.models import NOT_DISABLED
 
 
 class NullableChoiceField(serializers.ChoiceField):
-    def to_internal_value(self, data):
+    def to_internal_value(self, data: Any) -> Any:
```
domdinicola: fix these... now or next PRs

src/hope/api/endpoints/rdi/common.py
```diff
@@ -1,17 +1,19 @@
...
         if data == "":
             return None
         return super().to_internal_value(data)
 
 
 class DisabilityChoiceField(serializers.ChoiceField):
-    def to_internal_value(self, data):
+    def to_internal_value(self, data: Any) -> Any:
```
domdinicola: fix

src/hope/api/endpoints/rdi/lax.py
```diff
@@ -52,10 +53,13 @@
     RegistrationDataImport,
 )
 from hope.models.household import NOT_COLLECTED, SEX_CHOICE
-from hope.models.utils import Grant
 
 if TYPE_CHECKING:
-    from hope.models import BusinessArea
+    from uuid import UUID
+
+    from rest_framework.request import Request
```
domdinicola, line 60: shouldn't we add noqa here... its' weird it's not flagged

src/hope/api/caches.py
```diff
@@ -72,7 +75,7 @@ def wrapper(*args: Any, **kwargs: Any) -> Response:
             res.headers.setdefault("Vary", "Authorization, Cookie")
             return res
 
-        return wrapper
+        return cast("F", wrapper)
```
domdinicola: needed?

src/hope/apps/account/authentication.py
```diff
@@ -31,7 +31,7 @@ def user_details(
     user: Any | None = None,
     *args: Any,
     **kwargs: Any,
-) -> HttpResponse:
+) -> Any:
```
domdinicola: maybe a parent class of HttpResponse?

src/hope/apps/account/profile_cache.py
```diff
@@ -18,7 +18,7 @@ def _user_key(self, user_id: int) -> str:
         return f"profile:user:{_NS}:{user_id}"
 
     def _get_or_init(self, key: str, default: int = 1) -> int:
-        return cache.get_or_set(key, default, timeout=None)
+        return cast("int", cache.get_or_set(key, default, timeout=None))
```
domdinicola, line 21: 1

src/hope/apps/accountability/services/feedback_crud_services.py
```diff
@@ -56,7 +56,7 @@ def create(
 
         if not obj.program and cls._has_value(input_data, "program"):
             obj.program = get_object_or_404(Program, id=input_data["program"])
-        obj.created_by = user
+        obj.created_by = cast("User", user)
```
domdinicola, line 59: 1

src/hope/apps/generic_import/forms.py
```diff
@@ -91,7 +116,7 @@ def __init__(self, user, *args, **kwargs):
             self.fields["business_area"].initial = ba
 
             # Get programs for the single BA
-            programs = self._get_program_queryset(ba)
+            programs = self._get_program_queryset(cast("BusinessArea", ba))
```
domdinicola, line 119: 1

src/hope/apps/grievance/services/data_change/individual_data_update_service.py
```diff
@@ -344,7 +344,7 @@ def close(self, user: AbstractUser) -> None:
         new_individual = Individual.objects.select_for_update().get(id=individual.id)
 
         self._validate_phone_numbers(only_approved_data)
-        self._update_household_fields(household, only_approved_data)
+        self._update_household_fields(household, only_approved_data)  # type: ignore[arg-type]
```
domdinicola, line 353: a lot of # type: ignore[arg-type]

src/hope/apps/grievance/services/reassign_roles_services.py
```diff
@@ -218,11 +237,12 @@ def _process_disable_role_reassignment(role_data, individual_to_remove, user, pr
             household.individuals.exclude(id=new_individual.id).update(relationship=RELATIONSHIP_UNKNOWN)
         new_individual.relationship = HEAD
         new_individual.save()
+        _program_arg = getattr(program_or_qs, "pk", None) if isinstance(program_or_qs, UUID) else program_or_qs
         log_create(
             Individual.ACTIVITY_LOG_MAPPING,
             "business_area",
             user,
-            getattr(program_or_qs, "pk", None) if isinstance(program_or_qs, UUID) else program_or_qs,
+            cast("UUID | QuerySet[Program, Program] | str | Program | None", _program_arg),
```
domdinicola, line 245: 1

tests/unit/apps/periodic_data_update/test_api_utils.py
```diff
@@ -0,0 +1,39 @@
+import pytest
```
domdinicola, line 1: what do these _mypy file do?

src/hope/admin/account_filters.py
MarekBiczysko: def lookups(self, request, model_admin):
        """
        Must be overridden to return a list of tuples (value, verbose value)
        """

src/hope/admin/document.py
johniak: Because request.user is not typed as our User class. In Django, request.user is a special type that can be either AnonymousUser or our authenticated user. We are adding a cast here because we know the user is logged in at this point, so in this case it is indeed an instance of our User class.

src/hope/admin/account_filters.py
johniak: django-stubs declares SimpleListFilter.lookups as -> List[Tuple[Any, str]]. The original code returned a QuerySet from values_list().distinct(), which mypy flags as incompatible. I wrapped in list() to satisfy the type. As far as I know django evaluates the result immediately when building the filter choices, so there's no real performance difference

src/hope/admin/fsp.py
johniak: Because request.user is not typed as our User class. In Django, request.user is a special type that can be either AnonymousUser or our authenticated user. We are adding a cast here because we know the user is logged in at this point, so in this case it is indeed an instance of our User class.

src/hope/admin/fsp.py
```diff
@@ -50,15 +51,17 @@ def total_selected_columns(self, obj: Any) -> str:
     def save_model(
         self,
         request: HttpRequest,
-        obj: FinancialServiceProviderXlsxTemplate,
+        obj: Model,
```
MarekBiczysko: why Model not FinancialServiceProviderXlsxTemplate?

src/hope/admin/fsp.py
```diff
@@ -50,15 +51,17 @@ def total_selected_columns(self, obj: Any) -> str:
...
         form: "Form",
         change: bool,
     ) -> None:
+        if not isinstance(obj, FinancialServiceProviderXlsxTemplate):
```
MarekBiczysko: is it possible?

src/hope/admin/partner.py
```diff
@@ -53,14 +54,14 @@ def links_to_objects(cls, objects: Any) -> str:
         rel_list += "</ul>"
         return format_html(rel_list)
 
-    def get_readonly_fields(self, request: HttpRequest, obj: Partner | None = None) -> Sequence[str]:
+    def get_readonly_fields(self, request: HttpRequest, obj: Model | None = None) -> list[str]:
```
MarekBiczysko: why Model?

src/hope/admin/partner.py
```diff
@@ -53,14 +54,14 @@ def links_to_objects(cls, objects: Any) -> str:
...
         additional_fields = []
-        if obj and (obj.is_unicef or obj.is_unicef_subpartner):
+        if obj and isinstance(obj, Partner) and (obj.is_unicef or obj.is_unicef_subpartner):
```
MarekBiczysko: isinstance(obj, Partner) ?

src/hope/admin/partner.py
```diff
@@ -53,14 +54,14 @@ def links_to_objects(cls, objects: Any) -> str:
...
             additional_fields.extend(["name", "parent"])
         return list(super().get_readonly_fields(request, obj)) + additional_fields
 
     def get_form(
-        self, request: HttpRequest, obj: Partner | None = None, change: bool = False, **kwargs: Any
+        self, request: HttpRequest, obj: Model | None = None, change: bool = False, **kwargs: Any
```
MarekBiczysko: why Model?

src/hope/admin/geo.py
johniak: django-stubs declares the return type as list[tuple[...]] (see RelatedFieldListFilter.field_choices in stubs). The original code returned a QuerySet which mypy flags. We wrapped in list() to match the declared type. Django evaluates the result immediately when building filter choices, so no performance impact.

if I use queryset we need to add type ignore.

https://github.com/typeddjango/django-stubs/blob/master/django-stubs/contrib/admin/filters.pyi#L88

src/hope/admin/registration_data.py
johniak: User can be an Anonymus user which has None as type there. We know that we have logged user so we need to add a cast to have correct type

src/hope/admin/registration_data.py
johniak: User can be an Anonymus user which has None as type there. We know that we have logged user so we need to add a cast to have correct type

src/hope/admin/grievance.py
johniak: Problem: LinkedObjectsMixin from smart_admin inherits from admin.ModelAdmin without a generic type parameter.
super().get_queryset(request) - this returns without generic parameter so we can ether ignore or cast

src/hope/admin/household.py
johniak: Same 3rd party library does not use generic parameter

src/hope/admin/household.py
johniak: Same 3rd party library does not use generic parameter

src/hope/admin/rule.py
johniak: I just fixed type to real one. But real fix is just to remove this method at all, it just calls super.

src/hope/admin/individual.py
johniak: We could use IndividualCollection, but obj.representations_count is a dynamically added field from annotate() — it doesn't exist on the model. With IndividualCollection mypy would flag representations_count as unknown attribute. Any avoids that. Alternative: we could type it as IndividualCollection and access via getattr(obj, "representations_count"), but that's worse readability for no real gain.

src/hope/admin/user.py
johniak: Same root cause as the other admin casts. UserAdmin inherits from UserAdminPlus (from unicef_security) which extends admin.ModelAdmin without a generic type parameter.

src/hope/admin/rule.py
johniak: Same issue as the get_queryset cast in grievance.py — RuleAdmin inherits from SyncModelAdmin and LinkedObjectsMixin, both of which extend admin.ModelAdmin without a generic type parameter. This makes get_object() return Model | None instead of Rule | None.

src/hope/admin/kobo_template.py
johniak: request.FILES["xls_file"] returns UploadedFile per django-stubs. load_workbook(filename=...) expects str | Path | IO[bytes]. UploadedFile implements IO[bytes] at runtime but django-stubs doesn't declare this interface on UploadedFile.

src/hope/admin/user.py
johniak: Agree. We changed -> Sequence[str] to -> Any which is a regression. However, the correct solution is not as straightforward as it might seem.
This function overrides get_readonly_fields and returns either:
- super().get_readonly_fields() → list[str] | tuple[str, ...]
- self.get_fields() → list[str | list[str] | tuple[str, ...]]

These are different types. Sequence[str] was already incorrect  from first [commit](https://github.com/unicef/hope/commit/f7219143b31ddd74782f7fcfee96f8d526848ca8#diff-3127446b8d6d828fac2e5bdbb639d7d703e1fcef7b4b2a2dfdef5cb672b6078fR137-R141) — it never matched either return path.

Options:
1. Validate at runtime that get_fields returned only strings, then cast — but what do we do if it didn't?
2. Cast directly, accepting the potential type mismatch

Happy to discuss the best approach, but wanted to flag that this one needs a bit more thought than it looks like.

src/hope/admin/utils.py
johniak: django-stubs declares SimpleListFilter.lookups as -> List[Tuple[Any, str]]. The original code returned a QuerySet from values_list().distinct(), which mypy flags as incompatible. I wrapped in list() to satisfy the type. As far as I know django evaluates the result immediately when building the filter choices, so there's no real performance difference

src/hope/admin/xlsupdate.py
johniak: Agree, to fix

src/hope/admin/xlsupdate.py
johniak: Agree, to fix

src/hope/api/endpoints/lookups/base.py
johniak: Agree, I will fix that

src/hope/api/endpoints/rdi/common.py
johniak: To fix this I will need to do type ignore, parent has a type there to return str always, not str or None
it will be this
```
def to_internal_value(self, data: str) -> str | None:  # type: ignore[override]
```

I will do that in next push in this pr

src/hope/api/endpoints/rdi/common.py
johniak: To fix this because of parent type we need to add ignore
parent type says data can be str | int | tuple[str | int, str | int | tuple[Any, ...]]
```
def to_internal_value(self, data: str) -> str:  # type: ignore[override]
```
I will do that in next push in this pr

src/hope/api/endpoints/rdi/lax.py
johniak: It's not flagged because ruff intentionally doesn't lint imports inside TYPE_CHECKING blocks — they're valid there by design

src/hope/api/caches.py
johniak: I will replace it

src/hope/apps/accountability/services/feedback_crud_services.py
johniak: User has to be casted as above

src/hope/apps/account/profile_cache.py
johniak: cache.get_or_set() returns Any | None per django-stubs. But with default=1 it will never return None — if the key doesn't exist, it sets and returns the default. Django-stubs can't express this guarantee, so the cast narrows Any | None to int to match the return type. The cast is correct here.

src/hope/apps/generic_import/forms.py
johniak: business_areas.first() returns BusinessArea | None per django-stubs. _get_program_queryset() expects BusinessArea. The count() == 1 check above guarantees first() won't return None, but mypy can't infer that. The cast bridges the gap.

src/hope/apps/grievance/services/data_change/individual_data_update_service.py
johniak: In the description you can find my explanation. 
In:
Notes on # type: ignore usage
We would need to change some runtime logic.
I can do it, but I tried to avoid logic changes.

src/hope/apps/grievance/services/reassign_roles_services.py
johniak: getattr returns Any, abd we do not want any

tests/unit/apps/periodic_data_update/test_api_utils.py
johniak: I forgot to merge it into tets files. Fixing

src/hope/admin/partner.py
johniak: fixed

src/hope/admin/partner.py
johniak: fixed

src/hope/admin/partner.py
johniak: fixed

src/hope/admin/fsp.py
johniak: fixed

src/hope/admin/fsp.py
johniak: fixed

src/hope/apps/account/authentication.py
johniak: fixed

domdinicola (approved):

MarekBiczysko (approved):

Jeeziorny:
Can we add mypy to precommit?    
```
-   repo: local
        hooks:
            -   id: mypy
                name: mypy
                entry: uv run tox -e mypy
                language: system
                pass_filenames: false
```

PR #5830, opened 2026-03-19
Sentry n+1 

pavlo-mk:
[AB#302422](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/302422): improve performance and avoid Queries N+1


saxix (changes requested):
stop useing fixtures, please use Factories

pavlo-mk:
> stop useing fixtures, please use Factories

@saxix all this jsons are generated by `drf_api_checker.pytest frozenfixture `

saxix:
> > stop useing fixtures, please use Factories
> 
> @saxix all this jsons are generated by `drf_api_checker.pytest frozenfixture `

ok, so it is fine

src/hope/apps/household/migrations/0035_migration.py
```diff
@@ -0,0 +1,187 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("household", "0034_migration"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="household",
+            name="currency",
+            field=models.CharField(
```
saxix, line 15: let use a callable to avoid this choices and limit future migrations. Should this be linked to the new currency model (cc @domdinicola)

saxix (changes requested):

src/hope/apps/household/migrations/0035_migration.py
pavlo-mk: it can be updated in separate ticket I think

johniak (approved):

PR #5853, opened 2026-03-27
[AB#307495]-setup selenium base + add create program test 

mmaciekk:
[AB#307495](https://dev.azure.com/unicef/ICTD-HCT-MIS/_workitems/edit/307495)


tests/e2e/program_details/test_create_program.py
```diff
@@ -0,0 +1,375 @@
...
+    beneficiary_group: str = "Main Menu",
+) -> None:
+    """Fill mandatory fields on Step 1 (Details) of the programme wizard."""
+    sb.type(INPUT_NAME, name)
+    # MUI DatePicker: click to focus, send_keys to type, then click away to
+    # dismiss the picker overlay before interacting with the next field.
+    sb.click(INPUT_START_DATE)
+    sb.send_keys(INPUT_START_DATE, start_date)
+    sb.click(INPUT_NAME)  # dismiss start-date picker
+    sb.sleep(0.3)
```
Jeeziorny: Maybe we could use sb.wait_for_element_clickable here. Instead of hardcoding sleep we could in some cases proceed faster

tox.ini
```diff
@@ -49,6 +49,26 @@ pass_env =
...
+commands =
+    pytest \
+      -q \
+      -rfE \
+      --no-header \
+      --tb=short \
+      --randomly-seed=42 \
+      --create-db \
+      --reruns-delay 1 \
+      --dist loadscope \
```
Jeeziorny: What's the reason behind using this flag?

tests/e2e/program_details/test_create_program.py
mmaciekk: done.

tox.ini
mmaciekk: It's a pytest-xdist flag that groups tests by module or class so they run on the same worker. This was relevant when tests were class-based and shared state. Now that tests are plain functions with explicit fixtures (no shared class state), there's no reason to force same-worker grouping — the default xdist distribution strategy works fine and can distribute more evenly.

I removed the flag.

tests/e2e/helpers/selenium_base.py
```diff
@@ -0,0 +1,87 @@
...
+                indexedDB.deleteDatabase(db.name);
+            }));
+            window.localStorage.clear();
+            window.sessionStorage.clear();
+            """
+        )
+        self.wait_for_element_visible("#id_username")
+        self.type("#id_username", username)
+        self.type("#id_password", password)
+        self.click('//*[@id="login-form"]/div[3]/input', by="xpath")
```
saxix: prefer CSS to XPATH

tests/e2e/helpers/selenium_base.py
```diff
@@ -0,0 +1,87 @@
...
+        self.sleep(0.3)
```
saxix: avoid sleep

tests/extras/test_utils/selenium.py
```diff
@@ -0,0 +1,87 @@
...
+            window.localStorage.clear();
+            window.sessionStorage.clear();
+            """
+        )
+        self.wait_for_element_visible("#id_username")
+        self.type("#id_username", username)
+        self.type("#id_password", password)
+        self.click('//*[@id="login-form"]/div[3]/input', by="xpath")
+        self.sleep(0.3)
+        self.open("/")
```
saxix, line 45: does login already have a redirect ?

tests/e2e/helpers/selenium_base.py
```diff
@@ -0,0 +1,87 @@
...
+        self.open("/")
+        self.wait_for_ready_state_complete()
+
+    def select_listbox_element(self, name: str, timeout: int = 10):
+        self.wait_for_element_visible('ul[role="listbox"]', timeout=timeout)
+        self.sleep(1)
```
saxix: no sleep

tests/e2e/helpers/selenium_base.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+        self.wait_for_element_visible("#id_username")
+        self.type("#id_username", username)
+        self.type("#id_password", password)
+        self.click('//*[@id="login-form"]/div[3]/input', by="xpath")
+        self.sleep(0.3)
+        self.open("/")
+        self.wait_for_ready_state_complete()
+
+    def select_listbox_element(self, name: str, timeout: int = 10):
```
saxix: this is weird, this function will select any li by text. It should accept a selector and a value

tests/e2e/helpers/selenium_base.py
```diff
@@ -0,0 +1,87 @@
...
+        self.wait_for_element_visible('ul[role="listbox"]', timeout=timeout)
+        self.sleep(1)
+        self.click(f'ul[role="listbox"] li:contains("{name}")')
+        self.wait_for_element_absent('ul[role="listbox"]')
+
+    def select_option_by_name(self, option_name: str):
```
saxix: same here add selector argument

tests/e2e/program_details/test_create_program.py
```diff
@@ -0,0 +1,309 @@
...
+)
+from hope.models import BusinessArea
+
+pytestmark = pytest.mark.django_db()
+
+# ---------------------------------------------------------------------------
+# Selectors – Programme Management wizard (3-step)
+# ---------------------------------------------------------------------------
+# Navigation
+NAV_PROGRAMMES = 'a[data-cy="nav-Programmes"]'
```
saxix: All these values need to be double checked. 
All form element should have a name, so use it. All buttons should have a Label.
Not sure why this should be useful. 99% of the times each one should be used one time only, making these constance not really useful.

tests/e2e/program_details/test_create_program.py
```diff
@@ -0,0 +1,309 @@
...
+    end_date: str = "2033-12-12",
+    sector: str = "Child Protection",
+    dct: str = "Full",
+    beneficiary_group: str = "Main Menu",
+) -> None:
+    sb.type(INPUT_NAME, name)
+    sb.click(INPUT_START_DATE)
+    sb.send_keys(INPUT_START_DATE, start_date)
+    sb.click(INPUT_NAME)
+    sb.wait_for_element_clickable(INPUT_END_DATE)
```
saxix: It is better to take a different approach:
 
 - el = sb.get_element()
 - el.type("ssssss")
 - el.click()

the current one is more expensive

tests/e2e/program_details/test_create_program.py
```diff
@@ -0,0 +1,309 @@
...
+    sb.wait_for_text("Test Programme", DETAIL_HEADER)
+    sb.assert_text("DRAFT", DETAIL_STATUS)
+    sb.assert_text("1 May 2023", LABEL_START_DATE)
+    sb.assert_text("12 Dec 2033", LABEL_END_DATE)
+    sb.assert_text("Child Protection", LABEL_SECTOR)
+    sb.assert_text("Full", LABEL_DCT)
+    sb.assert_text("Regular", LABEL_FREQ)
+    sb.assert_text("-", LABEL_ADMIN_AREAS)
+    sb.assert_text("No", LABEL_CASH_PLUS)
+    sb.assert_text("0", LABEL_SIZE)
```
saxix: well done

tox.ini
```diff
@@ -49,6 +49,25 @@ pass_env =
 commands =
     pre-commit run --all-files --show-diff-on-failure
 
+[testenv:selenium]
```
saxix: selenium test should be executed together with other tests

tests/extras/test_utils/selenium.py
```diff
@@ -0,0 +1,87 @@
...
+- Mock only external dependencies (network, S3, Celery, integrations).
+  Never mock the code under test.
+- Do not create utility helpers — keep everything in fixtures and tests.
+"""
+
+from django.conf import settings
+from seleniumbase import BaseCase
+
+
+class HopeTestBrowser(BaseCase):
```
saxix, line 5: this should not be there. Move to a test_utils module

tests/e2e/program_details/test_create_program.py
```diff
@@ -0,0 +1,309 @@
...
+LABEL_START_DATE_FIELD = 'div[data-cy="date-picker-filter"]'
+LABEL_SECTOR_FIELD = 'div[data-cy="input-sector"]'
+LABEL_DCT_FIELD = 'div[data-cy="input-data-collecting-type"]'
+
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+@pytest.fixture
+def browser_sb(live_server_with_static, request) -> Generator[HopeTestBrowser, None, None]:
```
saxix: let use just `browser()` as in other  apps

tests/e2e/program_details/test_create_program.py
```diff
@@ -0,0 +1,309 @@
...
+    sb._using_sb_fixture_no_class = True
+    sb_config._sb_node[request.node.nodeid] = sb
+    yield sb
+    if sb._needs_tearDown:
+        sb.tearDown()
+        sb._needs_tearDown = False
+
+
+@pytest.fixture
+def login_sb(browser_sb: HopeTestBrowser) -> HopeTestBrowser:
```
saxix: remove

saxix (changes requested):
as general consideration, use `browser` instead of `sb`

tests/e2e/program_details/test_create_program.py
mmaciekk: It's easier to read     sb.type(INPUT_ADMIN_AREAS, "Kabul Province"),
than

sb.type(input[data-cy="input-administrativeAreasOfImplementation"], "Kabul Province").

It seems quite useful to me.

tests/extras/test_utils/selenium.py
mmaciekk: The self.open("/") after login is intentional — Django admin login redirects back to /api/admin/, not to the SPA at /. The explicit open("/") navigates to the frontend.

tests/e2e/new_selenium/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+        sb._needs_tearDown = False
+
+
+@pytest.fixture
+def login(browser: HopeTestBrowser) -> HopeTestBrowser:
+    browser.login()
+    return browser
+
+
+@pytest.fixture(autouse=True)
```
Jeeziorny: One of our criteria for new approach for rewriting UTs was to not to use autouse=True, feel like we should remove this one too

tests/e2e/new_selenium/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+def test_failed_check():
```
Jeeziorny: also I'm not sure if this function is needed, looks like leftover from class approach

tests/e2e/SELENIUM_E2E_PATTERNS.md
```diff
@@ -0,0 +1,757 @@
+# Selenium E2E Test Patterns — HOPE Project
+
+This document is a reference guide for creating new SeleniumBase E2E tests in the HOPE project. It describes the Aurora-style setup pattern, the `HopeTestBrowser` wrapper, available selectors, and a step-by-step template for writing new tests.
```
Jeeziorny: HopeTestBrowser no longer exist, am I right?

tests/extras/test_utils/selenium.py
```diff
@@ -0,0 +1,77 @@
...
+        self.open("/")
+        self.wait_for_ready_state_complete()
+
+    def select_listbox_element(
+        self, name: str, selector: str = 'ul[role="listbox"]', timeout: int = 10
+    ):
+        self.wait_for_element_visible(selector, timeout=timeout)
+        elements = self.find_elements(f"{selector} li")
+        for element in elements:
+            if element.text.strip() == name:
```
Jeeziorny, line 56: we shouldn't use if statements in tests from what I understand

tests/e2e/new_selenium/program_details/test_create_program.py
```diff
@@ -0,0 +1,288 @@
...
+    browser.assert_text("DRAFT", DETAIL_STATUS)
+
+
+def test_create_programme_validation_empty_fields(
+    browser: HopeTestBrowser, unhcr_partner: None
+) -> None:
+    browser.login()
+    _navigate_to_programme_management(browser)
+
+    browser.scroll_main_content(-600)
```
Jeeziorny: Is scroll necessary here?

tests/e2e/new_selenium/program_details/test_create_program.py
```diff
@@ -0,0 +1,288 @@
...
+
+# ---------------------------------------------------------------------------
+# Selenium page-interaction helpers (kept thin — no business logic)
+# ---------------------------------------------------------------------------
+def _navigate_to_programme_management(browser: HopeTestBrowser) -> None:
+    browser.click(NAV_PROGRAMMES)
+    browser.wait_for_text("Programme Management", HEADER_TITLE)
+
+
+def _fill_required_fields(
```
Jeeziorny: I think this function name should be more explicit. What name can we use for fields that this function fills? How about "_fill_programme_create_required_fields"?

tests/extras/test_utils/selenium.py
mmaciekk: The "no if/for/while" rule applies to test function bodies, not to helper methods on the HopeTestBrowser class.

tests/extras/test_utils/selenium.py
mmaciekk: we login via django admin, not azure.

tests/e2e/new_selenium/program_details/test_create_program.py
```diff
@@ -0,0 +1,267 @@
...
+    browser.wait_for_text("Full Programme", HEADER_TITLE)
+    browser.assert_text("DRAFT", DETAIL_STATUS)
+    browser.assert_text("1 Jan 2022", LABEL_START_DATE)
+    browser.assert_text("1 Feb 2035", LABEL_END_DATE)
+    browser.assert_text("Health", LABEL_SECTOR)
+    browser.assert_text("Partial", LABEL_DCT)
+    browser.assert_text("One-off", LABEL_FREQ)
+    browser.assert_text("Kabul Province", LABEL_ADMIN_AREAS)
+    browser.assert_text("Yes", LABEL_CASH_PLUS)
+    browser.assert_text("0", LABEL_SIZE)
```
johniak: This test is framed as covering the "all fields" path, but after save it never checks that description, budget, population goal, beneficiary group, or the created time-series field were actually persisted. A regression in any of those paths would still pass.

tests/e2e/new_selenium/program_details/test_create_program.py
```diff
@@ -0,0 +1,267 @@
...
+    )
+
+    browser.scroll_main_content(600)
+    browser.click(BTN_NEXT)
+
+    browser.wait_for_element_visible(BTN_SAVE)
+    browser.click(BTN_SAVE)
+
+    browser.wait_for_text("TSF Programme", HEADER_TITLE)
+    browser.assert_text("DRAFT", DETAIL_STATUS)
```
johniak: This scenario tests the time-series payload extensively, but the assertions only check the title and status. We should add assertions to verify that every field is set and displayed correctly.

johniak (changes requested):

tests/e2e/new_selenium/program_details/test_create_program.py
```diff
@@ -114,16 +111,10 @@ def _add_time_series_field(
         browser.type(INPUT_PDU_ROUND_NAME.format(index, ri), rname)
 
 
-# ---------------------------------------------------------------------------
-# Tests
-# ---------------------------------------------------------------------------
-def test_create_programme_mandatory_fields_only(
-    browser: HopeTestBrowser, unhcr_partner: None
-) -> None:
+def test_create_programme_mandatory_fields_only(browser: HopeTestBrowser, unhcr_partner: None) -> None:
```
saxix: what is the reason of this `unhcr_partner: None`

tests/e2e/new_selenium/program_details/test_create_program.py
```diff
@@ -59,9 +59,6 @@
 LABEL_DCT_FIELD = 'div[data-cy="input-data-collecting-type"]'
 
 
-# ---------------------------------------------------------------------------
-# Selenium page-interaction helpers (kept thin — no business logic)
-# ---------------------------------------------------------------------------
 def _navigate_to_programme_management(browser: HopeTestBrowser) -> None:
```
saxix: remove this "2 lines functions", they offer very poor improvements but lowering redability. Also the usage of constants should be avoided

tests/e2e/new_selenium/program_details/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+    PartnerFactory,
+    RoleAssignmentFactory,
+    RoleFactory,
+)
+from hope.models import BusinessArea
+
+
+@pytest.fixture
+def unhcr_partner() -> None:
+    """Create UNHCR partner with a role in Afghanistan."""
```
Jeeziorny:
```suggestion
```

tests/e2e/SELENIUM_E2E_PATTERNS.md
```diff
@@ -0,0 +1,256 @@
...
+| `business_area`           | Afghanistan BA with flags, settings  |
+| `live_server_with_static` | Django live server with static files |
+
+---
+
+## SeleniumBase API Quick Reference
+
+Full docs: [seleniumbase.io/help_docs/method_summary](https://seleniumbase.io/help_docs/method_summary/)
+
+### Core methods
```
Jeeziorny: I think this section can be removed, because above we have link to the reference. WDYT?

tests/e2e/SELENIUM_E2E_PATTERNS.md
```diff
@@ -0,0 +1,256 @@
...
+
+---
+
+## Writing a New Test
+
+### 1. File structure
+
+```
+tests/e2e/new_selenium/<module>/test_<feature>.py
+tests/e2e/new_selenium/<module>/conftest.py   # only if domain fixtures needed
```
Jeeziorny: fixtures should be defined in conftest if and only if they are shared between other test files. Otherwise fixtures should be keep within the same file as test that is using them

tests/e2e/SELENIUM_E2E_PATTERNS.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+### 2. Template
```
Jeeziorny, line 167: I think this section isn't necessary. You can point to the location of the first new-style selenium test

tests/e2e/SELENIUM_E2E_PATTERNS.md
```diff
@@ -0,0 +1,256 @@
...
+
+    # Interact
+    login.click(BTN_ACTION)
+    login.type('input[data-cy="input-field"]', "value")
+
+    # Assert
+    login.assert_text("Expected Text", 'div[data-cy="label-Field"]')
+```
+
+### 3. Domain fixture example (local conftest.py)
```
Jeeziorny: I think this one can be removed

tests/e2e/SELENIUM_E2E_PATTERNS.md
```diff
@@ -0,0 +1,256 @@
...
+Existing raw Selenium tests (`Common → BaseComponents → PageObject`) are unaffected.
+
+---
+
+## Fixture Setup
+
+The `browser` and `login` fixtures are defined in `tests/e2e/new_selenium/conftest.py`.
+Do NOT redefine them. Create local `conftest.py` only for domain-specific data fixtures.
+
+### Autouse fixtures (from parent conftest.py)
```
Jeeziorny, line 61: Add a note, that those are already existing autouse fixtures and we can't add new ones (according to the new style tests criteria)

Jeeziorny (changes requested):

pyproject.toml
```diff
@@ -203,6 +204,9 @@ package-dir = { "" = "src" }
 packages.find.where = [ "src" ]
 
 [tool.uv]
+override-dependencies = [
+  "pytest-html==4.0.2",
```
domdinicola, line 211: i'm not a fan of these overrides

mmaciekk: @domdinicola Me neither, but here is the reason why.
https://github.com/seleniumbase/SeleniumBase/issues/3619

tests/e2e/new_selenium/program_details/test_create_program.py
```diff
@@ -0,0 +1,338 @@
+import pytest
+
+from extras.test_utils.selenium import HopeTestBrowser
+
+pytestmark = pytest.mark.django_db()
+
+# ---------------------------------------------------------------------------
+# Selectors – Programme Management wizard (3-step)
```
Jeeziorny: I think those comments with -------- are unnecessary. What do you think?

tests/e2e/conftest.py
```diff
@@ -712,14 +712,6 @@ def pytest_runtest_makereport(item: Item, call: CallInfo[None]) -> None:
...
 @pytest.fixture(autouse=True)
-def test_failed_check(request: FixtureRequest, browser: Chrome) -> None:
-    yield
-    if request.node.rep_setup.failed:
-        pass
-    elif request.node.rep_setup.passed and request.node.rep_call.failed:
-        screenshot(browser, request.node.nodeid)
-
-
 def attach(data=None, path=None, name="attachment", mime_type=None, request=None):
```
Jeeziorny, line 727: In this PR it looks like `@pytest.fixture(autouse=True)` now is decorating attach method, which wasn't decorated before. If it's true, remove this autouse.

Jeeziorny (approved):

tests/e2e/new_selenium/program_details/test_create_program.py
mmaciekk: sure

tests/e2e/conftest.py
mmaciekk: ok

tests/e2e/new_selenium/grievance/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@pytest.fixture
+def social_worker_program() -> Program:
+    dct = DataCollectingTypeFactory(type=DataCollectingType.Type.SOCIAL)
+    beneficiary_group = BeneficiaryGroup.objects.get(name="People")
+    return ProgramFactory(
+        name="Social Program",
+        status=Program.ACTIVE,
+        business_area=BusinessArea.objects.get(slug="afghanistan"),
```
saxix: use factory here

tests/e2e/new_selenium/program_details/test_create_program.py
```diff
@@ -0,0 +1,336 @@
...
+LABEL_END_DATE = 'div[data-cy="label-END DATE"]'
+LABEL_SECTOR = 'div[data-cy="label-Sector"]'
+LABEL_DCT = 'div[data-cy="label-Data Collecting Type"]'
+LABEL_FREQ = 'div[data-cy="label-Frequency of Payment"]'
+LABEL_ADMIN_AREAS = 'div[data-cy="label-Administrative Areas of implementation"]'
+LABEL_CASH_PLUS = 'div[data-cy="label-CASH+"]'
+LABEL_SIZE = 'div[data-cy="label-Programme size"]'
+LABEL_DESCRIPTION = 'div[data-cy="label-Description"]'
+LABEL_BENEFICIARY_GROUP = 'div[data-cy="label-Beneficiary Group"]'
+LABEL_PROGRAMME_CODE = 'div[data-cy="label-Programme Code"]'
```
mmaciekk: Remove constants, move selectors directly to the tests.

mmaciekk: dont search twice for the object

mmaciekk: prefer to use input name (less expensive identifier) instead of data-cy. 
always use css path, never use xpath

mmaciekk: dont use css classes for identifying an object

tests/e2e/new_selenium/program_details/test_create_program.py
mmaciekk: dead code, not needed, removed.

tests/e2e/new_selenium/grievance/conftest.py
mmaciekk: ok

Jeeziorny (approved):

pavlo-mk (approved):

.github/workflows/tests.yml
```diff
@@ -232,7 +231,7 @@ jobs:
           key: ${{ runner.os }}-venv
 
       - name: Run Selenium suite
-        run: uv run tox -e tests --  pytest ./tests/e2e  -q -rfE --no-header --tb=short --no-migrations --randomly-seed=42 --dist=loadgroup --create-db  -n auto --cov-report xml:test-coverage/coverage.xml --junit-xml=test-results.xml --cov-config=.coveragerc --cov=hope
+        run: uv run tox -e tests --  pytest ./tests/e2e  -q -rfE --no-header --tb=short --no-migrations --randomly-seed=42 --dist=loadgroup --create-db  -n auto --headless --cov-report xml:test-coverage/coverage.xml --junit-xml=test-results.xml --cov-config=.coveragerc --cov=hope
```
saxix: why do we still need 2 different invocations for selenium/unit ?

tests/e2e/new_selenium/program_details/test_create_program.py
```diff
@@ -0,0 +1,271 @@
...
+        round_names=["Check 1", "Check 2", "Check 3"],
+    )
+
+    browser.scroll_main_content(600)
+    browser.click('button[data-cy="button-next"]')
+
+    browser.wait_for_element_visible('button[data-cy="button-save"]')
+    browser.click('button[data-cy="button-save"]')
+
+    browser.wait_for_text("TSF Programme", 'h5[data-cy="page-header-title"]')
```
saxix, line 215: Q: do we need all these asserts? Can they change based on the previous "click"? It this is the case pease add a note

tests/e2e/new_selenium/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+def browser(live_server_with_static, request) -> Generator[HopeTestBrowser, None, None]:
+    sb = HopeTestBrowser("base_method")
+    sb.live_server_url = str(live_server_with_static)
+    sb.setUp()
+    sb._needs_tearDown = True
+    sb._using_sb_fixture = True
+    sb._using_sb_fixture_no_class = True
+    sb_config._sb_node[request.node.nodeid] = sb
+    yield sb
+    if sb._needs_tearDown:
```
saxix: sb._needs_tearDown seems to be always True.  Clean this or explain the reason

tests/e2e/new_selenium/program_details/test_create_program.py
```diff
@@ -0,0 +1,271 @@
...
+    browser.click('button[data-cy="button-add-time-series-field"]')
+    browser.type(f'input[name="pduFields.{index}.label"]', label)
+    browser.select_dropdown_option(f"pduFields.{index}.pduData.subtype", subtype)
+    browser.select_dropdown_option(f"pduFields.{index}.pduData.numberOfRounds", num_rounds)
+    for ri, rname in enumerate(round_names):
+        browser.type(f'input[name="pduFields.{index}.pduData.roundsNames.{ri}"]', rname)
+
+
+def test_create_programme_mandatory_fields_only(browser: HopeTestBrowser) -> None:
+    browser.login()
```
saxix: do not see security check here. we should login with no-perm user and add required perm using context manager

tests/extras/test_utils/selenium.py
```diff
@@ -0,0 +1,87 @@
...
+    https://github.com/unicef/hope-aurora/blob/develop/tests/extras/testutils/selenium.py
+    """
+
+    live_server_url: str = ""
+
+    def setUp(self, masterqa_mode=False):
+        super().setUp()
+
+    def tearDown(self):
+        self.save_teardown_screenshot()
```
saxix: we are saving a screenshot for any test. SB do that bby default for failing tests

saxix (changes requested):

.github/workflows/tests.yml
mmaciekk: What do you think about a three commands approach?

1. Locally: pytest ./tests -m unit --no-cov should
     run only unit tests, skip e2e
     
2. Locally: pytest ./tests -m selenium --no-cov
     should run only e2e tests
     
3. Locally: pytest ./tests --no-cov should run
     everything
     
     This will allow someone who works on E2E tests to run them locally, and also all tests can be run together if needed.

tests/e2e/new_selenium/program_details/test_create_program.py
mmaciekk: Yes we do need these asserts because we need to check if the program has been created with the data added in the form and the user is redirected to the details page where all the program data is rendered correctly.

I don't understand the question "Can they change based on the previous click?".  We fill in the form, click save, get redirected to the details page and need to verify if all the data renders correctly.

tests/e2e/new_selenium/conftest.py
mmaciekk: This is not needed. I removed it.

tests/e2e/new_selenium/program_details/test_create_program.py
mmaciekk: Ok. I added a no-perm user and grant_permission function.

tests/extras/test_utils/selenium.py
mmaciekk: removed

tests/e2e/new_selenium/grievance/test_grievance_tickets.py
```diff
@@ -26,7 +26,8 @@ def test_create_new_ticket_referral(
         Permissions.POPULATION_VIEW_INDIVIDUALS_DETAILS,
     ):
         browser.login(username="noperm_user", password="testtest2")
-        browser.js_click('a[data-cy="nav-Grievance"]')
+        browser.wait_for_element_visible('a[data-cy="nav-Grievance"]', timeout=30)
+        browser.click('a[data-cy="nav-Grievance"]')
```
saxix: avoid double lookups:

```
el = browser.wait_for_element_visible('a[data-cy="nav-Grievance"]', timeout=30)
el.click()
```
OR 
 
```
browser.wait_for_element_visible('a[data-cy="nav-Grievance"]', timeout=30).click()
```

saxix (changes requested):

tests/e2e/new_selenium/grievance/test_grievance_tickets.py
mmaciekk: Fixed it.

saxix (approved):

Jeeziorny (approved):

PR #5856, opened 2026-03-30
v4.13

domdinicola:


johniak:
@domdinicola @ntrncic How it is possible that we have +19,620 lines modified  in this PR and 3 uncovered lines makes  96.80851% patch coverage?

@domdinicola Do we want to keep it in “Ready for review” status and clutter the list?

PR #5860, opened 2026-03-30
Fixes Sentry N+1

pavlo-mk:
[AB#302422](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/302422): improve performance and avoid Queries N+1


MarekBiczysko (approved):

tests/unit/apps/account/test_export_users_xlsx.py
```diff
@@ -0,0 +1,103 @@
+from typing import Any
```
Jeeziorny, line 1: Since we're solving query-related problem, maybe it's worth to add query count assertion in the tests?

Jeeziorny (approved):

johniak (approved):

PR #5863, opened 2026-03-31
[308030] Sentry - slow DB issues

pkujawa:
[AB#308030](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/308030)


src/hope/apps/grievance/migrations/0012_migration.py
```diff
@@ -0,0 +1,27 @@
...
+        ("core", "0017_migration"),
+        ("geo", "0006_migration"),
+        ("grievance", "0011_migration"),
+        ("program", "0017_migration"),
+        ("registration_data", "0012_migration"),
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+    ]
+
+    operations = [
+        migrations.AddIndex(
```
domdinicola: shall we create these async?

domdinicola (changes requested):

src/hope/apps/grievance/migrations/0012_migration.py
pkujawa: good point

src/hope/apps/account/api/views.py
```diff
@@ -115,16 +115,13 @@ def get_queryset(self) -> QuerySet[User]:
         if role_ids_filter := self.request.query_params.getlist("roles"):
             role_assignments_queryset = role_assignments_queryset.filter(role__id__in=role_ids_filter)
 
-        role_assignment_ids = list(role_assignments_queryset.values_list("id", flat=True))
+        user_has_role = role_assignments_queryset.filter(user_id=OuterRef("id"))
```
johniak, line 116: in line 106 we are adding select related to role_assignments_queryset. We should not pass it into exists queries.

---
_in theory postgres should ignore unnecessary joins but it is not guaranted_

johniak (approved):

src/hope/apps/account/api/views.py
pkujawa: actually the select related in 106 was not beneficial at all. Removed

pavlo-mk (approved):

domdinicola (approved):

PR #5865, opened 2026-04-01
Change Pragma CI

ntrncic:
Added a comment in the auth.py file.

### DIFF
--- a/.github/workflows/pragma.yml
+++ b/.github/workflows/pragma.yml
@@ -1,9 +1,6 @@
 name: Pragma
 on:
   workflow_call:
-  pull_request:
-    paths:
-      - '**.py' # Only run on Python files
 
 jobs:
   block-pragma:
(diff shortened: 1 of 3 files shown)


.github/workflows/pragma.yml
```diff
@@ -1,5 +1,6 @@
 name: Pragma
 on:
+  workflow_call:
   pull_request:
```
johniak: This `pull_request` trigger should not be there.

  Please look at the example in lint.yml:
```
  name: LINT
  permissions:
    contents: read
  on:
    workflow_call:
  jobs:
```
  It should not be there because we are calling this workflow from ci-pr, which already has:
```
  on:
    pull_request:
      branches:
        - develop
        - staging
        - master
        - ephemeral-*
        - long-term/*
```

johniak (changes requested):

.github/workflows/pragma.yml
ntrncic: removed. Thank you.

johniak (approved):

PR #5883, opened 2026-04-08
Fix `calculate_age_at_registration`

pavlo-mk:
Fix `calculate_age_at_registration`

Usually we are using `calculate_age_at_registration ` to calculate `age_at_registration` and field is `age_at_registration = models.PositiveSmallIntegerField(null=True, blank=True, help_text="Age at registration [sys]")` so we can have any values below 0

[AB#309211](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/309211): Angola: Kobo issue "Import Error" - example was this issue with kobo import


tests/unit/apps/utils/test_age_at_registration.py
```diff
@@ -0,0 +1,16 @@
+from datetime import datetime
+
+from hope.apps.utils.age_at_registration import calculate_age_at_registration
+
+
+def test_calculate_age_at_registration() -> None:
```
johniak: Per project test conventions, one test should equal one scenario (Arrange-Act-Assert). This single test function covers three distinct scenarios (ValueError, future date, happy path). These should be split into three separate test functions or use pytest.mark.parametrize:

tests/unit/apps/utils/test_age_at_registration.py
```diff
@@ -0,0 +1,16 @@
+from datetime import datetime
```
johniak, line 1: Consider adding a test case for birth_date == created_at (same day), which should return 0. This is a meaningful edge case for the boundary check on L15 of the source (birth_date_parsed > created_at_date).

johniak (changes requested):

tests/unit/apps/utils/test_age_at_registration.py
domdinicola: +1

tests/unit/apps/utils/test_age_at_registration.py
domdinicola: agree

domdinicola (changes requested):

johniak (approved):

domdinicola:
@saxix does this make sense to you?

domdinicola:
@pavlo-mk should not raise a validation error?

pavlo-mk:
it's better to have ValueError because of the `parser.parse(birth_date)` in my opinion, ValidationError can be added if result of calculate_age_at_registration is less then 0

PR #5888, opened 2026-04-09
Django - 5.2.13, weasyprint 68

pkujawa:


pyproject.toml
```diff
@@ -211,6 +211,8 @@ django_settings_module = "hope.settings"
 
 [tool.uv]
 package = true
+# bleach is deprecated and caps tinycss2<1.5; weasyprint>=68 needs tinycss2>=1.5
```
domdinicola, line 206: shall we remove bleach then?

pkujawa: it's needed for `django-markdownify` which is used in changelog admin for description. Do we want to get rid of markdownify?

johniak: @domdinicola do we still use this changelog feature?

johniak (approved):

pkujawa:
Django was upgraded now in another PR, we do not want to upgrade weasyprint yet

PR #5889, opened 2026-04-09
initdemo - const area p_codes for faster local testing

pkujawa:


tests/extras/test_utils/old_factories/geo.py
```diff
@@ -70,11 +70,50 @@ def generate_p_code(prefix: str, count: int) -> List[str]:
     return [f"{prefix}{random.randint(10, 99)}" for _ in range(count)]
```
Jeeziorny, line 70: I thought that old_factories is a directory that will be removed after we fully transition to new style tests. Should we modify files inside this directory? If so, isn't the "old" prefix misleading? (For me it is) @pkujawa @johniak what's your take on this?

pkujawa: So I believe we should wait for the decision on what to do with old_factories Factories and helpers used in initdemo. Let's handle this separately

domdinicola (approved):
let's merge it and we can manage the initdemo separately

johniak (approved):

PR #5890, opened 2026-04-09
bug/ fix API calls to programs/null on Program Details page

mmaciekk:


src/frontend/src/containers/tables/ProgramCycle/NewProgramCycle/AddNewProgramCycle.tsx
```diff
@@ -2,50 +2,34 @@ import { ButtonTooltip } from '@components/core/ButtonTooltip';
...
 import { ProgramStatusEnum } from '@restgenerated/models/ProgramStatusEnum';
-import { useQuery, useQueryClient } from '@tanstack/react-query';
+import { useQueryClient } from '@tanstack/react-query';
 import { ReactElement, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { hasPermissions, PERMISSIONS } from 'src/config/permissions';
-import { useProgramContext } from 'src/programContext';
 
 interface AddNewProgramCycleProps {
+  program: ProgramDetail;
```
johniak, line 17: We made this prop mandatory but we still have usage (in rc/frontend/src/containers/tables/ProgramCyclesTablePaymentModule/ProgramCyclesTablePaymentModule.tsx) of AddNewProgramCycle which does not pass this props

src/frontend/src/containers/dialogs/programs/ActivateProgram.tsx
johniak: unused import when we remove usage from line 30

src/frontend/src/containers/dialogs/programs/ActivateProgram.tsx
```diff
@@ -37,10 +36,6 @@ export const ActivateProgram = ({
         code: program.code,
       }),
     onSuccess: () => {
-      setSelectedProgram({
-        ...selectedProgram,
-        status: ProgramStatusEnum.ACTIVE,
-      });
```
johniak, line 43: I'm not sure here, why we remove this part? Isn't it needed?

johniak (changes requested):

src/frontend/src/containers/tables/ProgramCycle/NewProgramCycle/AddNewProgramCycle.tsx
mmaciekk: fixed. passed prop in both cases

src/frontend/src/containers/dialogs/programs/ActivateProgram.tsx
mmaciekk: I don't think it is needed. It seems to be some stale code before     queryClient.invalidateQueries({
        queryKey: ['program', businessArea, program.code], was added. 
        
        Besides, we don't want to add selected program to global program filter if we are in All programs view

domdinicola (approved):

pkujawa (approved):

PR #5913, opened 2026-04-16
[AB#301335] Add Latin name fields for beneficiaries

mmaciekk:

Beneficiary names can be written in a local script (e.g. Cyrillic). This PR adds `*_latin` fields that store the same names in the Latin alphabet. The existing name fields are not changed.

## Changes

- **Model**: new optional fields `given/middle/family/full_name_latin` on `Individual`, only Latin letters allowed
- **Import**: RDI (XLSX/Kobo/People), the RDI push API and Aurora fill the Latin names automatically (transliteration) when they are not provided
- **Search**: searching by name also checks the Latin fields — individual & household lists, office search, grievance list, payment collector search. ES part is behind the `ES_USE_LATIN_NAMES` constance flag (needs fleet reindex first), plain DB search works right away
- **Grievance**: changing a name through a data-change ticket refreshes the Latin fields; adding an individual fills them too
- **API**: `*_latin` fields returned by individual/household/grievance/payment endpoints
- **UI**: Latin name shown in small font under the regular name
- **Backfill**: one-time script fills Latin names for existing individuals — goes business area by business area, program by program, in small batches (safe for ~30M rows), can be stopped and re-run (never overwrites filled values), bad records are skipped and written to a JSONL report

## Rollout order

1. deploy (flag OFF) → 2. run backfill script → 3. `es_reindex --all` → 4. turn on `ES_USE_LATIN_NAMES`

image
image
image


johniak (commented):
I think there might be an issue with this PR. It requires an Elasticsearch reindex due to adding new fields to the index, which brings us back to @Jeeziorny  blue-green ES migration.

If we merge this PR, we would need to rebuild the ES index after the data migration.
FYI
@ntrncic

src/hope/one_time_scripts/migrate_latin_names.py
```diff
@@ -0,0 +1,72 @@
...
+    Populate latin name fields for Individual model
+    """
+    total_processed = 0
+    total_updated = 0
+    started_at = time.time()
+
+    print("Starting latin name migration...")
+
+    qs = (
+        Individual.objects
```
johniak: We ignore here Pending individuals, so when we merge it it will have empty fields. we should propably make this on all_objects

johniak (changes requested):
Here I see a consistency issue.

The validator `^[A-Za-z]+(?:[ '-][A-Za-z]+)*$` allows spaces, apostrophes, and hyphens as separators between groups of letters. This means that names like `O'Brien` and `Jean-Pierre` are valid according to the validator.

However, `to_latin` from `household/utils` behaves differently:

- Line 24 removes all apostrophes: the regular apostrophe `'`, the backtick `` ` ``, and typographic apostrophes `’` and `‘`, replacing them with an empty string.
- Line 25 replaces everything that is not an ASCII letter or whitespace with a space. This includes hyphens `-`, digits, special characters, etc.
- Line 26 compresses multiple spaces into one and trims leading/trailing spaces.

So the question is: do we allow apostrophes, backticks, and hyphens, or not? At the moment, the validator suggests that we do, but `to_latin` removes or replaces them.

pavlo-mk:
@ntrncic @domdinicola about data migration I'm not sure how the library gonna translate middle-east names... after research it says better to use hard coded mapping for names... but it's tons of names in our system... do you think we can run it first in EPH with prod data to test?

@johniak FYI

Jeeziorny:
If it's blocked it's blocked (not opened)

johniak:
Superseded by #6375 (same branch continued as `301335/latin-names-no-transliteration`): after the 2026-09-02 decision to store `*_latin` names exactly as provided, automatic transliteration, the backfill script and the ticket checkbox are removed there. This branch is kept as-is for reference.

PR #5914, opened 2026-04-17
Admin fixes, improvements

pkujawa:
[AB#311032](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/311032)

Fix known issues in the admin:

- Autocomplete for all Foreign Keys in Admin (no dropdowns)
- Switch inlines to list links in buttons for:
  - Payment Plans - add link to Western union payment plan reports
  - Payment Plans - add Payments (currently not listed anywhere in PPlan)
  - Messages - move Household Recipients
- Household - add fields that are missing but are visible on UI. 
- find all models that are missing blank=True - it is impossible to save existing objects because fields are required in admin but should not be
- Payments - optimize
- no count for GRV list admin (timeouts)
- ApiLogEntry - fix search


domdinicola:
@pkujawa pls fix conflicts

src/hope/admin/utils.py
```diff
@@ -103,7 +103,34 @@ class HopeModelAdminMixin(ExtraButtonsMixin, SmartDisplayAllMixin, AdminActionPe
...
 
 
-class HOPEModelAdminBase(HopeModelAdminMixin, JSONWidgetMixin, admin.ModelAdmin[_ModelT]):
+class AutocompleteForeignKeyMixin:
+    """Automatically use autocomplete widgets for all ForeignKey (and M2M) fields.
+
+    Only fields where the related model's admin has ``search_fields`` defined
+    are included. Fields already handled by ``filter_horizontal`` /
+    ``filter_vertical`` are left untouched so they keep their multi-select widget.
+    """
```
MarekBiczysko, line 113: please add search_fields implementation as required (validation) in HOPEModelAdminBase and make sure that all admins inherit HOPEModelAdminBase, otherwise we still can end up with default (wrong)  behavior when need admin page is created

MarekBiczysko (changes requested):

MarekBiczysko (approved):

src/hope/admin/utils.py
pkujawa: As discussed, instead of the validation that enforces `search_fields`, we decided to use `raw_id` fields in case of missing `search_fields` on the related admin model. Updated

pavlo-mk (approved):

PR #5917, opened 2026-04-20
Migrations regeneration api types precommit

Jeeziorny:
Recently I got a few issues bc I forgot to commit django migration files and/or REST API Types. Changes on this branch introduce precommit checks:
1. If there are django migration that should be added to the index but are not,
2. If there are REST API Types regeneration needed but the commit doesn't have it.


scripts/precommit_check_fe_types.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
domdinicola, line 1: interesting we need to check FE types...

domdinicola: what's the reason @mmaciekk

scripts/precommit_check_migrations.sh
```diff
@@ -0,0 +1,37 @@
...
+# producing spurious "Alter field id" migrations that would fail --check on CI.
+#
+# Convention: the app label equals the last path segment of the app's directory.
+# This matches every first-party apps.py we checked (e.g. `account/apps.py` sets
+# `label = "account"`; `payment/apps.py` has `name = "hope.apps.payment"` which
+# derives label `payment`). If a future app overrides `label` to something other
+# than its directory name, add an explicit mapping here.
+
+set -euo pipefail
+
```
domdinicola: I think we can reuse something

repos:
  - repo: local
    hooks:
      - id: django-migrations-check
        name: Check for missing Django migrations
        entry: python manage.py makemigrations --check --dry-run
        language: system
        pass_filenames: false

- repo: https://github.com/ecugol/pre-commit-hooks-django
  rev: v0.4.0
  hooks:
    - id: check-absent-migrations

@saxix ?

domdinicola (changes requested):

scripts/precommit_check_fe_types.sh
Jeeziorny: I don't understand what action this comment require

domdinicola: what is the idea of this hook?

MarekBiczysko: as staded in the code comment below:

This catches the common "forgot to regen FE types after BE filter/serializer
change" mistake

mmaciekk: Whenever the backend API structure changes, the frontend types must be regenerated so TypeScript reflects the latest schema.
Otherwise, the frontend may still assume outdated fields exist, and TypeScript won’t catch the mismatch.
For example, if individual.favoriteColor is removed on the backend but types aren’t updated, the frontend will still use it—leading to runtime errors or crashes.

johniak: Hi @domdinicola 
We write the frontend in TypeScript, which is basically typed JavaScript.

We use a type generation command based on the OpenAPI schema created by the backend for two purposes:

- To generate TypeScript types, so every request and response body is typed. This helps reduce mistakes.
- To generate React hooks for @tanstack/react-query with typed request and response bodies already in place. This speeds up development because we do not have to write API calls manually, and it also helps prevent mistakes such as sending the wrong request body to the wrong query.

This PR adds a pre-commit check to ensure that whenever an API endpoint is changed, the corresponding types and React hooks are regenerated correctly.

scripts/precommit_check_migrations.sh
Jeeziorny: Thanks — I looked at both. ecugol/pre-commit-hooks-django runs makemigrations --check unscoped, same as the plain local hook. That's what broke CI before this PR: third-party apps
(admin, advanced_filters, …) produce spurious "Alter field id" migrations and fail --check. 

The script here scopes the check to first-party app labels discovered from src/hope/**/apps.py
to filter those out. Happy to swap to a community hook if we find one that supports an app-label allowlist, but I don't see that option in either suggestion.

@domdinicola

scripts/precommit_check_migrations.sh
```diff
@@ -0,0 +1,37 @@
...
+while IFS= read -r apps_py; do
+    # Strip the leading "src/hope/" and trailing "/apps.py", keep the last segment.
+    rel="${apps_py#src/hope/}"
+    dir="${rel%/apps.py}"
+    label="${dir##*/}"
+    if [[ -z "${seen[$label]:-}" ]]; then
+        seen[$label]=1
+        labels+=("$label")
+    fi
+done < <(find src/hope -type f -name apps.py | sort)
```
saxix: this would require another check to be sure that `apps.py` is always present

scripts/precommit_check_migrations.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
```
saxix: this seems to check for missing "created" migrations, not that the migrations have been added to git.

saxix (changes requested):

scripts/precommit_check_migrations.sh
Jeeziorny: Good catch, thanks. I added python code to gather only hope's apps first:
```python
  labels = sorted({
      a.label
      for a in apps.get_app_configs()
      if a.name == 'hope' or a.name.startswith('hope.')
  })
```

scripts/precommit_check_migrations.sh
```diff
@@ -0,0 +1,44 @@
...
```
saxix: why are we using bash here? Using python seems more appropriate

saxix (changes requested):

Jeeziorny:
@saxix reminder

saxix (approved):

PR #5926, opened 2026-04-22
303284: Add country_workspace_id field to Individual model and update…

arsen-vs:
… related tests

### DIFF
--- a/src/hope/apps/household/migrations/0040_migration.py
+++ b/src/hope/apps/household/migrations/0040_migration.py
@@ -1,0 +1,24 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("household", "0039_migration"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="individual",
+            name="country_workspace_id",
+            field=models.CharField(
+                blank=True,
+                db_index=True,
+                help_text=(
+                    "Primary key of the Individual on the originating Country Workspace system. "
+                    "Used as the reference key when communicating with the Deduplication Engine."
+                ),
+                max_length=150,
+                null=True,
+            ),
+        ),
+    ]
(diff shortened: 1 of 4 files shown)


tests/unit/api/test_lax_individuals.py
```diff
@@ -134,6 +134,7 @@ def test_create_single_individual_success(lax_api_client, lax_push_url, document
             }
         ],
         "originating_id": "AUR#123#123",
+        "country_workspace_id": "42",
```
Jeeziorny, line 137: Can you please add number of queries assertion for this test?

tests/unit/api/test_push_people.py
```diff
@@ -409,3 +409,30 @@ def test_people_serializer_disability(common_serializer_data, disability_value,
...
+            "country": "AF",
+            "full_name": "John Doe",
+            "birth_date": "2000-01-01",
+            "sex": "MALE",
+            "type": "",
+            "program": str(program.id),
+            "country_workspace_id": "42",
+        }
+    ]
+    response = token_api_client.post(push_people_url, data, format="json")
```
Jeeziorny, line 430: SAA

Jeeziorny:
Don't we need to have serializer changes here?

Jeeziorny (changes requested):
1. PushPeopleSerializer and lax:IndividualSerializer both needs to have required param country_workspace_id. Without it we won't be able to match the results in HOPE.
2. country_workspace_id should be unique from what I understand. In such case pls list this field as partial unique in model we're changing here.
```python
  models.UniqueConstraint(
      fields=["country_workspace_id"],
      condition=Q(country_workspace_id__isnull=False),
      name="uniq_country_workspace_id",
  )
```

src/hope/apps/household/migrations/0040_migration.py
```diff
@@ -0,0 +1,24 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("household", "0039_migration"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="individual",
+            name="country_workspace_id",
+            field=models.CharField(
```
Jeeziorny, line 13: Shouldn't this one be numeric type? 

Image

Source: https://dedup-hope-dev.unitst.org/api/rest/swagger/#/deduplication_set_groups/deduplication_set_groups_findings_list

PR #5927, opened 2026-04-22
[AB#311718]-fix timezone issues for Americas

mmaciekk:
[AB#311718](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/311718)

⏺ **Summary**                                                                                                           
                                                                                                                  
  Fixed a timezone bug where objects with date attributes displayed/saved one day behind for users in negative-UTC  
  timezones (Haiti, Colombia, …). Root cause: date-only strings ("YYYY-MM-DD") were being parsed as UTC midnight,   
  which renders as the previous local day in UTC−N.                                                                 
    
  **Files changed**                                                                                                   

  _src/frontend/src/components/core/UniversalMoment.tsx_                                                              
  Table/detail renderer for dates. Replaced new Date(children) (UTC for date-only strings) with parseISO(children)
  when the input matches YYYY-MM-DD, so the stored calendar day is preserved.                                       
    
  _src/frontend/src/components/core/DatePickerFilter.tsx_                                                             
  List filter date picker. Dropped the T00:00:00.000Z suffix that was forcing UTC midnight; now calls
  parseISO(value) directly, yielding local midnight.                                                                
    
  _src/frontend/src/shared/Formik/FormikDateField/FormikDateField.tsx_                                                
  Form date picker. When a form value comes back as a full datetime (e.g. Django DateTimeField serialized as
  "…T00:00:00Z"), strip down to the YYYY-MM-DD portion before parseISO, so the picker always displays the backend's 
  intended calendar day.
                                                                                                                    
  _src/frontend/src/utils/utils.ts — dateToIsoString_                                                                 
  Filter range helper. Switched moment.utc(date).startOf('day')/.endOf('day') to                                  
  moment(date).startOf('day')/.endOf('day') so ranges align with the user's local calendar day (previously the last 
  local hours of a day were excluded for UTC−N users).


ntrncic:
write a PR description of the solution, not the problem.

pavlo-mk (approved):

johniak (approved):

PR #5928, opened 2026-04-22
AB#311326 skip unparseable DOBs and record them on individual

johniak:
[AB#311326](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/311326)
## Summary
  - Add `internal_data` JSONField on `SanctionListIndividual` for system-managed ingestion metadata.
  - UN + EU strategies: on unparseable birthdate, skip only that DOB entry and append `{raw, type}` to
  `internal_data["date_of_birth_parse_errors"]`; the individual is still created/updated.
  - Harden UN `BETWEEN` branch and `EXACT`/`APPROXIMATELY` path against missing tags.


pavlo-mk (approved):

src/hope/models/sanction_list_individual.py
```diff
@@ -34,7 +34,12 @@ def hard_delete(self) -> tuple[int, dict[str, int]]:
         return self.get_queryset().hard_delete()
 
 
-class SanctionListIndividual(TimeStampedUUIDModel):
+class SanctionListIndividual(TimeStampedUUIDModel, InternalDataFieldModel):
+    internal_data = models.JSONField(
```
MarekBiczysko: why did you add this field again, it's already in InternalDataFieldModel

MarekBiczysko (changes requested):

domdinicola (approved):

domdinicola (changes requested):
Check @MarekBiczysko 's comment

src/hope/apps/sanction_list/strategies/eu.py
```diff
@@ -123,10 +124,21 @@ def parse(self, root: Element) -> int:
...
                     date_data: dict[str, Any] = date
+                    try:
+                        datetime.strptime(date["date"], "%Y-%m-%d")
+                    except ValueError:
+                        logger.exception("Cannot parse birthdate %r for %s", date["date"], ind.reference_number)
+                        errors = ind.internal_data.setdefault("date_of_birth_parse_errors", [])
+                        errors.append({"raw": date["date"], "type": "EU"})
+                        invalid_dob_added = True
+                        continue
                     ind.dates_of_birth.get_or_create(**date_data)
```
domdinicola, line 138: We do create even with invalid date?

src/hope/apps/sanction_list/strategies/eu.py
```diff
@@ -123,10 +124,21 @@ def parse(self, root: Element) -> int:
...
+            if invalid_dob_added:
+                ind.save(update_fields=["internal_data"])
```
domdinicola, line 141: Can't we put the save in the except?

Jeeziorny: except block should be as small as possible to catch errors that are expected, don't agree with you @domdinicola

src/hope/apps/sanction_list/strategies/un.py
```diff
@@ -111,37 +111,65 @@ def _get_date_of_births(
...
+                            dob_type=type_of_date,
                         )
-                        for year in range(int(from_year), int(to_year) + 1)
-                    }
-                    dates_of_birth.update(years)
 
         return dates_of_birth
 
+    @staticmethod
+    def _record_dob_parse_error(
```
Jeeziorny: how about reusing this helper static method in src/hope/apps/sanction_list/strategies/eu.py:135-136?

If I missed something pls resolve

tests/unit/apps/sanction_list/test_un_load_xml.py
```diff
@@ -89,3 +89,26 @@ def test_execute(sanction_list: "SanctionList", program: "Program") -> None:
...
+@pytest.mark.elasticsearch
+def test_invalid_dates_of_birth_are_recorded_in_internal_data(
+    sanction_list: "SanctionList", program: "Program"
+) -> None:
+    main_test_files_path = Path(__file__).parent / "test_files"
+
+    task = LoadSanctionListXMLTask(sanction_list)
+    task.load_from_file(main_test_files_path / "broken-dob-consolidated.xml")
+
+    assert SanctionListIndividual.all_objects.count() == 1
```
Jeeziorny, line 104: Looks like we check the db state after the test. WDYT about adding assert num queries here as well?

nerjan (approved):

src/hope/models/sanction_list_individual.py
johniak: Good catch, thanks — it was a leftover from before I added InternalDataFieldModel to the inheritance. Removed the redeclaration and dropped help_text from the migration (the base class doesn't carry one, and none of the other models  using this mixin do either). Resolved in .

src/hope/apps/sanction_list/strategies/eu.py
johniak: I think this was exactly what we decided at the daily that we ignore broken dob, but individual is still saved with information that dob was bad

src/hope/apps/sanction_list/strategies/eu.py
johniak: I'd prefer to keep a single save(update_fields=["internal_data"]) after the loop — if an entry has N unparseable birthdays, putting it inside the except would trigger N UPDATEs instead of one. The invalid_dob_added flag is just to skip the save when there's nothing to persist.

src/hope/apps/sanction_list/strategies/un.py
johniak: Good idea — moved the helper onto the model

tests/unit/apps/sanction_list/test_un_load_xml.py
johniak: Added django_assert_num_queries(25)

Jeeziorny (approved):
I re-reviewed that with claude and seems that the only nice to have thing test case for UN: 

> test_bad_dob_does_not_poison_other_individuals: a fixture with two <INDIVIDUAL> blocks (A with a malformed BETWEEN range, B with a valid EXACT date), then assert SanctionListIndividual.all_objects.count() == 2 and B.dates_of_birth.count() == 1


Otherwise, lgtm.

MarekBiczysko (approved):

tests/unit/apps/sanction_list/test_un_load_xml.py
```diff
@@ -113,3 +113,26 @@ def test_invalid_dates_of_birth_are_recorded_in_internal_data(
...
+
+
+@pytest.mark.elasticsearch
+def test_bad_dob_does_not_poison_other_individuals(
+    sanction_list: "SanctionList", program: "Program", django_assert_num_queries: Any
+) -> None:
+    main_test_files_path = Path(__file__).parent / "test_files"
+
+    task = LoadSanctionListXMLTask(sanction_list)
+    with django_assert_num_queries(27):
```
Jeeziorny, line 125: 🥇

PR #5930, opened 2026-04-23
Admin smoke tests

pkujawa:
[AB#310011](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/310011)

New file: tests/unit/admin/test_admin_smoke.py                      
Created based on template: [https://github.com/unicef/hope-cookiecutters/blob/main/django-project/%7b%7bcookiecutter.project_dir%7d%7d/tests/unit/admin/test_admin_smoke.py](https://github.com/unicef/hope-cookiecutters/blob/main/django-project/%7b%7bcookiecutter.project_dir%7d%7d/tests/unit/admin/test_admin_smoke.py)

Adds automated Django admin smoke tests that exercise every registered model admin — changelist, changeform, add, delete, and extra button views — verifying HTTP responses without errors.Key implementation details:
  - get_factory_for_model implemented locally using @functools.cache — scans extras.test_utils.factories on first call, builds a model→factory registry, and picks the shortest-named factory when multiple exist for the same model                                                                                                                      
  - ROOT_TOKEN injected into the test client and _mock_request() so HOPE's is_root() checks pass in all admin views
  - GLOBAL_EXCLUDED_BUTTONS covers all buttons requiring live external connections (Azure AD, RapidPro, Aurora, sanction list)                                                                                                                                                                                              
  - GLOBAL_EXCLUDED_MODELS covers third-party/infrastructure models with no HOPE business logic                                                                        
                                                                                                                                                
 Several bugs were discovered by the smoke tests and fixed in this PR:
 - `reset_sync_date` on all admin models - changelist button was permanently inaccessible (403) - permission was passing obj, while obj=None for all changelist-level buttons
 - `linked_objects` buttons for multiple admin pages — lambda had wrong signature and permission was always granted
-  several models had the add view enabled but the add form would crash, as all meaningful fields are read-only, or records are exclusively created programmatically and never manually. Blocked with `has_add_permission=False`
- Individual admin - `household_members` button crashed on individuals without a household     
- Project - `Add Project page` crushing on get_form - obj was None but obj.organization.business_area was accessed unconditionally          
- added early return guards and safe attribute access wherever a missing FK or absent cleaned_data key would cause AttributeError, KeyError, or RelatedObjectDoesNotExist crashes                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
- `confirm_action` — was using positional instead of keyword arguments, causing TypeError - `gdpr_remove`, `logical_delete`, `mark_submissions`, `reset_sync_date`, and `reset_sync_date_single` would all show a 500 error on click rather than the confirmation dialog

New factories added to support the smoke tests


nerjan (approved):

src/hope/admin/individual.py
```diff
@@ -206,6 +206,8 @@ def formfield_for_foreignkey(self, db_field: Any, request: HttpRequest, **kwargs
     def household_members(self, request: HttpRequest, pk: UUID) -> HttpResponseRedirect:
         obj = Individual.all_merge_status_objects.get(pk=pk)
         url = reverse("admin:household_individual_changelist")
+        if obj.household is None:
+            return HttpResponseRedirect(url)
```
saxix, line 211: add message here

src/hope/admin/payment_plan.py
```diff
@@ -234,6 +234,8 @@ def regenerate_export_xlsx(self, request: HttpRequest, pk: "UUID") -> HttpRespon
     def related_configs(self, request: HttpRequest, pk: "UUID") -> HttpResponse:
         obj = PaymentPlan.objects.get(pk=pk)
         url = reverse("admin:payment_deliverymechanismconfig_changelist")
+        if not obj.delivery_mechanism or not obj.financial_service_provider:
```
saxix, line 232: add message here

saxix (changes requested):

src/hope/models/program.py
```diff
@@ -344,6 +340,9 @@ def __str__(self) -> str:
         return self.name
 
     def validate_unique(self, exclude: Collection[str] | None = ...) -> None:  # type: ignore
+        if not self.business_area_id:
+            super().validate_unique()
```
johniak: I think we should still pass exclude here

src/hope/models/program.py
johniak: I think we should still pass exclude here

tests/unit/admin/test_admin_smoke.py
```diff
@@ -0,0 +1,269 @@
...
+        r"payment.PaymentVerificationPlanAdmin:execute_sync_rapid_pro",  # RapidPro external
+        r"sanction_list.SanctionListAdmin:refresh",  # fetches external sanction list
+        # external Aurora service
+        r"aurora.RecordAdmin:fetch_aurora",
+        r"aurora.RecordAdmin:fetch",
+        r"aurora.RecordAdmin:extract_single",
+    ]
+)
+
+KWARGS: dict = {}
```
johniak, line 73: Is this used anywhere?

johniak (changes requested):

src/hope/models/program.py
pkujawa: Good point. Seems we never did.

tests/unit/admin/test_admin_smoke.py
pkujawa: The variable is used - 
`raw_kwargs = KWARGS.get(full_name, {})`
also added keyword handling 
`resolved_kwargs = {k: v() if callable(v) else v for k, v in raw_kwargs.items()}`

But you're right - currently it was not necessary for any model to use it. (Actually I used it but eventually handled the issues through factory or admin fixes). I think it's worth leaving so we can use it when factory data is not enough.

But I can remove it if you disagree

pavlo-mk (approved):

johniak (approved):

src/hope/admin/individual.py
saxix: I expect something like `self.message_user(request, "No HH for this individual")

src/hope/admin/payment_plan.py
saxix: see previous comment

saxix (changes requested):

src/hope/admin/individual.py
pkujawa: @saxix

Image

This is the current code. Is this not fine?

I see there are some issues on github. Old snapshot is showing above comment but no Outdated tag. Please check these places in the "Files changed" tab

src/hope/admin/payment_plan.py
pkujawa: Image
this was also addressed

saxix (approved):

PR #5932, opened 2026-04-23
306141_PG_Integration_send_all_records_not_only_API_TYPE_this_should_…

MarekBiczysko:
…be_feature_flagged

[AB#306141](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/306141)

Adds an admin-managed use_payment_gateway flag on PaymentPlan to allow sending payment plans through Payment Gateway even when the selected FSP is not API.

Payment plan gateway behavior is now computed from:
payment_plan.use_payment_gateway OR payment_plan.financial_service_provider.communication_channel == API

This makes flagged XLSX/manual FSP plans behave like gateway plans while preserving normal manual/XLSX flow for unflagged plans.

  - Added PaymentPlan.use_payment_gateway field and migration.
  - Exposed the flag in Django admin list/filter.
  - Added uses_payment_gateway to payment plan detail API responses.
  - Removed frontend dependency on raw FSP communication channel for gateway/manual decisions.
  - Disabled manual reconciliation upload for gateway-enabled payment plans.
  - Disabled manual mark-as-failed / revert for gateway-enabled payment plans.
  - Updated payment gateway send/sync logic to include flagged non-API plans.
  - Made follow-up payment plans inherit the flag from the source payment plan.
  - Removed old serializer fields tied to raw FSP gateway checks.
  - Updated API contract fixtures and payment-related unit tests.


Jeeziorny:
I like the branch name

pavlo-mk (approved):

domdinicola:
@MarekBiczysko @johniak I think we should send if type is API  OR check use_payment_gateway flag for now

MarekBiczysko:
> @MarekBiczysko @johniak I think we should send if type is API OR check use_payment_gateway flag for now

what do you mean, please elaborate

domdinicola:
@MarekBiczysko i see now in the description of PR
payment_plan.use_payment_gateway OR payment_plan.financial_service_provider.communication_channel == API


pls raise coverage

src/hope/apps/payment/api/serializers.py
```diff
@@ -600,6 +597,7 @@ class PaymentPlanCreateFollowUpSerializer(serializers.Serializer):
 class PaymentPlanDetailSerializer(AdminUrlSerializerMixin, PaymentPlanListSerializer):
     background_action_status_display = serializers.CharField(source="get_background_action_status_display")
     program_cycle = ProgramCycleSmallSerializer()
+    uses_payment_gateway = serializers.BooleanField(source="is_payment_gateway")
```
nerjan: you have added `uses_payment_gateway` and here uses_payment_gateway is actually `is_payment_gateway` while also having `use_payment_gateway` is that intentional? Quite misleading, why not just is_payment_gateway?

src/hope/apps/payment/services/payment_gateway.py
```diff
@@ -133,7 +133,7 @@ def _map_financial_institution(self, obj: Payment, account_data: dict) -> dict:
                 account_data["service_provider_code"] = fsp_mapping.code
 
             except FinancialInstitutionMapping.DoesNotExist:
-                raise Exception(
+                logger.error(
```
nerjan, line 136: This and the further change makes this function to never throw an error but actually to return the data as if everything was ok. Is that intentional?

nerjan (approved):

src/hope/apps/payment/services/payment_gateway.py
MarekBiczysko: yes, we don't want to block send action if mapping doesnt exists, just send all existing data

PR #5945, opened 2026-04-28
[LONG-TERM] Payments group and purpose

pkujawa:


scripts/precommit_check_migrations.sh
nerjan, line 1: this change is stale right?

src/hope/apps/payment/api/views.py
```diff
@@ -1642,6 +1558,263 @@ class PaymentPlanGlobalViewSet(
...
+        "create": [Permissions.PM_CREATE],
+        "lock": [Permissions.PM_LOCK_AND_UNLOCK],
+        "unlock": [Permissions.PM_LOCK_AND_UNLOCK],
+        "lock_fsp": [Permissions.PM_LOCK_AND_UNLOCK_FSP],
+        "unlock_fsp": [Permissions.PM_LOCK_AND_UNLOCK_FSP],
+        "send_for_approval": [Permissions.PM_SEND_FOR_APPROVAL],
+        "approve": [Permissions.PM_ACCEPTANCE_PROCESS_APPROVE],
+        "authorize": [Permissions.PM_ACCEPTANCE_PROCESS_AUTHORIZE],
+        "mark_as_released": [Permissions.PM_ACCEPTANCE_PROCESS_FINANCIAL_REVIEW],
+        "delivery_export_xlsx": [Permissions.PM_VIEW_LIST],
```
nerjan, line 1600: I think we should have separate permission for that.

nerjan (dismissed):

src/hope/models/payment_plan_group.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+
+    class Meta:
+        app_label = "payment"
+        verbose_name = "Payment Plan Group"
+        unique_together = ("cycle", "name")
+        ordering = ["created_at"]
+
+    def delete(self, *args: object, **kwargs: object) -> tuple[int, dict]:
+        if self.cycle.payment_plan_groups.count() == 1:
```
saxix: fix TOCTOU here

saxix (changes requested):
- please add documentation

src/hope/apps/payment/api/views.py
nerjan: @MarekBiczysko

src/hope/models/payment_plan_group.py
nerjan: it should be fixed right now.

PR #5946, opened 2026-04-28
Remove old_factories

pavlo-mk:
[AB#309335](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/309335): decommission old_factories

Remove old_factories, this PR is not refactoring of e2e, just removed old factories and made e2e works FYI


domdinicola (approved):

tests/extras/test_utils/fixtures/dashboard.py
```diff
@@ -27,22 +30,20 @@ def _populate_dashboard_cache(
...
-                    "male_age_group_60_disabled_count": 1,
-                    "admin1": AreaFactory(
-                        name="Kabul",
-                        area_type__name="Province",
-                        area_type__area_level=1,
-                    ),
-                    "program": program,
-                    **(household_extra_args or {}),
-                }
+            IndividualFactory(household=None, business_area=afghanistan, program=program)
```
kosior: The ModifiedPaymentFactory is not being created for an individual, so do we need to create the individual here?

tests/extras/test_utils/factories/payment.py
```diff
@@ -6,6 +6,7 @@
 from django.utils import timezone
 import factory
 from factory.django import DjangoModelFactory
+from pytz import utc
```
kosior: I'd use `from datetime import UTC`.  Basically pytz has been replaced with zoneinfo, Django no longer uses pytz, so maybe in the future we could get rid of pytz entirely.

src/hope/apps/core/management/commands/initdemo.py
```diff
@@ -64,42 +67,1486 @@
...
+    PartnerFactory(name="UNHCR")
+    PartnerFactory(name="WFP")
+
+
+def generate_country_codes() -> None:
+    for country in Country.objects.all():
+        CountryCodeMap.objects.get_or_create(country=country, defaults={"ca_code": country.iso_code3})
+
+
+def generate_business_areas() -> None:
```
kosior: I'd suggest extracting the helper functions to a different place if they are being used in more than one place. It would shrink this large file.

tests/extras/test_utils/factories/geo.py
```diff
@@ -36,3 +39,79 @@ class Meta:
     name = factory.Sequence(lambda n: f"Area {n}")
     p_code = factory.Sequence(lambda n: f"PC{n:06d}")
     area_type = factory.SubFactory(AreaTypeFactory)
+
```
nerjan, line 41: This is duplicated with src/hope/apps/geo/management/commands/init_geo_fixtures.py, we could use these in the command.

tests/unit/apps/payment/test_currency_fk_services.py
```diff
@@ -144,7 +144,7 @@ def test_validate_transfer_to_digital_wallet_rejects_usdc_with_non_digital(
 
     from hope.apps.payment.services.payment_plan_services import PaymentPlanService
 
-    with django_assert_num_queries(83):
+    with django_assert_num_queries(86):
         dm = DeliveryMechanismFactory(transfer_type=DeliveryMechanism.TransferType.CASH)
         pp = PaymentPlanFactory(currency=currency_usdc, delivery_mechanism=dm)
         service = PaymentPlanService(pp)
```
nerjan: its a great example why its very bad to do with django_assert_num_queries on a bigger part of tests that only tested  feature - we have change in query count only cause we changed factories.
Move `with` lower so it handles service only, here and 2 below.

nerjan: I have actually changes it but its in LONGTERM branch. You could do
with django_assert_num_queries(0), pytest.raises(ValidationError, match="Transfer to Digital Wallet") even.

src/hope/apps/core/management/commands/initdemo.py
nerjan: Yes, definitely we should extract it, especially that it will grow with each new functionality we add. Also some of these functions are imported in another command its better to do some helper storing these. Or even better - separate directory with files that encapsulate logic separately, if it makes sens.

nerjan (changes requested):

tests/extras/test_utils/fixtures/dashboard.py
pavlo-mk: updated

src/hope/apps/core/management/commands/initdemo.py
pavlo-mk: extracted in `core.management.commands.demo_data`

tests/e2e/filters/test_filters.py
```diff
@@ -279,17 +276,17 @@ def create_rdi() -> None:
...
     beneficiary_group = BeneficiaryGroup.objects.filter(name="Main Menu").first()
-    ProgramFactory(
+    program = ProgramFactory(
         name="Test Programm",
         status=Program.ACTIVE,
         business_area=business_area,
         data_collecting_type=dct,
         beneficiary_group=beneficiary_group,
     )
+    ProgramCycleFactory(program=program)
```
kosior: I've seen creation of ProgramCycle couple times usually after calling ProgramFactory. I'm wondering if it would make sense to add this as a RelatedFactory:

```
class ProgramFactory(DjangoModelFactory):
    class Meta:
        model = Program

    ...
    cycle = factory.RelatedFactory(
        "extras.test_utils.factories.ProgramCycleFactory",
        factory_related_name='program',
    )
```

Then it would be created automatically - similar to what we do in the ProgramViewSet.

Also in the old_factories ProgramFactory had some additional functionality like the cycle @factory.post_generation, is it ok to remove all of that?

nerjan (approved):

tests/extras/test_utils/factories/program.py
```diff
@@ -29,12 +29,18 @@ class Meta:
     business_area = factory.SubFactory(BusinessAreaFactory)
     biometric_deduplication_enabled = False
 
+    @factory.post_generation
+    def cycle(self, create, extracted, **kwargs):
+        if not create or extracted is False:
+            return
+        ProgramCycleFactory(program=self)
```
kosior: This is something that we had previously:
```suggestion
        ProgramCycleFactory(program=self, **kwargs)
```

if we pass kwargs we should be able to do:
```
program = ProgramFactory(
        cycle__title="Default Programme Cycle",
        cycle__status=program_cycle_status,
        cycle__start_date=cycle_start_date,
        cycle__end_date=None,
)
```

kosior (approved):

src/hope/apps/core/management/commands/initdemo.py
```diff
@@ -64,37 +64,33 @@
...
-    generate_feedback,
-    generate_messages,
-)
-from extras.test_utils.old_factories.aurora import generate_aurora_test_data
-from extras.test_utils.old_factories.core import (
+from extras.test_utils.factories.account import create_superuser
+from extras.test_utils.factories.geo import generate_area_types
+from hope.apps.core.management.commands.demo_data.accountability import generate_feedback, generate_messages
+from hope.apps.core.management.commands.demo_data.aurora import generate_aurora_test_data
+from hope.apps.core.management.commands.demo_data.core import (
```
pkujawa, line 71: there is `demo_data` but this module was not added. Probably missed in committing?

tests/extras/test_utils/factories/__init__.py
```diff
@@ -173,6 +174,27 @@
     "GrievanceDocumentFactory",
     "GrievanceTicketFactory",
     "HouseholdCollectionFactory",
+    "GrievanceComplaintTicketWithoutExtrasFactory",
```
pkujawa: some of the new entries already existed and are duplicated - like `GrievanceComplaintTicketWithoutExtrasFactory`

src/hope/apps/core/management/commands/generatefixtures.py
```diff
@@ -121,17 +117,20 @@ def _generate_program_with_dependencies(options: dict, business_area_index: int)
...
+                    household=None,
+                    business_area=business_area,
+                    program=program,
+                    registration_data_import=registration_data_import,
+                )
+                household = HouseholdFactory(
+                    registration_data_import=registration_data_import,
+                    business_area=business_area,
+                    admin1=Area.objects.filter(area_type__business_area=business_area).order_by("?").first(),
+                    program=program,
```
pkujawa, line 131: the hoh is never assigned as a hoh to this household

tests/extras/test_utils/factories/geo.py
```diff
@@ -36,3 +38,79 @@ class Meta:
...
+            ("County", 3),
+            ("Region", 4),
+            ("District", 5),
+        ]:
+            parent = AreaTypeFactory(parent=parent, name=area_type_name, country=country, area_level=level)
+
+
+def generate_p_code(prefix: str, count: int) -> list[str]:
+    """generate a list of unique random p-codes with a given prefix."""
+    return [f"{prefix}{random.randint(10, 99)}" for _ in range(count)]
```
pkujawa: btw maybe   `return [f"{prefix}{n}" for n in random.sample(range(10, 100), count)]` so we don't hit duplicate

PR #5958, opened 2026-05-04
celery_recovery_timing_windows

MarekBiczysko:
- Celery missing-job recovery windows are now derived from CELERY_BROKER_VISIBILITY_TIMEOUT instead of a hardcoded 4h05m-12h range. The effective visibility timeout is also enforced to be at least task time limit
    plus 15 minutes.
  - Async Job admin was improved to match that logic: the “missing in Celery” filter now shows a dynamic recoverable window, and the changelist filters/autocompletes were tightened to only show values actually used
    by AsyncJob.


pavlo-mk (approved):

johniak (approved):

src/hope/admin/async_job.py
```diff
@@ -40,23 +61,125 @@ def queryset(self, request: HttpRequest, queryset: QuerySet[AsyncJob]) -> QueryS
...
+    parent = "program__business_area"  # type: ignore[assignment]
+    autocomplete_view_name_suffix = "used_program_autocomplete"
+
+
+class UsedContentTypeListFilter(UsedValuesListFilter):
+    title = "content type"
+    parameter_name = "content_type__exact"
+
+    def lookups(self, request: HttpRequest, model_admin: admin.ModelAdmin[Any]) -> list[tuple[str, str]]:
+        used_content_type_ids = (
```
nerjan, line 117: .distinct() seems like a good idea here

nerjan (approved):

ntrncic:
write a description please.

src/hope/admin/async_job.py
```diff
@@ -102,7 +105,7 @@ class UsedBusinessAreaAutoCompleteFilter(AsyncJobAdminLinkedAutoCompleteFilter):
 
 
 class UsedProgramAutoCompleteFilter(AsyncJobAdminLinkedAutoCompleteFilter):
-    parent: str | None = "program__business_area"
+    parent = "program__business_area"  # type: ignore[assignment]
```
domdinicola: can't we set a type?

PR #5990, opened 2026-05-11
Add search by Facility Name

pavlo-mk:
[AB#313059](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/313059): Search by Facility_Name

Added search by Facility Name page Household list view
image


src/frontend/src/containers/tables/population/HouseholdTable/HouseholdTableHeadCells.tsx
```diff
@@ -68,7 +68,7 @@ export const headCells: HeadCell<HouseholdList>[] = [
   {
     disablePadding: false,
     label: 'Facility Name',
-    id: 'facilityName',
+    id: 'facility__name',
```
MarekBiczysko, line 71: ordering doesn't work in all columns with snakecase id (ex. head of housold etc)

MarekBiczysko (changes requested):

nerjan (dismissed):

src/hope/apps/household/filters.py
```diff
@@ -226,6 +228,7 @@ def _search_db(self, qs: QuerySet[Household], value: str, program: Program | Non
                     | Q(phone_no_alt_normalized__icontains=search)
                     | Q(detail_id__icontains=search)
                     | Q(program_registration_id__icontains=search)
+                    | Q(facility__name__icontains=search)
```
Jeeziorny, line 232: Hi, we're modifying both DB and ES branches. Why is that? Are we taking care of the case when ES is down here, do I understand it correctly?

Jeeziorny (dismissed):
Just wondering: do we want to have separate FE field for that search? What's your take on this, @pavlo-mk, @nerjan

src/hope/apps/household/filters.py
pavlo-mk: yes we have to cover both cases

pavlo-mk:
> Just wondering: do we want to have separate FE field for that search? What's your take on this, @pavlo-mk, @nerjan

in the task wasn't any information about new separate search input for that
@Jeeziorny

pkujawa (approved):

Jeeziorny (approved):

PR #5994, opened 2026-05-12
feat: Add currency endpoint

roma-valor:
[AB#314285](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/314285)

Add currency endpoint list for hope


MarekBiczysko (approved):

Jeeziorny (changes requested):
Pls add assertion for number of queries in tests. Otherwise LGTM

Jeeziorny:
Today I woke up and realised that I don't get the point of this PR. If I introduced an interface change when moving currencies to database model, shouldn't we fix the interface to match the state from before that change? My understanding is that this PR tries to solve the problem, that should be resolved in other way, by simply allowing CW to push currency code instead of currency id @MarekBiczysko @saxix @domdinicola

PR #5995, opened 2026-05-12
Currency hotfixes

Jeeziorny:
Summary

  Hotfixes for currency handling regressions introduced after the Currency CharField→FK migration merged to 

  Fixes

1. Kobo RDI Import [AB#315387](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/315387):
    1. added "SDG" currency to tests/unit/apps/registration_data/test_file/kobo_submissions.json, 
    2. extended unit tests suite to test "SDG" currency flow and flow with unknown currency

2. Universal Update — Import & Export
    1. Symptom: Import wrote raw strings onto the FK; export wrote Currency.__str__ ("USD - US Dollar") instead of "USD", breaking the round-trip
    2. Fix: New validate_currency + handle_currency_field handlers replace the generic validate_choices/handle_simple_field; GENERATOR_TYPE_HANDLER now emits the bare .code for Currency

3. Generic Import (CW → HOPE)
    1. Symptom: "currency" was in the EXCLUDE list — households were saved with currency = NULL regardless of payload
    2. Fix: Removed from EXCLUDE; added a per-import code→id cache and a resolution step that errors out on unknown codes instead of silently dropping

4. Grievance — Household Data Update
    1. Symptom: save()/update() raised TypeError: Object of type Currency is not JSON serializable; close() raised ValueError: Cannot assign "'USD'" on .update(currency=...)
    2. Fix: save()/update() coerce Currency → .code before writing to the JSONField (mirrors existing country handling); close() resolves the code back to a Currency FK before applying the update

**Steps to replicate:** 
1. Grievance → Create Ticket → Data Change → Household Data Update → "+ Add new field" → "Which currency will be used for financial questions?" → set EUR → Submit. 
2. Assign the ticket to me, go through other stages and close the ticket.

### DIFF
--- a/src/hope/apps/generic_import/generic_upload_service/importer.py
+++ b/src/hope/apps/generic_import/generic_upload_service/importer.py
@@ -199,7 +199,7 @@ class Importer:
             "representatives",
         ]
 
-        if (currency_code := household_data.get("currency")) and isinstance(currency_code, str):
+        if currency_code := household_data.get("currency"):
             if currency_id := self._currencies.get(currency_code):
                 household_data = {**household_data, "currency": currency_id}
             else:
--- a/src/hope/apps/generic_import/generic_upload_service/importer.py
+++ b/src/hope/apps/generic_import/generic_upload_service/importer.py
@@ -17,6 +17,7 @@ from hope.models import (
     IndividualIdentity,
     RegistrationDataImport,
 )
+from hope.models.currency import Currency
 
 if TYPE_CHECKING:
     from django.core.files.uploadedfile import UploadedFile
@@ -100,6 +101,9 @@ class Importer:
 
         # Cache Country lookups (iso_code3 -> id mapping)
         self._countries = {c.iso_code3: c.id for c in Country.objects.all()}
+
+        # Cache Currency lookups (code -> id mapping)
+        self._currencies = {c.code: c.id for c in Currency.objects.all()}
 
         # Dictionary to store household instances by their parser ID (for FK linking)
         self._household_instances = {}
@@ -192,9 +196,21 @@ class Importer:
             "org_enumerator",
             "collect_type",
             "registration_method",
-            "currency",
             "representatives",
         ]
+
+        if (currency_code := household_data.get("currency")) and isinstance(currency_code, str):
+            if currency_id := self._currencies.get(currency_code):
+                household_data = {**household_data, "currency": currency_id}
+            else:
+                self.errors.append(
+                    {
+                        "type": RecordType.HOUSEHOLD,
+                        "data": household_data,
+                        "errors": {"currency": [f"Unknown currency code: {currency_code}"]},
+                    }
+                )
+                return
 
         household_instance, errors = self._build_unsaved_instance(
             model_cls=Household, data=household_data, exclude=exclude
--- a/src/hope/apps/grievance/services/data_change/household_data_update_service.py
+++ b/src/hope/apps/grievance/services/data_change/household_data_update_service.py
@@ -23,6 +23,7 @@ from hope.apps.household.services.household_recalculate_data import (
     recalculate_data,
 )
 from hope.models import Area, Household, Individual, country as geo_models, log_create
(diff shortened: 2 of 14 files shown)


pavlo-mk (approved):

src/hope/apps/grievance/services/data_change/household_data_update_service.py
```diff
@@ -173,6 +178,11 @@ def close(self, user: AbstractUser) -> None:
             household_data["country"]["value"] = geo_models.Country.objects.filter(  # type: ignore[index]
                 iso_code3=country.get("value")
             ).first()
+        currency = household_data.get("currency", {})
+        if currency.get("value") is not None:
+            household_data["currency"]["value"] = Currency.objects.filter(  # type: ignore[index]
+                code=currency.get("value")
+            ).first()
```
nerjan, line 185: idk if thats possible scenario but if code does not exists in our database we will silently save it as None, on the other hand if thats not possible scenario we might just use get.

src/hope/apps/generic_import/generic_upload_service/importer.py
```diff
@@ -192,10 +196,22 @@ def _import_household(self, household_data: dict[str, Any]) -> None:
             "org_enumerator",
             "collect_type",
             "registration_method",
-            "currency",
             "representatives",
         ]
 
+        if (currency_code := household_data.get("currency")) and isinstance(currency_code, str):
```
nerjan: I think that `isinstance(currency_code, str)` part only introduces possibility for it to silently fail. If currency_code is not string then it will be added to errors which seems like a good idea.

src/hope/apps/payment/services/payment_household_snapshot_service.py
```diff
@@ -92,6 +93,8 @@ def get_household_snapshot(household: Household, payment: Payment | None = None)
     for key in keys:
         value = all_household_data_dict[key]
         household_data[key] = handle_type_mapping(value)
+    currency = Currency.objects.filter(pk=household_data.get("currency_id")).first()
```
nerjan: As it is run for every Payment in PaymentPlan which might be a big number it could be good idea to cache Currencies in a dict as done in the [importer](https://github.com/unicef/hope/pull/5995/changes#diff-080194c525c19605230ad8806f955cc88bef580ba0704f88632e72cd70747f42R106:~:text=self%2E%5Fcurrencies%20%3D%20%7Bc%2Ecode%3A%20c%2Eid%20for%20c%20in%20Currency%2Eobjects%2Eall%28%29%7D)  to prevent 1 additional query per payment

nerjan (changes requested):

src/hope/apps/grievance/services/data_change/household_data_update_service.py
Jeeziorny: > if thats possible scenario but if code does not exists in our database we will silently save it as None,
- Household.currency is null=True, blank=True -> we can save household with currency set to None,
> on the other hand if thats not possible scenario we might just use get
- get will raise does not exist, I think filter().first() is more elegant, because it returns None if no objects are found.

@nerjan I would keep it as it is. It's also consistent with country_origin and country few lines above. What do you think?

src/hope/apps/generic_import/generic_upload_service/importer.py
Jeeziorny: Good idea, thanks Jan!

src/hope/apps/payment/services/payment_household_snapshot_service.py
Jeeziorny: Yesterday I had discussion with Mark and we concluded that changes in src/hope/apps/payment/services/payment_household_snapshot_service.py can be reverted from this PR. While your comment is valid and good for the previous state, now seems that it's not applying well. I'm resolving this.

nerjan (approved):

PR #6005, opened 2026-05-14
Remove multiple cash retry for dashboard

srugano:
[AB#315953](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/315953)
Removed the frontend retry loop to prevent redundant API calls while data is loading. Implemented a Redis cache lock in the  backend to ensure only one dashboard generation task runs per business area concurrently.

### DIFF
--- a/src/hope/apps/dashboard/celery_tasks.py
+++ b/src/hope/apps/dashboard/celery_tasks.py
@@ -1,6 +1,7 @@
 from datetime import date
 import logging
 
+from celery import Task
 from django.conf import settings
 from django.core.cache import cache
 from django.db import OperationalError, ProgrammingError
@@ -94,7 +95,7 @@ def update_recent_dashboard_figures() -> None:
 @app.task(bind=True)
 @log_start_and_end
 @sentry_tags
-def generate_dash_report_task(self, business_area_slug: str) -> None:
+def generate_dash_report_task(self: Task, business_area_slug: str) -> None:
     """Celery task to refresh dashboard data for a specific business area (full refresh) or the global dashboard."""
     is_retrying = False
     try:
--- a/src/hope/apps/dashboard/celery_tasks.py
+++ b/src/hope/apps/dashboard/celery_tasks.py
@@ -31,16 +31,12 @@ def update_dashboard_figures() -> None:
 
     for business_area in business_areas_with_households:
         set_sentry_business_area_tag(business_area.slug)
-        try:
-            DashboardDataCache.refresh_data(business_area.slug)
-        finally:
-            cache.delete(f"dash_report_task_running_{business_area.slug}")
+        DashboardDataCache.refresh_data(business_area.slug)
+        cache.delete(f"dash_report_task_running_{business_area.slug}")
 
     set_sentry_business_area_tag("global")
-    try:
-        DashboardGlobalDataCache.refresh_data()
-    finally:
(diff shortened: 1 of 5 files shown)


src/hope/apps/dashboard/celery_tasks.py
```diff
@@ -45,37 +59,47 @@ def update_recent_dashboard_figures(self: Any) -> None:
...
+                business_area = BusinessArea.objects.using(settings.DASHBOARD_DB).get(slug=business_area_slug)
+            except BusinessArea.DoesNotExist:
+                logger.error(
+                    f"Dashboard report generation failed: Business area with slug '{business_area_slug}' not found."
+                )
+                return
+            set_sentry_business_area_tag(business_area.slug)
+            DashboardDataCache.refresh_data(business_area.slug)
+    finally:
+        cache.delete(f"dash_report_task_running_{business_area_slug}")
```
johniak: Lock release in finally clashes with autoretry_for. When refresh_data raises OperationalError, the finally deletes dash_report_task_running_<slug> before Celery schedules the retry. During the 60s backoff the lock is gone and the data is still uncached, so any GET/POST does a successful cache.add and enqueues another task. So under DB pressure — exactly when the retries kick in — duplicate tasks pile up, which is what this PR is meant to prevent.

johniak (changes requested):

src/hope/apps/dashboard/celery_tasks.py
srugano: Ok. Let me fix it.

srugano: I removed the `finaly` block and made it simple.

src/hope/apps/dashboard/celery_tasks.py
```diff
@@ -79,3 +94,5 @@ def generate_dash_report_task(self: Any, business_area_slug: str) -> None:
             return
```
johniak: maybe here we need to remove lock as well

src/hope/apps/dashboard/celery_tasks.py
```diff
@@ -45,26 +55,31 @@ def update_recent_dashboard_figures(self: Any) -> None:
...
 
-@app.task(bind=True, default_retry_delay=60, max_retries=3)
+@app.task(
+    autoretry_for=(OperationalError, ProgrammingError, psycopg2.errors.InvalidCursorName),
+    retry_kwargs={"max_retries": 3, "countdown": 60},
+)
 @log_start_and_end
 @sentry_tags
-def generate_dash_report_task(self: Any, business_area_slug: str) -> None:
+def generate_dash_report_task(business_area_slug: str) -> None:
```
johniak: There is also another bug with retries. Lock leaks outside. If we have 3 retries and all 3 fails, the lock will be still there. 

To fix this we would probably need a manual retry process not from decorator

johniak (changes requested):

domdinicola:
@johniak can you pls rereview?

src/hope/apps/dashboard/celery_tasks.py
srugano: I removed it and used try/except

src/hope/apps/dashboard/celery_tasks.py
srugano: Indeed. I updated with bind instead of automatic retry

johniak (approved):

PR #6008, opened 2026-05-18
[312521] Payments: payment plan groups send to payment gateway

nerjan:
[AB#312521](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/312521) 
- New action: PaymentPlanGroup.send_to_payment_gateway()
  - Validates all plans in group are in ACCEPTED status
  - Calls existing per-plan PG send logic for each plan in sequence
  - Updates group-level status
- New field PaymentPlanGroup.status (PENDING / SENDING / SENT / FAILED)
- API endpoint: POST /api/payment-plan-groups/{id}/send-to-payment-gateway/


src/hope/apps/payment/api/views.py
```diff
@@ -2354,6 +2358,46 @@ def export(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+
+            send_jobs = AsyncJob.objects.filter(
+                content_type=ContentType.objects.get_for_model(PaymentPlan),
+                object_id__in=[str(plan.pk) for plan in plans],
+                job_name=send_to_payment_gateway_async_task.__name__,
+            )
+            in_progress_ids = {job.object_id for job in send_jobs if job.task_status in job.ACTIVE_STATUSES}
+            plans = [plan for plan in plans if str(plan.pk) not in in_progress_ids]
+            if not plans:
+                raise ValidationError("This selected group is already being sent to payment gateway.")
```
MarekBiczysko: As discussed offline, let's prevent double clicks on the lower level (execute_update_status_action)

src/hope/models/payment_plan_group.py
```diff
@@ -48,3 +56,29 @@ def delete(self, *args: object, **kwargs: object) -> tuple[int, dict]:
...
+            & Q(financial_service_provider__isnull=False)
+            & (
+                Q(use_payment_gateway=True)
+                | Q(
+                    financial_service_provider__communication_channel=FinancialServiceProvider.COMMUNICATION_CHANNEL_API
+                )
+            )
+            & Q(has_unsent_splits=True)
+            & ~Q(background_action_status=PaymentPlan.BackgroundActionStatus.SEND_TO_PAYMENT_GATEWAY)
+        )
```
MarekBiczysko, line 84: TBH I would simplify this and just iterate over PPs on the Python level and check for existing properties

something like

return [pp for pp in self.payments_plans.all() if pp.can_send_to_payment_gateway]


This is more readable and all logic related to "use_payment_gateway" flag is still in one place

MarekBiczysko (changes requested):

src/hope/models/payment_plan_group.py
nerjan: I don't know it thats worth - we could have a lot of paymentplans in one group and this query is reallys saving us a lot of queries. I added comment to original function to remember to update this place if the other is changing, but losing optimization only for a little bit more readibility doesn't seem right. But if you insist I can change it as you suggest ofc.

src/hope/apps/payment/api/views.py
nerjan: Done.

PR #6017, opened 2026-05-19
Add Inform API form discovery for RDI imports

kaltaher:
## Summary

This PR adds an initial Inform API integration scaffold for HOPE Country Workspace RDI imports.

The main purpose is to allow country offices using UNICEF Inform for data collection to discover available Inform forms dynamically from HOPE, rather than manually exporting data and reshaping it outside the system.

## What this PR adds

- New Inform API configuration using environment variables.
- Inform API client for listing available Inform forms.
- Inform API client method for retrieving form submissions by form ID.
- New backend endpoint to return Inform forms available to the configured Inform user.
- Frontend-generated model/service additions for consuming the Inform forms list endpoint.

## Why this is needed

Some country offices and partners may use Inform as their primary data collection platform. At the moment, HOPE Country Workspace appears to support import flows mainly through KoBo, Aurora, and XLS/RDI.

This can create additional manual steps where ICT/IM teams export data from Inform, clean it in Excel, reshape it to match RDI requirements, and then import it into HOPE.

Adding Inform as an additional API-based source would improve interoperability and reduce manual handling.

## Notes

This PR intentionally keeps the implementation generic. It does not hard-code any Inform form ID. Forms are discovered dynamically through the configured Inform forms endpoint.

The Inform API token and base URL are configured via environment variables and are not stored in source code.

## Follow-up work

Further implementation may be needed to complete the end-to-end import flow, including:

- mapping Inform submissions to HOPE RDI structures;
- handling repeat groups for household members;
- adding import polling/status handling;
- adding validation messages for Inform-specific data issues;
- connecting the form selection UI into the full RDI import workflow.

### DIFF
--- a/src/frontend/src/restgenerated/index.ts
+++ b/src/frontend/src/restgenerated/index.ts
@@ -109,6 +109,8 @@ export type { FullList as FullListCamelCase } from './models/FullList';
 export type { GenericImportResponse as GenericImportResponseCamelCase } from './models/GenericImportResponse';
 export type { GenericImportUpload as GenericImportUploadCamelCase } from './models/GenericImportUpload';
 export type { GetKoboAssetList as GetKoboAssetListCamelCase } from './models/GetKoboAssetList';
+export type { GetInformFormList as GetInformFormListCamelCase } from './models/GetInformFormList';
+export type { InformFormObject as InformFormObjectCamelCase } from './models/InformFormObject';
 export type { GrievanceChoices as GrievanceChoicesCamelCase } from './models/GrievanceChoices';
 export type { GrievanceComplaintTicketExtras as GrievanceComplaintTicketExtrasCamelCase } from './models/GrievanceComplaintTicketExtras';
 export type { GrievanceCreateNote as GrievanceCreateNoteCamelCase } from './models/GrievanceCreateNote';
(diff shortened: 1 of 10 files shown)


MarekBiczysko (commented):
Just curious, why this new API is not placed in the Country Workspace if the RDI data goes through CW anyway? If I understand correctly, at the end, this is the CW -> HOPE RDI import flow?

In case it stays in HOPE, can we place it in registration_data app, since this is strictly RDI related? Can we reuse/extend BaseAPI if possible (src/hope/apps/core/api/mixins.py)?

kaltaher:
> Just curious, why this new API is not placed in the Country Workspace if the RDI data goes through CW anyway? If I understand correctly, at the end, this is the CW -> HOPE RDI import flow?
> 
> In case it stays in HOPE, can we place it in registration_data app, since this is strictly RDI related? Can we reuse/extend BaseAPI if possible (src/hope/apps/core/api/mixins.py)?

Thanks, that makes sense.

Yes, my understanding is that the intended flow is Country Workspace → HOPE RDI import. I placed the initial scaffold under `core` because the existing KoBo form-discovery endpoint is there, but I agree Inform is RDI-specific and should sit under `registration_data`.

I can revise it in that direction and reuse/extend the existing `BaseViewSet` / mixins where appropriate.

One question: Do you suggest the Inform form-discovery endpoint to live under the existing RDI import viewset, or as a separate registration data API endpoint?

ntrncic:
@kaltaher We are strategically moving the RDI feature to https://github.com/unicef/hope-country-workspace. Could you please open PR there. Thank you.

PR #6026, opened 2026-05-20
feat(payment): add top-up payment plan creation with fixed and xlsx modes

johniak:
- support fixed-amount and custom xlsx top-up creation in CreateTopUpPaymentPlan
- add create-top-up, create-top-up-from-xlsx, and top-up-template endpoints
- enable CreateTopUpPaymentPlan via canCreateTopUp flag on PaymentPlanDetail


src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -968,6 +965,128 @@ def create_follow_up(
 
         return follow_up_pp
 
+    @staticmethod
+    def build_equal_share_amounts(source_pp: PaymentPlan, total: Decimal) -> dict[str, Decimal]:
```
MarekBiczysko: build_equal_share_topup_amounts

src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -968,6 +965,128 @@ def create_follow_up(
...
+    def create_top_up_payments(self, beneficiary_amounts: dict[str, Decimal]) -> None:
+        self.payment_plan.exchange_rate = self.payment_plan.get_exchange_rate()
+        self.payment_plan.save(update_fields=["exchange_rate", "updated_at"])
+
+        eligible = self.payment_plan.source_payment_plan.eligible_payments_for_top_up().select_related(
+            "household", "currency"
+        )
+
+        if not (split := self.payment_plan.splits.first()):
+            split = PaymentPlanSplit.objects.create(payment_plan=self.payment_plan)
```
MarekBiczysko: this should be created in create_top_up?

src/hope/models/payment_plan.py
```diff
@@ -758,6 +756,28 @@ def unsuccessful_payments_for_follow_up(self) -> "QuerySet":
...
 
+    def eligible_payments_for_top_up(self) -> "QuerySet":
+        """Select successful + pending payments eligible for top-up.
+
+        Must be called on the source (Standard) Payment Plan. Excludes withdrawn
+        households and beneficiaries that already appear in another TopUp under
+        the same source plan (slide 10 "Not Allowed #1": one top-up per
+        beneficiary per cycle).
+        """
+        top_up_households = Payment.objects.filter(
```
MarekBiczysko, line 775: eligible() ?

MarekBiczysko (changes requested):

MarekBiczysko (commented):
Why do we need custom import XLSX service for topup amount, can't we just use entitlements import?

Can't we extend/modify follow up creation service (or create generic "create child pp" service), to handle successful payments and use existing tools to upload entitlement (steficon, xlsx import)?

src/hope/apps/payment/api/views.py
```diff
@@ -843,18 +847,17 @@ def create(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
         )
 
-    @extend_schema(
-        request=PaymentPlanCreateFollowUpSerializer,
-        responses={201: PaymentPlanDetailSerializer},
-    )
-    @action(detail=True, methods=["post"], url_path="create-follow-up")
-    @transaction.atomic
-    def create_follow_up(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+    def _create_child_plan_response(self, request: Request, service_method_name: str) -> Response:
```
MarekBiczysko: I would rather pass a PaymentPlan.PlanType arg and use some inner map {type: action}, up to you

src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -968,34 +1046,60 @@ def create_follow_up(
...
-            dispersion_end_date=dispersion_end_date,
-            start_date=source_pp.start_date,
-            end_date=source_pp.end_date,
-            use_payment_gateway=source_pp.use_payment_gateway,
-            delivery_mechanism=source_pp.delivery_mechanism,
-            financial_service_provider=source_pp.financial_service_provider,
-            payment_plan_group=source_pp.payment_plan_group,
-            follow_up_instruction=follow_up_instruction,
+        follow_up_pp = self._create_child_payment_plan(
+            user, dispersion_start_date, dispersion_end_date, PaymentPlan.PlanType.FOLLOW_UP, " Follow Up"
```
MarekBiczysko, line 1063: can't we create postfix based on plan type (title()?)

src/hope/apps/payment/celery_tasks.py
```diff
@@ -880,32 +880,57 @@ def prepare_payment_plan_async_task(payment_plan: PaymentPlan) -> bool | None:
...
     return True
 
 
-def prepare_follow_up_payment_plan_async_task(payment_plan: PaymentPlan) -> bool | None:
+def prepare_child_payment_plan_async_task(payment_plan: PaymentPlan, payments_method: str) -> bool | None:
+    """Queue copying of payments for a child plan (follow-up / top-up / top-up amendment).
+
+    ``payments_method`` is the name of the ``PaymentPlanService`` method that copies
+    the payments (``create_follow_up_payments`` / ``create_top_up_payments`` /
+    ``create_top_up_amendment_payments``).
```
MarekBiczysko: same here, methods as strings

MarekBiczysko (changes requested):

src/hope/apps/payment/services/payment_plan_services.py
johniak: .label/.title() doesn't fit TOP_UP_AMENDMENT: the source plan is already a TopUp, so the suffix is intentionally just " Amendment" (not " Top Up Amendment"). The mapping isn't a pure function of plan type, so I'd keep the suffixes explicit at the call site.

PR #6029, opened 2026-05-20
handle vision currency 

domdinicola:
[AB#317031](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/317031)

We want to add few values for Currency model:
- vision_code: which is the value used for the exchange rate
- active: to flag if the currrency is available for selection
- number of digits


MarekBiczysko (approved):

src/hope/admin/geo.py
```diff
@@ -90,8 +90,8 @@ def get_list_filter(self, request: "HttpRequest") -> list:
 
 @admin.register(Country)
 class CountryAdmin(ValidityManagerMixin, SyncModelAdmin, FieldsetMixin, HOPEModelAdminBase):
-    list_display = ("name", "short_name", "iso_code2", "iso_code3", "iso_num")
-    search_fields = ("name", "short_name", "iso_code2", "iso_code3", "iso_num")
+    list_display = ("name", "short_name", "iso_code2", "iso_code3", "iso_num", "iso_code_4217")
```
Jeeziorny: From wikipedia I see that ISO 4217 is about currencies. Why do we change Country related stuff?

src/hope/apps/core/api/views.py
```diff
@@ -145,8 +145,8 @@ class ChoicesViewSet(ViewSet):
     def currencies(self, request: Request) -> Response:
         from hope.models.currency import Currency
 
-        choices = Currency.objects.values_list("code", "name").order_by("code")
-        resp = ChoiceSerializer(to_choice_object(list(choices)), many=True).data
+        currencies = Currency.objects.values("code", "name", "iso_code_4217").order_by("code")
+        resp = [{"value": c["code"], "name": c["name"], "iso_code_4217": c["iso_code_4217"]} for c in currencies]
```
Jeeziorny: Isn't what we're doing here a serialiser responsibility? In other words, shouldn't the logic you're doing in this line exists in some serialiser?

src/hope/config/fragments/vision.py,cover
```diff
@@ -0,0 +1,5 @@
+> from hope.config.env import env
```
Jeeziorny: What's the purpose of this file?

src/hope/contrib/api/serializers/vision.py,cover
```diff
@@ -0,0 +1,26 @@
+! from rest_framework import serializers
```
Jeeziorny: SAA

src/hope/models/country.py
```diff
@@ -42,6 +42,7 @@ class Country(NaturalKeyModel, MPTTModel, UpgradeModel, TimeStampedUUIDModel):
     iso_code2 = models.CharField(max_length=2, unique=True)
     iso_code3 = models.CharField(max_length=3, unique=True)
     iso_num = models.CharField(max_length=4, unique=True)
+    iso_code_4217 = models.CharField(max_length=5, blank=True, default="")
```
saxix: why country has iso_code_4217? It should have a FK to Currency

src/hope/models/currency.py
```diff
@@ -6,6 +8,7 @@ class Currency(models.Model):
     code = models.CharField(max_length=5, unique=True, db_index=True)
     name = models.CharField(max_length=255)
     is_crypto = models.BooleanField(default=False, help_text=_("Whether this is a cryptocurrency (e.g. USDC)"))
+    iso_code_4217 = models.CharField(max_length=5, blank=True, default="")
```
saxix: be consistent with Country. Add iso_code3 and iso_num refer to 4217 in the help_text. we also need to add 'active' (bool), decimal_digits (int), and missing 'help_text'

saxix (changes requested):

src/hope/models/currency.py
```diff
@@ -1,11 +1,18 @@
+from typing import Any
+
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
 
 class Currency(models.Model):
-    code = models.CharField(max_length=5, unique=True, db_index=True)
-    name = models.CharField(max_length=255)
+    code = models.CharField(max_length=5, unique=True, db_index=True, help_text=_("The currency code (e.g. USD, EUR)"))
```
Jeeziorny: Can we add that it is [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) standard in help_text?

src/hope/apps/core/api/views.py
```diff
@@ -145,8 +145,8 @@ class ChoicesViewSet(ViewSet):
     def currencies(self, request: Request) -> Response:
         from hope.models.currency import Currency
 
-        choices = Currency.objects.values_list("code", "name").order_by("code")
-        resp = ChoiceSerializer(to_choice_object(list(choices)), many=True).data
+        currencies = Currency.objects.values("code", "name", "vision_code").order_by("code")
+        resp = [{"value": c["code"], "name": c["name"], "vision_code": c["vision_code"]} for c in currencies]
```
Jeeziorny: Why don't we use serialiser for this operation?

src/hope/apps/payment/services/payment_gateway.py
```diff
@@ -104,7 +104,6 @@ class PaymentPayloadSerializer(serializers.Serializer):
     first_name = serializers.CharField(required=False, allow_blank=True)
     middle_name = serializers.CharField(required=False, allow_blank=True)
     full_name = serializers.CharField(required=False, allow_blank=True)
-    destination_currency = serializers.CharField(required=True)
```
nerjan, line 107: why is this deleted?

src/hope/models/currency.py
domdinicola: yes please

src/hope/apps/payment/services/payment_gateway.py
domdinicola: we have it at payment plan level @nerjan

nerjan (approved):

PR #6038, opened 2026-05-22
[AB#317779]-Change hardcoded choices on the frontend to API query 

mmaciekk:
---
  Replace hardcoded status strings with backend-generated enum
   types
  
  Problem

  The frontend had two categories of hardcoded magic strings:
  
  1. Status-to-color mapping functions in utils.ts used raw
  string literals ('Pending', 'ACTIVE', 'Merged', etc.)
  instead of the TypeScript enums auto-generated from backend
  DRF serializers in src/restgenerated/models/.
  2. Filter UI components used 'active' and 'all' as plain
  strings for the programme state filter, duplicated across 12
   files with no single source of truth.

  Two functions (registrationDataImportStatusToColor,
  programCycleStatusToColor) also had a latent bug: they
  switched on human-readable display strings (e.g. 'Merged',
  'Draft') while callers passed raw API enum values ('MERGED',
   'DRAFT'), so their cases never matched and all statuses
  fell through to the default color.

  Additionally, gender and status filter dropdowns in
  PeopleFilter and
  FilterIndividualsOffline/FilterIndividualsOnline were
  hardcoded instead of being driven by the choices API.

  Solution

  utils.ts — enum replacement:
  Replaced hardcoded string literals with the appropriate
  generated enum members across 10 status-to-color functions,
  importing:
  - PaymentStatusEnum, PaymentVerificationStatusEnum,
  PaymentVerificationSummaryStatusEnum
  - RegistrationDataImportStatusEnum,
  DeduplicationEngineStatusEnum
  - BuildStatusEnum, ProfileStatusEnum, ProgramStatusEnum
  
  The two broken functions were also fixed: their cases now
  use the correct raw enum values that callers actually pass.
  
  constants.ts — PROGRAM_STATE_FILTER:
  Introduced PROGRAM_STATE_FILTER = { ACTIVE: 'active', ALL: 
  'all' } as const as a single source of truth for the UI-only
   programme state filter values. These are not backend enum
  values — 'active' maps to the isActiveProgram: true API
  parameter and 'all' disables the filter. Updated all 12
  files that referenced these strings.

  Filter dropdowns — choices API:
  Replaced hardcoded gender and individual status MenuItem
  lists in PeopleFilter, FilterIndividualsOffline, and
  FilterIndividualsOnline with dynamic lists fetched from the
  choices API endpoint.

  Why TODOs were placed

  Four functions were left with hardcoded strings and a // 
  TODO: replace with generated enum once backend exposes these
   as a typed enum comment:

  - periodicDataUpdatesOnlineEditsStatusToColor — statuses
  like 'NEW', 'READY', 'APPROVED', 'FAILED_MERGE', etc.
  - periodicDataUpdateTemplateStatusToColor — statuses like
  'TO_EXPORT', 'EXPORTING', 'CANCELED', etc.
  - periodicDataUpdatesUpdatesStatusToColor — statuses like
  'SUCCESSFUL', 'NOT_SCHEDULED', etc.
  - householdStatusToColor / individualStatusToColor —
  'ACTIVE' / 'INACTIVE'
  
  In all these cases the backend OpenAPI schema exposes the
  field as plain string rather than a typed enum, so no
  generated TypeScript enum exists in restgenerated/ to
  import. Using ProfileStatusEnum for household/individual
  would be semantically incorrect (it represents user account
  state). The TODOs mark these for follow-up once the backend
  adds proper enum annotations to the serializers.


pavlo-mk (approved):

pavlo-mk (approved):
LGTM

pkujawa (approved):

PR #6044, opened 2026-05-26
Generate REST API types on the fly in CI instead of committing them

mmaciekk:
**Goal**: stop committing `src/frontend/src/restgenerated/` (~25k LOC of openapi-typescript-codegen output) and regenerate it in CI from the live Django OpenAPI schema instead.

**Changes**

Removed from git (now generated in CI):
- `src/frontend/src/restgenerated/index.ts`
- `src/frontend/src/restgenerated/models/*` — ~373 model files
- `src/frontend/src/restgenerated/services/*` — `FieldsAttributesService.ts`, `HhStatusService.ts`, `RestService.ts`
- `src/frontend/.gitignore` updated to ignore `models/`, `services/`, `index.ts` under `restgenerated/`

New workflow — `.github/workflows/generate-types.yml`:
- Boots Postgres via GitHub Actions native `services:` (no docker-compose, no Redis)
- Runs `manage.py migrate --run-syncdb`
- Starts Django (`runserver --classic --noreload 8080`) with `DEBUG=True`, `ALLOWED_HOSTS=*`, `CACHE_ENABLED=False`
- Waits up to 120 s for `/api/rest/` (bounded, with PID liveness check and log dump on failure)
- Runs `bun run generate-rest-api-types-camelcase`, uploads the result as the `restgenerated-types` artifact
- Final `bun run build` smoke-tests that the generated types actually compile

Downstream workflows wired to consume the artifact (via `actions/download-artifact@v4` into `src/frontend/src/restgenerated/`):
- `ci-pr.yml` — `frontend`, `tests`, `build` jobs now `needs: [generate-types]`
- `frontend-checks.yml`
- `tests.yml`
- `build-and-push.yml`

**Misc**:
- `src/frontend/enhance_and_camelize_openapi.js` — minor tweak (uses `BACKEND_URL` env var)
- `src/frontend/package.json` — script wiring for the codegen pipeline


.github/workflows/generate-types.yml
```diff
@@ -0,0 +1,127 @@
...
+        run: docker-compose -f development_tools/compose.yml up -d db redis
+
+      - name: Wait for Postgres to be ready
+        run: |
+          until docker exec $(docker-compose -f development_tools/compose.yml ps -q db) pg_isready -U postgres; do
+            echo "Waiting for Postgres..."
+            sleep 2
+          done
+
+      - name: Create database
```
Jeeziorny: Do we really need database to run just for creating FE types?

.github/workflows/generate-types.yml
```diff
@@ -0,0 +1,127 @@
...
+      ALLOWED_HOSTS: "*"
+      PYTHONPATH: ./src
+      ELASTICSEARCH_SYNONYMS_FILE: ${{ github.workspace }}/src/data/synonyms.txt
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
```
Jeeziorny, line 38: Is it minimal setup possible? The point of this is to read the models (code) and generate artefact from it. Question: How many dependencies we need to run this operation, to make it as lighweight as possible, but working properly?

.github/workflows/generate-types.yml
```diff
@@ -0,0 +1,127 @@
...
+            libpangocairo-1.0-0 \
+            shared-mime-info
+
+      - name: Install Docker Compose
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y docker-compose
+
+      - name: Launch services with Docker Compose
+        run: docker-compose -f development_tools/compose.yml up -d db redis
```
Jeeziorny: Hm is redis required here?

.github/workflows/generate-types.yml
mmaciekk: - Database? Still needed (for constance + currencies
  choices), but now set up with two lines of YAML instead
   of 4 steps + docker-compose install.

.github/workflows/generate-types.yml
mmaciekk: - Minimal setup? Drops from 9 service-setup steps to 0
  — GitHub Actions handles postgres automatically.

.github/workflows/generate-types.yml
mmaciekk: Redis is not needed.

pavlo-mk (approved):

domdinicola:
@kosior @MarekBiczysko @pkujawa need your review here

Jeeziorny:
Here is my thought process:

1. `ci-pr.yml`:
    - file that represents a workflow being triggered every time commit arrives on PR. 
    - You modified it, so now on every commit within PR `generate-types.yml` will be executed and uploaded somewhere in github, so next jobs can download it and use it.
    - frontend, tests and build will download them and use them.
Everything is clear here.

2. `ci-cd.yml`:
    - a file, that represents a workflow being triggered on every PUSH to develop, staging, master and ephemeral within HOPE repo. 
    - It uses `.github/workflows/frontend-checks.yml`, `.github/workflows/tests.yml` and `.github/workflows/build-and-push.yml` which now have `Download generated types` step.
    - Since you didn't alter `ci-cd.yml` by adding `generate-types` before the above steps, the types **won't be generated**
    - Every push to develop, staging, master and ephemeral will go red.

Either I'm missing something and need an explanation from you where I did a mistake or we need to alter `ci-cd.yml` too. I'm wondering how about `cd.yml`, but didn't have time to analyse it.

Jeeziorny (approved):

PR #6052, opened 2026-06-01
Sentry N+1 add select_related

pavlo-mk:


src/hope/apps/grievance/utils.py
```diff
@@ -28,17 +28,22 @@ def traverse_sibling_tickets(grievance_ticket: GrievanceTicket, selected_individ
...
+        .exclude(base_exclude)
         .prefetch_related("possible_duplicates")
         .distinct()
     )
 
-    selected_individuals_set = {str(i.id) for i in selected_individuals}
+    selected_individuals_set = {str(i) for i in selected_individual_ids}
     for ticket_details in ticket_details_queryset:
         possible_duplicates_set = {str(i.id) for i in ticket_details.possible_duplicates.all()}.union(
             {str(ticket_details.golden_records_individual.id)}
```
nerjan: as you are optimizing queries in this PR this is also worth looking for either `select_related("golden_records_individual")` or just `ticket_details.golden_records_individual_id` but confirm with less queries in tests :)

nerjan (dismissed):

src/hope/apps/grievance/utils.py
pavlo-mk: yep make sense, updating it and gonna check tests. thank you

src/hope/admin/payment_plan.py
```diff
@@ -104,6 +104,13 @@ class PaymentPlanAdmin(HOPEModelAdminBase, PaymentPlanCeleryTasksMixin):
     search_fields = ("id", "unicef_id", "name")
     date_hierarchy = "updated_at"
     inlines = [FundsCommitmentItemInline]
+    raw_id_fields = (
+        "imported_file",
+        "export_file_entitlement",
+        "export_file_per_fsp",
+        "export_pdf_file_summary",
+        "reconciliation_import_file",
+    )
```
MarekBiczysko, line 113: this still leaves us ability to easily download the file for debugging purposes?

src/hope/apps/household/api/serializers/individual.py
```diff
@@ -286,8 +286,8 @@ class Meta:
 
     def get_role(self, obj: dict) -> str | None:
         roles = getattr(obj, "prefetched_roles", None)
-        if roles:
-            role = roles[0]
+        if roles is not None:
+            role = roles[0] if roles else None
```
MarekBiczysko, line 290: tbh I don;t get this one

src/hope/apps/registration_data/tasks/rdi_merge.py
```diff
@@ -342,18 +358,23 @@ def _update_household_collections(self, households: QuerySet[Household], rdi: Re
...
-            # find other individual with the same unicef_id and group them in the same collection
-            individual_from_collection = (
-                Individual.objects.filter(
-                    unicef_id=individual.unicef_id,
-                    business_area=rdi.business_area,
-                )
-                .exclude(registration_data_import=rdi)
-                .first()
+        individuals_list = list(individuals)
+        new_unicef_ids = [ind.unicef_id for ind in individuals_list]
```
MarekBiczysko, line 364: individuals.values_list(unicef_id)

why
individuals_list = list(individuals)
?

MarekBiczysko (changes requested):

src/hope/apps/household/api/serializers/individual.py
pavlo-mk: just to make it more clear

src/hope/admin/payment_plan.py
pavlo-mk: do we need all this information? because to download file we still need to get url from shell
image
...
image

src/hope/apps/registration_data/tasks/rdi_merge.py
pavlo-mk: - `list(individuals)` evaluates the queryset once, pulling the full rows you need for the loop.
  - `[ind.unicef_id for ind in individuals_list]` then reads unicef_id off objects already in memory → zero extra DB hits.
  - `individuals.values_list("unicef_id", flat=True)` is a separate SQL query. Querysets are lazy and re-evaluate on each
  iteration, so you'd hit the DB once for the ids and again when you loop for the objects — two queries over the same rows,
  for no benefit.
  So `values_list` would be more DB work here, not less. It only wins when ids are all you need and you never touch the full
  objects.

src/hope/admin/payment_plan.py
MarekBiczysko: i thought there was a way to get the file directly from the admin

src/hope/apps/household/api/serializers/individual.py
MarekBiczysko: not sure if this is more clear :D

src/hope/apps/registration_data/tasks/rdi_merge.py
MarekBiczysko: just worried about fetching the whole list of individuals, this could kill db in some instances

src/hope/apps/household/api/serializers/individual.py
pavlo-mk: reverting xD

src/hope/admin/payment_plan.py
pavlo-mk: nope... 
I don't know if all this info can be helpful... like `payment_plan_payment_list_PP-0060-24-00000055.zip - 2024-07-10 10:58:36.096929+00:00` just a name of the file. 
If so we can keep it

src/hope/apps/household/api/serializers/individual.py
pavlo-mk: so basically I checked tests and in some point it's also less queries on serializing Individual:
Both return the same value (None);
only is not None honors the prefetch and avoids the per-row query. 
That's exactly the difference behind the caching test's counts going from 18→20 / 13→15 in `test_individual_list_caching`

src/hope/apps/registration_data/tasks/rdi_merge.py
pavlo-mk: `- list(individuals) does not issue a new query. It iterates the already-populated _result_cache.
  - It does not pull a second copy of the rows from the DB. It builds a new Python list holding references to the same model
  instances already in memory`

MarekBiczysko (approved):

nerjan (approved):

PR #6053, opened 2026-06-01
Add special page with admin-configurable content

mmaciekk:
Hidden page at /surprise, Photo, heading and subheading are configurable via Django admin under "Special Page Configuration". Falls back to a bundled local photo if no image is uploaded.


src/hope/apps/grievance/api/views.py
mmaciekk, line 1: had to make this change for mypy check to pass on the CI

src/hope/apps/core/migrations/0030_migration.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 5.2.14 on 2026-06-01 08:47
```
Jeeziorny: I think you can merge those migrations into one file

mmaciekk: done

src/hope/models/surprise_page_config.py
```diff
@@ -0,0 +1,35 @@
+from typing import Any
+
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+
+class SurprisePageConfig(models.Model):
```
domdinicola: not sure we want to add another model...

mmaciekk: Right, it makes sense. 
I changed it and now it is configurable in constance.
image

domdinicola (approved):

Jeeziorny (approved):

PR #6066, opened 2026-06-04
feat: add SENTRY_TRACES_SAMPLE_RATE

elnino-nic:
Changing SENTRY_TRACES_SAMPLE_RATE in Sentry affects how many performance traces your application sends to Sentry.

What it does

SENTRY_TRACES_SAMPLE_RATE is a value between 0 and 1:

0.0 → sends no traces (performance monitoring effectively off)
1.0 → sends 100% of traces
0.1 → sends 10% of traces, etc.


Jeeziorny:
What problem this PR solves?

elnino-nic:
> What problem this PR solves?

This will help us to control the amount of traces we send to Sentry.
And as a result the capacity we dedicate for self-hosted Sentry.

PR #6085, opened 2026-06-11
ES upgrade 9.0.1 -> 9.4.2 step 3

pavlo-mk:
**Upgrade and clean up**
### Step 3 — Official 9.0.1 → Official 9.4.2 (~30 s downtime, rolling-restart). tag bump (required, low risk)

This is a same-major same-vendor tag bump. No index format changes, no UID changes.
No breaking changes between 9.0 and 9.4 affect HOPE (TSDB and dense-vector changes are irrelevant).

1. **Bump the image tag** in the Helm chart:
      ```
      elasticsearch:9.4.2
      ```
2. **Remove the `initContainers` block** (Step 2 one-time chown is already done; volume is now owned by 1000:0).


johniak (approved):

.github/workflows/tests.yml
```diff
@@ -69,6 +69,20 @@ jobs:
             sleep 2
           done
 
+      - name: Wait for Elasticsearch to be ready
```
Jeeziorny, line 72: What's the purpose of those waits here? I'm not sure what it was absent before and now we're adding it. Do you have any idea @johniak ?

Jeeziorny: Discussed offline: new ES needs more time to wake up.

Jeeziorny (approved):

PR #6089, opened 2026-06-15
[HOTFIX] Add requeue functionality for AsyncJobs. registration_kobo_import_asy…

nerjan:
[AB#321571](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/321571)
Problem
A KOBO Registration Data Import could run its import step twice for the same RDI, producing duplicated PendingHousehold/PendingIndividual records that were then merged — doubling the beneficiary population (observed on a Syria RDI: 8412 individuals / 1896 households, exactly 2×).

Root cause
registration_kobo_import_async_task_action ran RdiKoboCreateTask.execute() with no concurrency or status guard 

Solution
Guard the KOBO action — wrap it in locked_cache + an IMPORT_SCHEDULED/IMPORT_ERROR status gate (matching the XLSX/program-population actions), so concurrent runs and stale retries no-op instead of re-importing.
Add BaseAsyncJob.requeue() (used by the KOBO task wrapper) — skips queueing if a job is already running, otherwise terminates any stuck PENDING/QUEUED job and queues a fresh one.


src/hope/apps/registration_data/celery_tasks.py
```diff
@@ -204,26 +204,45 @@ def registration_program_population_import_async_task(
...
-            program_id=program_id,
-        )
+        with locked_cache(key=f"registration_kobo_import_async_task-{registration_data_import_id}") as locked:
+            if not locked:
+                raise AlreadyRunningError(
+                    f"Task with key registration_kobo_import_async_task"
+                    f" {registration_data_import_id} is already running"
+                )
+            rdi = RegistrationDataImport.objects.get(id=registration_data_import_id)
+            set_sentry_business_area_tag(rdi.business_area.name)
```
MarekBiczysko: can be removed, the inner task is wrapped with tags

src/hope/apps/registration_data/celery_tasks.py
```diff
@@ -243,7 +262,7 @@ def registration_kobo_import_async_task(
         "business_area_id": business_area_id,
         "program_id": program_id,
     }
-    AsyncRetryJob.queue_task(
+    AsyncRetryJob.requeue(
```
MarekBiczysko: I guess this should stay as queue_task(), the admin action should use requeue

MarekBiczysko (changes requested):

src/hope/apps/registration_data/celery_tasks.py
nerjan: donedone

src/hope/apps/registration_data/celery_tasks.py
nerjan: done

src/hope/admin/registration_data.py
```diff
@@ -99,25 +99,24 @@ def rerun_rdi(self, request: HttpRequest, pk: str) -> HttpResponse | None:
...
+                "registration_data_import": obj,
+                "import_data_id": str(obj.import_data_id),
+                "business_area_id": str(obj.business_area_id),
+                "program_id": str(obj.program_id),
+            }
             if obj.data_source == RegistrationDataImport.XLS:
                 from hope.apps.registration_data.celery_tasks import registration_xlsx_import_async_task
 
-                celery_task = registration_xlsx_import_async_task
+                registration_xlsx_import_async_task(**common_kwargs)
```
MarekBiczysko, line 112: lets requeue=True here also

pavlo-mk (dismissed):

MarekBiczysko (changes requested):

src/hope/admin/registration_data.py
nerjan: ok I added here, but it is xlsx path

MarekBiczysko (approved):

Jeeziorny (approved):
I like it

PR #6098, opened 2026-06-17
version start

domdinicola:


MarekBiczysko (dismissed):

tests/extras/test_utils/fixtures/dashboard.py
```diff
@@ -14,8 +14,7 @@
 
 
 @pytest.fixture
-@pytest.mark.django_db(databases=["default", "read_only"])
-def populate_dashboard_cache() -> Callable[[BusinessAreaFactory, Optional[Dict]], Household]:
+def populate_dashboard_cache(django_db) -> Callable[[BusinessAreaFactory, Optional[Dict]], Household]:
```
nerjan: django_db is not a fixture its a mark - you can use it in pytest.mark.django_db, the fixture is just `db`.
Also I think it doesn't matter here because it seems this fixture is not used anywhere - we define couple of populate_dashboard_cache fixtures unique to test files, so this one is actually not used.

nerjan (changes requested):

tests/extras/test_utils/fixtures/dashboard.py
domdinicola: good catch!

nerjan (approved):

PR #6100, opened 2026-06-18
Refactor choices change migration

pavlo-mk:
[AB#322548](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/322548): choices change migration

`All the choices should have a callable/function instead of static choices.
This will reduce number of migrations.`


pkujawa (dismissed):

domdinicola:
@pavlo-mk can we resolve conflicts?

domdinicola (approved):

PR #6101, opened 2026-06-18
change drf settings

domdinicola:
All endopoint should be protected at least by authentication


Jeeziorny:
Why?

kosior (dismissed):

domdinicola:
@kosior what happen if we add a new admin area, or area type?
we have to find a way to fetch the data without this trick and dynamically

Jeeziorny (changes requested):
This PR has no description

kosior:
> @kosior what happen if we add a new admin area, or area type? we have to find a way to fetch the data without this trick and dynamically

@domdinicola Are `areas/` and `areatypes/` endpoints used outside of HOPE? Could you explain which data we want to fetch dynamically?

I added two commits:
1. workaround for ChoicesViewSet so we can generate frontend types in CI and locally without authentication
2.  update `areas/` and `areatypes/` endpoints to use HOPEAPIView which basically configures endpoints to use HOPEPermission and HOPEAuthentication

domdinicola:
my undestading is that we build the choices using the endpoint, what happen if the areas or area type change?
why the frontend can't just fetch and cache the endpoint?

tests/unit/apps/core/test_choices.py
```diff
@@ -158,14 +158,43 @@ def test_choices_permissions_returns_full_permission_catalog(authenticated_clien
...
+)
+def test_choices_denies_unauthenticated_access_outside_dev(anonymous_client, settings, debug, env):
+    settings.DEBUG = debug
+    settings.ENV = env
+    response = anonymous_client.get(reverse("api:choices-permissions"))
+
+    assert response.status_code == 403
+
+
+@pytest.mark.django_db
```
Jeeziorny: Please add
```
pytestmark = pytest.mark.django_db()
```
at the top of the while and remove `@pytest.mark.django_db` decorator from function definitions

tests/unit/api_contract/test_lookups.py
```diff
@@ -133,21 +133,21 @@ def test_lookups_program_statuses(superuser, api_token):
...
+def test_areatypes(superuser, api_token, area):
+    recorder = HopeRecorder(DATA_DIR, as_user=superuser, api_token=api_token)
     recorder.assertGET("/api/rest/areatypes/")
 
 
+# ---------------------------------------------------------------------------
+# Default DRF auth (AllowAny, no token needed)
+# ---------------------------------------------------------------------------
+
+
```
Jeeziorny:
```suggestion
```

src/hope/apps/core/api/views.py
```diff
@@ -142,6 +143,12 @@ class ChoicesViewSet(ViewSet):
     Response([{"value": k, "name": v} for k, v in PaymentPlan.Status.choices])
     """
 
+    def get_permissions(self) -> Any:
```
Jeeziorny: I'm not sure if it's good idea to have production code running dev checks. 
```
if settings.DEBUG and settings.ENV == "dev":
```
should be checked once at the initialisation phase. WDYT @kosior

domdinicola:
@Jeeziorny @kosior let's start like this while we work a better solution for type generation

PR #6103, opened 2026-06-19
bitcaster

MarekBiczysko:
Phase 1: Complete alternate Bitcaster flow under feature flag

  - [x] Wire remaining active direct email flows under BITCASTER_ENABLED.
  - [x] Cover hope/apps/grievance/notifications.py email notifications.
  - [x] Cover direct user.email_user(...) usages that do not go through send_email_notification(...).
  - [x] Cover direct send_email_async_task.delay(...) / Mailjet usages outside payment plan and PDU workflow notifications.
  - [x] Cover XlsxPaymentPlanExportPerFspService.send_email_with_passwords(...).
  - [x] Cover Western Union report email flow.
  - [x] Cover payment reconciliation overdue email flow.

Phase 2: Mailjet/template cleanup and configuration
  - [ ] Configure Bitcaster.
  - [ ] Move local templates to Mailjet (?)
  - [ ] refactor sanction list email to store the result file in storage and provide a download link
  - [ ] Delete confirmed unused email templates
  
  https://github.com/bitcaster-io/bitcaster-sdk/pull/11


Jeeziorny:
Do we need to update .env on our machines?

johniak (dismissed):

src/hope/apps/core/notifications/bitcaster_client.py
```diff
@@ -0,0 +1,74 @@
...
+        sdk_client.trigger(
+            project=self.project_slug,
+            application=self.application_slug,
+            event=event_name,
+            context=payload,
+            options=options or {},
+            cid=cid,
+        )
+        logger.info("Successfully triggered Bitcaster event: %s", event_name)
+        return True
```
saxix, line 65: let open a PR to the original code, so we can hopefully remove this monkepatch asap. Add a note here mentioning the PR

src/hope/apps/sanction_list/tasks/check_against_sanction_list.py
```diff
@@ -148,11 +167,30 @@ def execute(self, uploaded_file_id: "UUID", original_file_name: str) -> None:
...
         email.attach_file(
             attachment=base64_encoded_content,
-            filename=f"{subject}.xlsx",
-            mimetype="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+            filename=attachment.filename,
+            mimetype=attachment.content_type,
         )
         email.send_email()
 
+        if bitcaster_enabled():
```
saxix: bitcaster shound not receive the message, just context to the event endpoint

saxix (changes requested):

src/hope/apps/core/notifications/bitcaster_client.py
MarekBiczysko: https://github.com/bitcaster-io/bitcaster-sdk/pull/11

src/hope/apps/core/notifications/payloads.py
```diff
@@ -0,0 +1,17 @@
...
+class EmailAttachmentPayload:
+    filename: str
+    content_type: str
+    base64_content: str
+
+
+@dataclass(frozen=True, kw_only=True)
+class EmailPayload:
+    recipients: list[str]
+    subject: str
```
johniak: Worth settling whether `subject` should be in the event payload at all. The SDK treats `context` as server-side template data and renders the message itself, so the subject line is arguably something Bitcaster's event/template should own rather than something HOPE passes in. Recipients can stay for now; the open question is just subject vs. context-only.

src/hope/apps/sanction_list/tasks/check_against_sanction_list.py
```diff
@@ -148,11 +158,29 @@ def execute(self, uploaded_file_id: "UUID", original_file_name: str) -> None:
...
+                EmailPayload(
+                    recipients=[uploaded_file.associated_email],
+                    subject=subject,
+                    cc=[settings.SANCTION_LIST_CC_MAIL],
+                    context={
+                        "results_count": len(results_dict),
+                        "file_name": original_file_name,
+                        "today_date": timezone.now().isoformat(),
+                        "title": "Sanction List Check",
+                        "attachments": [asdict(attachment)],
```
johniak: The whole XLSX is base64-encoded and put into the event context as `attachments`. A full file inside a JSON event body risks hitting request-size limits and timeouts, and Bitcaster can't template a base64 blob anyway. Send a reference or metadata (name, row count) and let the file live where it already does. This is saxix's L166 point about sending context, not the message.

src/hope/apps/core/utils.py
```diff
@@ -610,25 +612,53 @@ def chunks(it: Iterable, size: int) -> Iterator[list]:
...
+    text_body = render_to_string(service.text_template, context=context)
     user.email_user(
-        subject=context["title"],
-        html_body=render_to_string(service.html_template, context=context),
-        text_body=render_to_string(service.text_template, context=context),
+        subject=subject,
+        html_body=html_body,
+        text_body=text_body,
+    )
+    publish_email_notification(
```
johniak: The legacy `user.email_user(...)` above runs unconditionally and Bitcaster is published right after, so with `BITCASTER_ENABLED` on and a Bitcaster email channel configured the recipient gets two emails. If this is a migration the legacy send should move under `if not bitcaster_enabled():`; if the dual send is intentional for Phase 1, please state that explicitly since as written it's a latent double-delivery.

src/hope/apps/core/notifications/bitcaster_client.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def _configure_request_timeout(self, sdk_client: Any) -> None:
```
johniak, line 67: Add a short comment marking this as temporary until that PR lands and ships a native `timeout=`, so it gets removed instead of lingering.

src/hope/apps/core/utils.py
johniak: Flag-guarding is inconsistent across the wired flows: payment, PDU and sanction-list wrap the publish in `if bitcaster_enabled():`, while this helper and `admin/api_token.py` rely on the guard inside `publish_email_notification`. Pick one convention so an unguarded call site doesn't read like a bug.

src/hope/apps/grievance/notifications.py
```diff
@@ -110,71 +159,76 @@ def _prepare_for_approval_recipients(self) -> "QuerySet[User]":
...
         context["created_by"] = f"{created_by.first_name} {created_by.last_name}"
-        context["ticket_note"] = self.extra_data.get("ticket_note")
-        text_body = render_to_string("note_added_notification_email.txt", context=context)
-        html_body = render_to_string("note_added_notification_email.html", context=context)
-        return (
-            text_body,
-            html_body,
-            f"New note in Grievance & Feedback ticket has been left {self.grievance_ticket.unicef_id}",
-        )
+        context["ticket_note_description"] = ticket_note.description
```
johniak: `ticket_note.description` will raise `AttributeError` if `ticket_note` is missing, whereas the previous code used `self.extra_data.get("ticket_note")` and the template tolerated `None`. Unless NOTES_ADDED is guaranteed to carry the note, guard it (e.g. `getattr(ticket_note, "description", "")`).

johniak (changes requested):

src/hope/apps/core/notifications/payloads.py
MarekBiczysko: I'll remove subject

src/hope/apps/sanction_list/tasks/check_against_sanction_list.py
MarekBiczysko: This sanction list attachment currently exists only in memory. I didn't want to refactor existing email logic; that's why I decided to just pass the attachments as base64. I can drop attachment support for now and refactor the sanction list email later to store the file and provide a link to storage

src/hope/apps/core/utils.py
MarekBiczysko: it's intentionally additive

johniak (dismissed):

src/hope/apps/core/notifications/handlers.py
```diff
@@ -0,0 +1,340 @@
...
+
+    try:
+        serialized_payload = _json_safe_context(asdict(payload))
+        correlation_id = correlation_id or _build_email_correlation_id(event_name, payload)
+        send_bitcaster_event_task.delay(event_name, serialized_payload, correlation_id)
+    except Exception:
+        logger.exception("Failed to queue Bitcaster event '%s'.", event_name)
+
+
+def _handle_event(
```
domdinicola: this function just call another function?

src/hope/apps/core/notifications/handlers.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+def _handle_email_event(
+    event_name: str,
+    payload: EmailPayload,
+    correlation_id: str | None = None,
+    *,
+    business_area: BusinessArea | None = None,
+    application_enabled: bool = True,
+) -> None:
```
domdinicola, line 98: let's add comment here

PR #6119, opened 2026-06-23
auth to fix

domdinicola:
Remove the `AllowAny` override from `ChoicesViewSet` so `/api/rest/choices/*` now requires authentication (project default `IsAuthenticated`).

It was only there to let FE type generation fetch the endpoints anonymously; 
that's no longer needed since type generation loads enum choices from a file
instead of the live API.

### DIFF
--- a/tests/unit/apps/core/test_choices.py
+++ b/tests/unit/apps/core/test_choices.py
@@ -158,12 +158,10 @@ def test_choices_permissions_returns_full_permission_catalog(authenticated_clien
 
 
 @pytest.mark.django_db
-def test_choices_permissions_allows_unauthenticated_access(anonymous_client):
-    # FE type generation (enhance_and_camelize_openapi.js) fetches this endpoint without credentials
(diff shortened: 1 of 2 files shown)


tests/unit/apps/core/test_choices.py
```diff
@@ -158,12 +158,10 @@ def test_choices_permissions_returns_full_permission_catalog(authenticated_clien
 
 
 @pytest.mark.django_db
-def test_choices_permissions_allows_unauthenticated_access(anonymous_client):
-    # FE type generation (enhance_and_camelize_openapi.js) fetches this endpoint without credentials
+def test_choices_permissions_denies_unauthenticated_access(anonymous_client):
     response = anonymous_client.get(reverse("api:choices-permissions"))
```
Jeeziorny, line 162: I'm wondering if we

Jeeziorny (approved):

johniak (approved):

pkujawa (approved):

pavlo-mk (approved):

PR #6121, opened 2026-06-23
protect callback

domdinicola:
https://github.com/unicef/hope-dedup-engine/blob/c7e7bb3d241d915d2f919713b4ec4e6e89c39519/src/hope_dedup_engine/apps/api/utils/notification.py#L31


src/hope/models/grant.py
```diff
@@ -15,6 +15,8 @@ def _generate_next_value_(self: str, start: int, count: int, last_values: list[A
     API_BENEFICIARY_TICKET_CREATE = auto()
     API_GENERIC_IMPORT = auto()
 
+    API_DEDUP_FETCH_FINDINGS = auto()
```
pkujawa, line 18: migration is missing

src/hope/apps/registration_data/api/views.py
```diff
@@ -88,7 +100,18 @@ class RegistrationDataImportViewSet(
...
     }
+    token_permissions_by_action = {
+        "webhook_deduplication": Grant.API_DEDUP_FETCH_FINDINGS,
+    }
     filter_backends = (OrderingFilter, DjangoFilterBackend)
+
+    def get_permissions(self) -> list[BasePermission]:
+        if self.is_external_request():
+            self.permission_classes = [HOPEPermission]
+            self.permission = self.token_permissions_by_action.get(self.action, self.token_permission)
```
pkujawa: `token_permissions_by_action` covers only `webhook_deduplication`, so every other action on this viewset will be allowed if you have token `API_READ_ONLY` (`self.token_permission`).
Having `API_READ_ONLY` will cause user to be able to do merge, erase, refuse etc which is not correct.
why not `self.permission = self.token_permissions_by_action.get(self.action)`

pkujawa (changes requested):

pkujawa:
Also I found a gap with check for `valid_for` in the `PermissionsMixin`.(It is covered by SelectedBusinessAreaMixin so the HOPEAPIView endpoints checked it.
Fix would be 
```
class HOPEPermission(IsAuthenticated):
    def has_permission(self, request: Request, view: Any) -> bool:
        if not bool(request.auth):
            return False
        if view.permission == "any":
            return True
        if not view.permission or view.permission.name not in request.auth.grants:
            return False
        # token must be valid for the business area in the URL (skipped when there is none)
        ba_slug = view.kwargs.get("business_area_slug")
        if ba_slug and not request.auth.valid_for.filter(slug=ba_slug).exists():
            return False
        return True
```
to include the check here.
Can you apply it and add a test case like `test_webhook_deduplication_with_api_token` but with check for valid_for as well?
Or should I make the fix in a separate PR? (let's merge this one first)

src/hope/models/grant.py
domdinicola: I think after @pkujawa's migration should be required anymore

domdinicola:
@pkujawa can you take it over?

johniak (approved):

vitali-yanushchyk-valor (approved):

PR #6128, opened 2026-06-25
Fix/payment module activity log

johniak:
## Problem

Payment and PaymentPlan changes applied **outside a logging view** — async tasks, payment-gateway sync, XLSX imports, the Steficon engine rule, and the exclude / lock / unlock flows — mutated records via plain `.update()` and so left **no activity-log entry**. There was no audit trail of who changed what on those paths.

## Solution

Route those mutations through activity-log helpers and record the originating user.

- **`PaymentQuerySet.update_and_log` / `bulk_log_payment_changes`** replace the silent `.update()` calls. `update_and_log` lives on the Payment queryset so the bulk update and the log write stay coupled in one call — `.update()` bypasses `save()`/signals, so bundling the log write is what guarantees rows can't be updated without an audit entry. They build the `LogEntry` directly (mirroring `log_create`), **bulk-insert** entries plus the program M2M, and **skip no-op diffs** so unchanged rows produce no log.
- The **originating user is threaded through** the relevant call sites (XLSX import tasks, gateway sync, Steficon rebuild, admin re-runs) via `user_id`; system-initiated runs pass `None`.

## Performance

- Logging cost is constant in queries per 1000-row bulk-insert batch, **not per row** (bulk insert + one M2M insert), guarded by `django_assert_num_queries` tests.
- Gateway sync **flushes logs per instruction** to bound memory on large plans (no-op for the common single-split case).
- No-op rows are skipped before any insert.

### Benchmarks

Measured against real dev-DB data (seeded PaymentPlan), gateway HTTP mocked. Time is clean wall-clock; memory is `tracemalloc` peak allocation captured in a separate run (tracemalloc inflates wall-time ~2.7x, so the two are not measured together).

**`update_and_log` — the refactored path (this PR):**

| Scenario | 5k | 10k | Scaling | SQL queries |
|---|---|---|---|---|
| all rows change | 0.54s | 0.93s | 1.7x — linear | 13 -> 23 |
| no-op (already set) | 0.19s | 0.36s | 1.9x — linear | 3 -> 3 |

| Peak alloc (tracemalloc) | 5k | 10k | per 1k payments |
|---|---|---|---|
| `update_and_log` | 11.5 MB | 22.6 MB | ~2.3 MB |

SQL is `3 fixed (values snapshot + update + user lookup) + 2 per 1000-row batch` — a naive per-row `save()` + log would be ~3x row-count queries. Both time and memory scale linearly.

**Gateway flow (consumes `bulk_log_payment_changes`):**

| Step | 5k | 10k | Scaling |
|---|---|---|---|
| `add_records` (-> SENT_TO_PG) | 7.6s | 15.2s | 2.0x — linear |
| `sync_payment_plan` (-> DISTRIBUTION_SUCCESS) | 31.3s | 67.6s | 2.2x — linear |

The gateway steps scale linearly; their cost is dominated by **pre-existing** per-payment work (`household_snapshot` serialization / per-row `save()`), not the logging this PR adds — the bulk log write is a negligible constant on top.

## Tests

Activity-log helper tests covering the insert/no-op/FK-valued/create paths, query-count assertions, and the per-call-site user threading.

- 100% patch coverage
- `tox -e mypy`, `tox -e lint` clean


src/hope/apps/payment/services/payment_gateway.py
```diff
@@ -472,17 +476,21 @@ def change_payment_instruction_status(
```
MarekBiczysko, line 480: Something to consider, keep user_id as service instance attribute (pass to PaymentGatewayService.__init__()) as use in all activity log occurrences. Now some path are not covered with user instance (add_records_to_payment_instructions, sync_missing_records etc).

src/hope/apps/payment/utils.py
```diff
@@ -6,19 +6,248 @@
...
+            new_compare[field] = value.pk
+            new_repr[field] = str(value)
+        else:
+            column = field
+            new_compare[field] = value
+            new_repr[field] = _value_repr(value)
+        column_for_field[field] = column
+
+    snapshot_columns = ["id", "unicef_id", "business_area_id", "program_id", *column_for_field.values()]
+    rows = list(queryset.values(*snapshot_columns))
```
MarekBiczysko: distinct?

src/hope/admin/payment_plan.py
```diff
@@ -118,6 +120,21 @@ def get_form(self, request: HttpRequest, obj: Any = None, change: bool = False,
         request._payment_plan_obj = obj
         return super().get_form(request, obj, change, **kwargs)
 
+    def save_model(self, request: HttpRequest, obj: PaymentPlan, form: Any, change: bool) -> None:
+        old_payment_plan = copy_model_object(PaymentPlan.objects.get(pk=obj.pk)) if change and obj.pk else None
+        super().save_model(request, obj, form, change)
+        # skip a no-op log when an edit touched only fields outside ACTIVITY_LOG_MAPPING
+        if old_payment_plan is not None and not create_diff(old_payment_plan, obj, PaymentPlan.ACTIVITY_LOG_MAPPING):
+            return
```
MarekBiczysko, line 128: maybe this can be executed inside log_create()

src/hope/apps/payment/services/payment_gateway.py
```diff
@@ -700,49 +713,62 @@ def sync_records(self) -> None:
...
         exchange_rate = payment_plan.exchange_rate
+        old_payment_plan = cast("PaymentPlan", copy_model_object(payment_plan))
 
         if not payment_plan.is_payment_gateway:
             return  # pragma: no cover
 
         payment_instructions = payment_plan.splits.filter(sent_to_payment_gateway=True)
+        # Resolve the user once and flush logs per instruction: this keeps only one split's
+        # payments (with their household_snapshot) referenced at a time, matching the pre-logging
+        # per-split GC profile instead of retaining the whole plan's payments until the end.
```
MarekBiczysko: i guess this comment can br removed, code is selfexplanatory

pavlo-mk (dismissed):

src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -268,8 +268,10 @@ def lock(self) -> PaymentPlan:
         if not self.payment_plan.can_be_locked:
             raise ValidationError("At least one valid Payment should exist in order to Lock the Payment Plan")
 
-        self.payment_plan.eligible_payments_with_conflicts.filter(payment_plan_hard_conflicted=True).update(
-            conflicted=True
+        update_payments_and_log(
```
MarekBiczysko: IMHO we shouldn't mix update and log inside this function, this logic should be separated

src/hope/apps/payment/xlsx/xlsx_payment_plan_import_service.py
```diff
@@ -61,15 +62,22 @@ def validate(self) -> None:
...
 
         for row in self.ws_payments.iter_rows(min_row=2):
             if not any(cell.value for cell in row):
                 continue
             if payment := self._import_row(row):
                 payments_to_save.append(payment)
+                old_payment = old_by_id.get(payment.id)
+                if old_payment is not None:
+                    new_for_log = copy_model_object(old_payment)
+                    new_for_log.entitlement_quantity = payment.entitlement_quantity
```
MarekBiczysko, line 81: I don't get this one

src/hope/apps/payment/celery_tasks.py
```diff
@@ -698,9 +725,15 @@ def payment_plan_apply_engine_rule_async_task_action(job: AsyncRetryJob) -> None
         pp_exchange_rate = payment_plan.exchange_rate
         pp_currency_exchange_date = payment_plan.currency_exchange_date
 
+        steficon_user_id = job.config.get("user_id")
+        # Resolve once: bulk_log_payment_changes runs per batch in the loop below, so passing the
+        # pre-resolved user avoids re-fetching the same user on every flush.
```
MarekBiczysko: remove comment

src/hope/apps/payment/utils.py
```diff
@@ -6,19 +6,248 @@
...
+    ]
+    if links:
+        through.objects.bulk_create(links, batch_size=1000)
+
+
+def bulk_log_payment_changes(
+    old_new_pairs: "list[tuple[Payment | None, Payment]]",
+    user_id: str | None = None,
+    *,
+    user: Any = _UNSET,
```
MarekBiczysko: why not just user_id ?

src/hope/apps/payment/utils.py
```diff
@@ -6,19 +6,248 @@
...
+        program_ids.append(new.program_id)
+    _persist_payment_logs(logs, program_ids)
+
+
+def log_payment_change(old: "Payment | None", new: Payment, user_id: str | None) -> None:
+    """Activity-log a single payment change (manual edits: mark-as-failed, revert, ...)."""
+    bulk_log_payment_changes([(old, new)], user_id)
+
+
+def update_payments_and_log(
```
MarekBiczysko: not a huge fan of doing djanog model isntances updates here

MarekBiczysko (changes requested):

pkujawa (dismissed):

src/hope/apps/payment/services/payment_gateway.py
johniak: Good call. Moved `user_id` into `PaymentGatewayService.__init__` and resolve the acting user once there — every log write now uses `self.user`. This also closes the gap you flagged: `send_to_payment_gateway` and the admin syncs now pass the user, so `_handle_pg_errors`/`_handle_pg_success` attribute correctly instead of `None`. The periodic `sync_records` stays user-less on purpose (system run), so it logs as `None`.

src/hope/apps/payment/utils.py
johniak: No — `payment_plan_hard_conflicted` is annotated via `Exists()` subqueries (see `with_payment_plan_conflicts`), not a join, so the queryset returns one row per payment. `.values()` can't duplicate here, so `distinct` isn't needed.

src/hope/admin/payment_plan.py
johniak: I'd keep it out of `log_create` — it always writes an entry today and several callers rely on that. Skipping empty diffs there would change behavior for everyone, so I kept the guard local to `save_model`. Happy to make it opt-in in `log_create` as a separate change if you think it's worth it.

src/hope/apps/payment/services/payment_gateway.py
johniak: Removed.

src/hope/apps/payment/services/payment_plan_services.py
johniak: The coupling is intentional: `QuerySet.update()` bypasses signals/`save()`, so there's no hook to auto-log — bundling update + log in one call is what guarantees you can't silently update without an entry (the bug this PR fixes). I agree it reads as two responsibilities though; if you'd prefer, I can move it onto a Payment queryset/manager method (`Payment.objects…update_and_log(...)`) so it's not sitting in `utils`, as a follow-up. LMK.

src/hope/apps/payment/xlsx/xlsx_payment_plan_import_service.py
johniak: `_import_row` returns a fresh `Payment(id, entitlement_quantity, entitlement_date, entitlement_quantity_usd)` — every other field is at its model default, not the real DB value. So diffing `(old, payment)` directly would log bogus changes for `status`/`delivered_quantity`/etc. (real value vs default). Copying `old` and overlaying only `entitlement_quantity` keeps the diff to the one field the import actually changes. Added a comment to explain.

src/hope/apps/payment/celery_tasks.py
johniak: Removed.

src/hope/apps/payment/utils.py
johniak: The extra `user` kwarg existed to avoid re-fetching the same user on every flush inside loops (gateway per-instruction, steficon per-batch). Simplified it now to a single `user` param — callers resolve the `user_id` once and pass the object, so no per-call lookup and no dual param.

src/hope/apps/payment/utils.py
johniak: Same reasoning as the `lock()` thread — coupled on purpose because `.update()` skips signals, so there's no auto-log hook. If the concern is a DB-mutating helper living in `utils`, I'm happy to move it onto a Payment queryset/manager method as a follow-up.

src/hope/apps/payment/services/payment_plan_services.py
MarekBiczysko: Payment.objects…update_and_log(...) sounds good

pkujawa (dismissed):

MarekBiczysko (approved):

pkujawa (approved):

PR #6129, opened 2026-06-26
Align the currency model with Vision codes

domdinicola:
## For reviewers - what to look at

Base rule: after a redenomination two rows share `code = SYP` (deprecated `SYP` / active `SYP01`), so **`code` alone no longer identifies a row, and amounts in the two rows are not comparable (100:1)**.

**Decisions to challenge**

- **No default resolver.** Every place that turns a code into a `Currency` picks one of [`currency_resolution`](https://github.com/unicef/hope/blob/53b0ab7c18/src/hope/apps/core/currency_resolution.py) explicitly:
  - **Value sent back unchanged (Payment Plan API)** -> an unchanged code keeps the row already on the record, so GET -> PUT does not move a plan (and its payments) onto the new pound. `CurrencySlugRelatedField` requires [`on_unchanged_code`](https://github.com/unicef/hope/blob/53b0ab7c18/src/hope/api/utils.py#L14); a misconfigured field fails at bind time.
  - **Imports (RDI XLSX/Kobo/people, RDI API, collision merge)** -> always the active row. An import is authoritative for the whole household.
  - **Household edits (grievance, universal update)** -> always the active row. `Household.currency` feeds neither payments nor aggregations, so no amount changes meaning.
- **`SYP01` as a transitional input alias.** An unknown `code` falls back to an active row's `vision_code`, logged on every use ([`get_active_by_code_or_none`](https://github.com/unicef/hope/blob/53b0ab7c18/src/hope/models/currency.py#L23)). Input only; a withdrawn currency is never resolved through another row's alias. End date TBD.
- **FSPs / Payment Gateway get only the ISO code** -> a plan on the inactive row is refused before any API call or export ([`inactive_currency_reason`](https://github.com/unicef/hope/blob/53b0ab7c18/src/hope/apps/payment/utils.py#L313)).
- **Targeting by currency matches all denominations of a code** ([`get_currency_query`](https://github.com/unicef/hope/blob/53b0ab7c18/src/hope/apps/core/attributes_qet_queries.py#L291)).

**API / UI changes**

- `vision_code` added to the lookup endpoint (also searchable); read-only `currency_vision_code` on households, payments, payment plans, verification plans and follow-up instructions.
- Aggregations (delivered quantity, dashboard) group by the currency row, not `code`.
- FE shows the vision code next to amounts when it differs from the ISO code.
- Activity log and `Payment.signature_fields` can tell the two rows apart.

**Out of scope / still to do**

- **Activating SYP is not part of this PR.** No data migration: after this release runs on all pods, two admin edits (`SYP/SYP` -> inactive, then `SYP01/SYP01` -> `code=SYP`), keeping all PKs/FKs.
- **Public API after activation.** Input: `SYP01` keeps working through the alias until it is removed (end date TBD). Output: responses return `code=SYP` for the new pound, with `SYP01` only in `vision_code`; clients matching on `"SYP01"` in responses need to switch. FSPs start receiving `SYP` for the new pound.

---

## Problem

Following the Syrian Central Bank's redenomination (effective 2026-01-01, 1 new SYP = 100 old SYP),
external APIs introduced a new internal code `SYP01` for the Syrian Pound and deprecated the old `SYP`.

HOPE's `Currency.code` is unique, so there was nowhere to put the new row. Production worked around
this by storing the upstream (Vision) code in `code` itself:

| name | code | vision_code |
|---|---|---|
| Syrian pound Old | SYP | SYP |
| Syrian pound | SYP01 | SYP01 |

That leaks an upstream-Vision-internal identifier into the user-facing ISO code: `SYP01` shows up in currency
dropdowns, the lookups API and XLSX imports, and users have to know which of the two "currencies" is
the real one. Any future redenomination would repeat the problem.

The two fields have different jobs - `code` is the ISO 4217 code users type and see, `vision_code` is
what Vision/UNORE keys on. Only `vision_code` needs to be globally unique. Both rows should share
`code = SYP`, with the `active` flag saying which one is current:

| name | code | vision_code | active |
|---|---|---|---|
| Syrian pound Old | SYP | SYP | false |
| Syrian pound | SYP | SYP01 | true |

## Solution

- **Constraints** - `code` is no longer unique. Instead: partial unique on `Lower(code)` where
  `active`, and a global unique on `Lower(vision_code)`. At most one active row per code, so
  "give me SYP" stays unambiguous.
- **`CurrencyQuerySet`** - code lookups go through `resolve_code()` / `resolve_code_or_none()`, which
  resolve against the active row. All call sites that looked up a currency by code were converted
  (RDI XLSX/Kobo/people import, API serializers, grievance data-change, universal update script,
  exchange rates).
- **`CurrencySlugRelatedField`** - API serializers that accept a currency code resolve it through the
  same path and return a proper 400 on an unknown/inactive code.
- **Exchange rates / Vision payload** - switched from `code` to `vision_code`, which is what Vision
  actually sends back.
- **Admin** - a `Deprecate` action that atomically swaps `active` between the old and new row of the
  same code, plus `clean()` backfilling `vision_code` so creating the second row from the admin form
  gives a validation error instead of a 500.
- **Lookups endpoint** - now lists only active currencies.


kosior:
@saxix please take a look at the PR description. The model change needs your sign-off: Currency gets a `partial unique on Lower(code) WHERE active` plus a `unique on Lower(vision_code)`. Once that's approved I'll take it forward and fix the remaining items.

https://github.com/unicef/hope/blob/235c1792e28d18deac00e38c6c80e72f0cd72550/src/hope/models/currency.py#L45-L48

src/hope/apps/payment/api/serializers.py
```diff
@@ -621,7 +622,7 @@ class PaymentPlanCreateUpdateSerializer(serializers.ModelSerializer):
     target_population_id = serializers.UUIDField(source="id")
     dispersion_start_date = serializers.DateField()
     dispersion_end_date = serializers.DateField()
-    currency = serializers.SlugRelatedField(slug_field="code", queryset=Currency.objects.all(), allow_null=True)
+    currency = CurrencySlugRelatedField(queryset=Currency.objects.active(), allow_null=True)
```
johniak: This serializer also backs `partial_update` (views.py:776/780), and the FE edit page always sends `currency` back (`EditPaymentPlanPage.tsx:129`). For a plan on the deprecated `SYP` row, GET returns `"SYP"`, PUT resolves it to the active `SYP01` row, and `PaymentPlanService.update` (L768-773) sees a different currency, swaps the plan and bulk-updates every Payment. The user only edited dates. You list this as a TODO, but it ships here. Simplest fix: when `self.parent.instance` already has a currency with the same `code`, return that instance instead of re-resolving.

src/hope/api/utils.py
```diff
@@ -1,8 +1,22 @@
...
 
+class CurrencySlugRelatedField(serializers.SlugRelatedField):
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(slug_field="code", **kwargs)
+
+    def to_internal_value(self, data: Any) -> Any:
+        queryset = self.get_queryset()
+        try:
+            return queryset.resolve_code(data)
+        except queryset.model.DoesNotExist:
```
johniak: Only `DoesNotExist` is caught here. DRF's `SlugRelatedField.to_internal_value` also catches `(TypeError, ValueError)` and fails with `invalid`. With this override a NUL byte in the code (`"USD\x00"`) reaches psycopg and 500s on the public RDI endpoints instead of a 400. Add the second `except` mirroring the base class.

tests/unit/api/test_lookups.py
```diff
@@ -305,6 +305,17 @@ def test_get_currencies(api_client_read: APIClient) -> None:
     assert "next" in response.json()
 
 
+def test_get_currencies_lists_only_the_active_row_for_a_deprecated_code(api_client_read: APIClient) -> None:
+    CurrencyFactory(code="SYP", name="Syrian pound Old", vision_code="SYP", active=False)
```
johniak: Factories are called in the test body. Repo convention is that test data comes only from fixtures, and the other new test files in this PR do exactly that. Move these two into local fixtures.

tests/unit/apps/payment/test_payment_utils.py
```diff
@@ -108,6 +108,24 @@ def test_get_quantity_in_usd_does_not_lookup_when_exchange_rate_provided(
         assert result == Decimal("5.00")
 
 
+def test_get_quantity_in_usd_keys_fx_feed_by_vision_code(django_assert_num_queries) -> None:
+    currency = CurrencyFactory(code="SYP", vision_code="SYP01")
```
johniak: Same as in test_lookups: factory call in the test body. Move it into a local fixture.

tests/unit/models/test_currency_query_patterns.py
```diff
@@ -0,0 +1,118 @@
+import pytest
```
johniak: Most of this file tests `handle_currency_field` / `validate_currency` from `universal_update_script`, so it belongs under `tests/unit/apps/universal_update_script/`. The three `active()` tests at the bottom fit in `test_currency.py`.

johniak (changes requested):

tests/unit/api/test_lookups.py
kosior: Fixed.

src/hope/apps/generic_import/generic_upload_service/importer.py
```diff
@@ -104,7 +104,7 @@ def __init__(
         self._countries = {c.iso_code3: c.id for c in Country.objects.all()}
 
         # Cache Currency lookups (code -> id mapping)
-        self._currencies = {c.code: c.id for c in Currency.objects.all()}
+        self._currencies = {c.code: c.id for c in Currency.objects.active()}
```
nerjan, line 107: Shouldn't we also use `resolve_active_currency_or_none` here?

nerjan (dismissed):

src/hope/api/utils.py
```diff
@@ -1,7 +1,86 @@
...
+            )
+        current = getattr(instance, self.source_attrs[0], None)
+        return current if isinstance(current, self.get_queryset().model) else None
+
+    def to_internal_value(self, data: Any) -> Any:
+        try:
+            return resolve_currency_for_update(data, self._current_currency())
+        except ObjectDoesNotExist:
+            self.fail("does_not_exist", slug_name=self.slug_field, value=smart_str(data))
+        except (TypeError, ValueError):
```
johniak: With psycopg 3 (already in this branch's `pyproject.toml`) a NUL byte in `code` raises `django.db.utils.DataError`, not `ValueError`, so this still 500s on the public RDI endpoints and `test_field_nul_byte_is_validation_error_not_a_driver_error` fails on CI. Guard before the query instead: `if not isinstance(data, str) or "\x00" in data: self.fail("invalid")` (DRF's `CharField` does this via `ProhibitNullCharactersValidator`).

tests/unit/api/test_push_people.py
```diff
@@ -384,6 +384,66 @@ def test_push_single_person_with_currency(
     assert ind.household.currency.code == "USD"
 
 
+@pytest.mark.parametrize(("code", "expected_queries"), [("SYP", 25), ("SYP01", 27)])
```
johniak: These counts no longer match develop: CI does 27/29 here and 22/24 for the twin test in `test_upload.py` (L731). Rebase and re-baseline; as is, the unit job is red.

src/hope/models/currency.py
```diff
@@ -1,33 +1,93 @@
...
+
+        The ``vision_code`` fallback is a transitional alias: after a redenomination moves the new
+        row's ``code`` onto the ISO code, clients may still send its former code, which is kept as
+        ``vision_code``. ``code`` goes first, so an ISO code never resolves through another row's alias,
+        and a deactivated ``code`` stays rejected even when an active row carries it as ``vision_code``.
+        """
+        currency = self.filter(code=code, active=True).first()
+        if currency is not None:
+            return currency
+        # TODO(<ticket>): everything below is the vision_code alias; when the transition period ends,
```
johniak: `TODO(<ticket>)` is a placeholder. Put the real ticket number in or drop the marker.

src/hope/apps/core/attributes_qet_queries.py
```diff
@@ -287,6 +288,14 @@ def get_collector_has_valid_phone_no_query(
     return has_valid_collector_phone if wants_valid else ~has_valid_collector_phone
 
 
+def get_currency_query(comparison_method: str, args: Any, is_social_worker_query: bool = False) -> Q:
+    currency = resolve_active_currency_or_none(args[0])
+    if currency is None:
+        raise ValidationError(f"Unknown currency code: {args[0]}")
```
johniak: A targeting rule saved for a currency that later gets deactivated without a successor (the `currency_retired` case) will raise here on every evaluation, so the whole target population stops resolving. Since the match is on `currency__code` anyway, fall back to `args[0]` when resolution returns `None` instead of raising?

src/hope/admin/currency.py
```diff
@@ -8,5 +11,38 @@
...
+        if len(active) != 1 or len(inactive) != 1:
+            self.message_user(
+                request,
+                "Select exactly two currencies: one selected currency must be active and the other inactive.",
+                level=messages.ERROR,
+            )
+            return
+
+        old, new = active[0], inactive[0]
+        if old.code != new.code:
```
johniak: This compares case-sensitively while the constraint is on `Lower(code)`. `casefold()` both sides.

johniak (changes requested):

domdinicola:
@kosior pls regenerate the conflict migration

src/hope/models/currency.py
```diff
@@ -1,33 +1,93 @@
...
+
+    def get_active_by_code_or_none(self, code: str) -> "Currency | None":
+        """Return the active row whose ``code`` is ``code``, else the active row whose ``vision_code`` is.
+
+        The ``vision_code`` fallback is a transitional alias: after a redenomination moves the new
+        row's ``code`` onto the ISO code, clients may still send its former code, which is kept as
+        ``vision_code``. ``code`` goes first, so an ISO code never resolves through another row's alias,
+        and a deactivated ``code`` stays rejected even when an active row carries it as ``vision_code``.
+        """
+        currency = self.filter(code=code, active=True).first()
```
johniak, line 31: This is exact-match while the constraint is on `Lower(code)` and `deprecate_currency` now compares `casefold()`. After the swap `test_deprecate_treats_codes_differing_only_in_case_as_the_same` sets up (`SYP` inactive, `syp` active), `resolve_active_currency_or_none("SYP")` returns `None`: this filter misses `syp`, and the `filter(code=code).exists()` fallback hits the deprecated `SYP` row. Only `SYP01` still works. Simplest fix: uppercase `code`/`vision_code` in `clean()`/`save()`, or use `code__iexact` here and `casefold()` in `resolve_currency_for_update`.

johniak (changes requested):

src/hope/apps/registration_data/tasks/rdi_xlsx_create.py
```diff
@@ -651,9 +652,7 @@ def _process_regular_field(self, header: str, value: Any, cell: Any, obj_to_crea
             if isinstance(obj_to_create, PendingHousehold):
                 obj_to_create.facility = facility
         elif header == "currency_h_c":
-            from hope.models.currency import Currency
-
-            setattr(obj_to_create, self.COMBINED_FIELDS[header]["name"], Currency.objects.get(code=value))
+            setattr(obj_to_create, self.COMBINED_FIELDS[header]["name"], resolve_active_currency(value))
```
nerjan: Maybe use _resolve_currency_id to use memoized data?

tests/unit/api/test_lookups.py
```diff
@@ -299,12 +300,32 @@ def test_get_currencies(api_client_read: APIClient) -> None:
 
     assert response.status_code == status.HTTP_200_OK
     assert response.json()["results"] == [
-        _currency_response("AFN", "Afghani", False, afn.id),
-        _currency_response("USD", "United States Dollar", False, usd.id),
+        _currency_response(afn),
+        _currency_response(usd),
```
nerjan: I think that direct assertion was better as it was not dependent on what is inside fixture.

kosior: The object-taking form isn't new here - _country_response, _area_response, _area_type_response and _fi_response all took their model instance before this branch. `_currency_response` was the odd one out and this PR aligned it. CurrencyFactory is fully deterministic, so nothing is hidden behind the fixture.

nerjan (approved):

johniak (approved):

PR #6137, opened 2026-06-29
Currency codes for api households

Jeeziorny:
When CW pushes Households to HOPE through API they get errors for `currency`: "got str, expected pk". This PR hotfixes it


johniak (approved):

nerjan (approved):
LGTM
The only think is I would add tests that check what would happen when we pass currency as None and as "", but should work fine anyway

PR #6138, opened 2026-06-30
fix(grievance): offer to close linked Needs Adjudication ticket when its Data Change fix closes

johniak:
## Problem

[AB#320839](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/320839)

When two distinct individuals are registered with the **same national ID**, the system raises a **Needs Adjudication** ticket of type *Unique Identifiers Similarity* (`issue_type = 23`). An operator can correct one individual's ID via a separate **Data Change** ticket — but after that ticket closes and the IDs differ, the original Needs Adjudication ticket **stays open forever**. Nothing re-evaluates or closes it, and the two tickets are not linked.

Reported example (Myanmar, "Multi-Purpose Cash Assistance - KOBO"): NA ticket `GRV-3495774`, linked Data Change ticket `GRV-3495814`.

## Solution

When an operator closes the **Data Change** ticket that corrected an individual's document, and that individual is party to an open Unique Identifiers Similarity ticket whose conflict is now gone, offer a one-click **"mark unique and close"** for the linked ticket right there. The decision is made **after the Data Change ticket's close actually applies the document edit**, by comparing documents live (no re-deduplication), gated by strict guards.

### Guards (all must hold)
1. `issue_type == ISSUE_TYPE_UNIQUE_IDENTIFIERS_SIMILARITY` (24/25 are name/biometric based — not resolved by an ID fix)
2. No shared document signature remains among unresolved individuals (handles partial resolution with >2 individuals)
3. No other open Needs Adjudication ticket involves these individuals (dedup status is cleared globally on the individual)
4. Ticket is open and passes the existing close validation

The action marks every individual on the linked ticket as **distinct** and closes it **directly, bypassing `FOR_APPROVAL`** (intentional — the ticket's sole premise is gone).

## Backend
- **`Document.dedup_signature`** — single source of truth for the dedup signature (`type_id--number--country_id`, honouring `valid_for_deduplication`); `HardDocumentDeduplication._generate_signature` now delegates to it.
- **`TicketNeedsAdjudicationDetails.documents_no_longer_conflict()`** — pairwise, signature-accurate conflict check that ignores invalidated documents. `has_duplicated_document` is left untouched (it only feeds an existing FE display field).
- **`find_open_unique_identifiers_ticket_for_individual()`** — finds the open Unique Identifiers Similarity ticket (if any) an individual is a party to; exposed as `linked_needs_adjudication_ticket_id` on `IndividualDataUpdateTicketDetailsSerializer` so the frontend knows which ticket to offer once the Data Change ticket closes.
- **`mark_unique_and_close()` / `can_close_as_unique()`** services + **`can_close_as_unique`** flag on the Needs Adjudication ticket detail serializer.
- **`POST .../close-as-unique`** action on `GrievanceTicketViewSet` with close permissions, per-individual area-access checks, concurrency check, activity log and cache clear.

## Frontend
- `GrievanceDetailsToolbar`: closing a **Data Change** ticket now checks `ticketDetails.linkedNeedsAdjudicationTicketId`; if present, it re-fetches that ticket, and if `canCloseAsUnique` is now true, asks the operator to mark its individuals as unique and calls `restBusinessAreasGrievanceTicketsCloseAsUniqueCreate` for it. The linked ticket's number in the dialog is a clickable link (opens in a new tab) to its detail page. The Needs Adjudication ticket's own Close button is unchanged from `develop` — closing it directly still goes through the normal status-change flow only.
- The generated REST client (`src/restgenerated`, git-ignored) is regenerated from the backend OpenAPI in CI.

## Tests
`tests/unit/apps/grievance/test_close_needs_adjudication_as_unique.py`: signature branches, conflict detection incl. partial resolution and invalid-document handling, all service guards, the `close-as-unique` API action, `find_open_unique_identifiers_ticket_for_individual` (incl. a query-count assertion), and the Data Change ticket detail exposing `linked_needs_adjudication_ticket_id`.

- ✅ 100% patch coverage on changed backend lines
- ✅ `tox -e mypy`, `tox -e lint` clean; FE `eslint` clean

## Manual testing

Verified end-to-end locally (Afghanistan business area), one Program per scenario, root superuser, clicking through the real UI:

| Scenario | Setup | Result |
|---|---|---|
| Conflict resolved | Data Change ticket edits the shared national ID to a new, unique number | Closing the Data Change ticket shows the "Close linked ticket as unique" dialog; confirming closes the linked Needs Adjudication ticket and marks both individuals distinct |
| Unrelated edit | Data Change ticket edits an unrelated field (`given_name`), document untouched | Dialog does **not** appear (document conflict still holds) |
| Partial multi-duplicate resolution | 3 individuals share the ID; Data Change ticket fixes only one duplicate | Dialog does **not** appear (the other duplicate still conflicts) |
| Sibling open ticket | Conflict resolved, but a second open Needs Adjudication ticket also involves the same individual | Dialog does **not** appear (guard blocks: another open ticket exists) |
| Wrong issue type | Needs Adjudication ticket is `ISSUE_TYPE_BIOGRAPHICAL_DATA_SIMILARITY` (24), not Unique Identifiers Similarity | No link is found at all, no dialog offered |

Also confirmed: closing a Needs Adjudication ticket directly (not via a Data Change ticket) still goes through the plain status-change close flow, unchanged from `develop`.

### DIFF
--- a/src/hope/apps/grievance/api/views.py
+++ b/src/hope/apps/grievance/api/views.py
@@ -72,6 +72,7 @@ from hope.apps.grievance.api.serializers.grievance_ticket import (
     BulkUpdateGrievanceTicketsUrgencySerializer,
     CreateGrievanceTicketSerializer,
     GrievanceChoicesSerializer,
+    GrievanceCloseAsUniqueSerializer,
     GrievanceCreateNoteSerializer,
     GrievanceDeleteHouseholdApproveStatusSerializer,
     GrievanceHouseholdDataChangeApproveSerializer,
@@ -95,6 +96,9 @@ from hope.apps.grievance.models import (
 from hope.apps.grievance.notifications import GrievanceNotification
 from hope.apps.grievance.services.bulk_action_service import BulkActionService
 from hope.apps.grievance.services.data_change_services import update_data_change_extras
+from hope.apps.grievance.services.needs_adjudication_ticket_services import (
+    mark_unique_and_close,
+)
 from hope.apps.grievance.services.payment_verification_services import (
     update_ticket_payment_verification_service,
 )
@@ -409,6 +413,7 @@ class GrievanceTicketGlobalViewSet(
         "create": CreateGrievanceTicketSerializer,
         "partial_update": UpdateGrievanceTicketSerializer,
         "status_change": GrievanceStatusChangeSerializer,
+        "close_as_unique": GrievanceCloseAsUniqueSerializer,
         "create_note": GrievanceCreateNoteSerializer,
         "approve_individual_data_change": GrievanceIndividualDataChangeApproveSerializer,
         "approve_household_data_change": GrievanceHouseholdDataChangeApproveSerializer,
@@ -482,6 +487,11 @@ class GrievanceTicketGlobalViewSet(
             Permissions.GRIEVANCES_CLOSE_TICKET_FEEDBACK,
             Permissions.GRIEVANCES_CLOSE_TICKET_FEEDBACK_AS_CREATOR,
             Permissions.GRIEVANCES_CLOSE_TICKET_FEEDBACK_AS_OWNER,
+            Permissions.GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK,
+            Permissions.GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK_AS_CREATOR,
+            Permissions.GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK_AS_OWNER,
+        ],
+        "close_as_unique": [
             Permissions.GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK,
             Permissions.GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK_AS_CREATOR,
             Permissions.GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK_AS_OWNER,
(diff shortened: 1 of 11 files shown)


MarekBiczysko (dismissed):

src/hope/apps/grievance/models.py
```diff
@@ -952,6 +952,37 @@ def has_duplicated_document(self) -> bool:
...
+        """
+        from hope.models import Document
+
+        selected_duplicates = set(self.selected_individuals.all())
+        unresolved = [
+            individual
+            for individual in [self.golden_records_individual, *self.possible_duplicates.all()]
+            if individual not in selected_duplicates
+        ]
+        if len(unresolved) < 2:
```
pkujawa: I think it handles incorrectly some cases.
So the target case for this is: we have NA ticket with A and B sharing the same document. Someone updates it for B in Data Update ticket and then we want to close the NA with marking them as unique. And this will be fine.
But if we have a NA ticket where B is already marked as selected duplicate, and we make some other change for A in Data Update ticket, we end up marking them as unique too. We exclude this from queryset, because B is a selected duplicate, we return True - documents no longer conflict and we try to close with marking them as unique?
Not sure why we want to exclude the selected_duplicates here. In some places we exclude them because they will be removed anyway but here we need to compare them and we cannot exclude.

src/hope/apps/grievance/models.py
```diff
@@ -952,6 +952,37 @@ def has_duplicated_document(self) -> bool:
...
         return False
 
+    def documents_no_longer_conflict(self) -> bool:
+        """Return True when no two unresolved individuals on the ticket still share a document signature.
+
+        Unlike ``has_duplicated_document`` this uses the real dedup signature
+        (``Document.dedup_signature``) and detects conflicts pairwise, so a partial
+        resolution (e.g. A and B separated while C still shares the id) is correctly
+        reported as a remaining conflict. Invalidated documents are ignored. Callers
+        should prefetch ``documents__type``.
```
pkujawa: the mentioned prefetch is not applied in code.
`needs_adjudication_ticket_details__golden_records_individual__documents__type`
`needs_adjudication_ticket_details__possible_duplicates__documents__type`
maybe we can add to `get_queryset` in `GrievanceTicketViewSet`?

src/hope/apps/grievance/api/views.py
```diff
@@ -832,6 +842,59 @@ def status_change(self, request: Request, *args: Any, **kwargs: Any) -> Response
...
+
+        ticket_details = grievance_ticket.ticket_details
+        if not isinstance(ticket_details, TicketNeedsAdjudicationDetails):
+            raise ValidationError("Ticket is not a Needs Adjudication ticket.")
+
+        partner = user.partner
+        all_individuals = [
+            ticket_details.golden_records_individual,
+            *ticket_details.possible_duplicates.select_related("household__admin2", "program").all(),
+        ]
```
pkujawa:
```
  individual_ids = [                                                                                                                                                                                                                                                                                                         
      ticket_details.golden_records_individual_id,                                                                                                                                                                                                                                                                           
      *ticket_details.possible_duplicates.values_list("id", flat=True),                                                                                                                                                                                                                                                      
  ]                                                                                                                                                                                                                                                                                                                          
  individuals = Individual.objects.filter(id__in=individual_ids).select_related("household__admin2", "program")  
```

src/hope/apps/grievance/api/views.py
```diff
@@ -832,6 +842,59 @@ def status_change(self, request: Request, *args: Any, **kwargs: Any) -> Response
             status=status.HTTP_202_ACCEPTED,
         )
 
+    @transaction.atomic
+    @extend_schema(request=GrievanceCloseAsUniqueSerializer, responses={202: GrievanceTicketDetailSerializer})
+    @action(detail=True, methods=["post"], url_path="close-as-unique")
+    def close_as_unique(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        grievance_ticket = self.get_object()
```
pkujawa, line 855: missing check for owner/creator logic. (user with `GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK_AS_CREATOR` would be able to close any ticket)
```
  if permissions_to_use := self.get_permissions_for_status_change(                                                                                                                                                                                                                                                           
      GrievanceTicket.STATUS_CLOSED,                                                                                                                                                                                                                                                                                         
      grievance_ticket.status,                                                                                                                                                                                                                                                                                               
      grievance_ticket.is_feedback,                                                                                                                                                                                                                                                                                          
  ):                                                                                                                                                                                                                                                                                                                         
      check_creator_or_owner_permission(                                                                                                                                                                                                                                                                                     
          user,                                                                                                                                                                                                                                                                                                              
          permissions_to_use,                                                                                                                                                                                                                                                                                                
          self.business_area,
```

pkujawa (changes requested):

src/hope/apps/grievance/models.py
johniak: Good catch. Dropped the selected_individuals exclusion — we now compare all parties, so a selected duplicate that still shares a document counts as a real conflict and can_close_as_unique stays false. In that case the dialog just doesn't show up, so nothing gets closed as unique.

src/hope/apps/grievance/models.py
johniak: Right, the prefetch only lived in the docstring. Wired it into GrievanceTicketGlobalViewSet.get_queryset: needs_adjudication_ticket_details__golden_records_individual__documents__type and the same for possible_duplicates.

src/hope/apps/grievance/api/views.py
johniak: Done — switched to a single Individual.objects.filter(id__in=[...]).select_related("household__admin2", "program") so the golden record doesn't lazy-load its household/program in the loop.

src/hope/apps/grievance/api/views.py
johniak: Yep, was missing the creator/owner scoping. Added check_creator_or_owner_permission with the close-excluding-feedback perms, same pattern as status_change.

pkujawa (approved):

kosior (approved):

PR #6139, opened 2026-06-30
New selenium tests

mmaciekk:
[AB#325540](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/325540)

Selenium tests covering some hotfixes, the rest will be covered by unit tests.

### DIFF
--- a/tests/e2e/new_selenium/grievance/conftest.py
+++ b/tests/e2e/new_selenium/grievance/conftest.py
@@ -3,8 +3,9 @@ from datetime import datetime
 from dateutil.relativedelta import relativedelta
 import pytest
 
+from extras.test_utils.factories.household import HouseholdFactory
 from extras.test_utils.factories.program import ProgramFactory
-from hope.models import BeneficiaryGroup, BusinessArea, DataCollectingType, Program
+from hope.models import BeneficiaryGroup, BusinessArea, DataCollectingType, Household, Program
 
 
 @pytest.fixture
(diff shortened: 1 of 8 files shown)


johniak (dismissed):

tests/e2e/new_selenium/targeting/test_exclude_individual_ids_social_worker.py
```diff
@@ -0,0 +1,150 @@
...
+    # Program.clean() requires at least one purpose before status=ACTIVE save
+    program.payment_plan_purposes.add(social_purpose)
+    program.status = Program.ACTIVE
+    program.save()
+    return program
+
+
+@pytest.fixture
+def social_people(social_program: Program, business_area: BusinessArea) -> dict:
+    # unicef_id is normally assigned by a DB trigger that isn't active in this suite, so set it
```
nerjan: Are you sure it isn't active here? I don't see a reason for that.

tests/e2e/new_selenium/grievance/test_household_data_update_admin_area.py
```diff
@@ -0,0 +1,52 @@
...
+    browser.click('button[data-cy="button-submit"]')
+
+    browser.wait_for_ready_state_complete()
+    browser.click('[data-cy="input-consent"]')
+    browser.click('button[data-cy="button-submit"]')
+
+    browser.type('textarea[name="description"]', "Admin area update regression #6006")
+    browser.select_dropdown_option("householdDataUpdateFields[0].fieldName", ADMIN_AREA_FIELD_LABEL)
+
+    admin_area_input = f'//label[normalize-space()="{ADMIN_AREA_FIELD_LABEL}"]/following-sibling::div//input'
```
nerjan: we should probably add a name to this and select via input or something like that.

tests/e2e/new_selenium/grievance/conftest.py
```diff
@@ -19,3 +20,30 @@ def social_worker_program(business_area: BusinessArea) -> Program:
         start_date=datetime.now() - relativedelta(months=1),
         end_date=datetime.now() + relativedelta(months=1),
     )
+
+
+@pytest.fixture
+def household_update_program(business_area: BusinessArea) -> Program:
+    beneficiary_group, _ = BeneficiaryGroup.objects.get_or_create(
```
nerjan: BeneficiaryGroupFactory probably better?

nerjan (dismissed):

tests/e2e/new_selenium/grievance/conftest.py
ntrncic: I agree

ntrncic (changes requested):

tests/e2e/new_selenium/grievance/conftest.py
mmaciekk: Done — switched to `BeneficiaryGroupFactory(name="Household Group", ...)` in `grievance/conftest.py`.

tests/e2e/new_selenium/grievance/test_household_data_update_admin_area.py
mmaciekk: Good catch — the shared `FormikAsyncAutocomplete` had no stable attribute on its input (unlike the other Formik fields). Added `data-cy=`input-${field.name}`` to it and the test now selects by `[data-cy="input-householdDataUpdateFields[0].fieldValue"]` instead of the positional XPath.

tests/e2e/new_selenium/targeting/test_exclude_individual_ids_social_worker.py
mmaciekk: You're right, the trigger IS active here (re-applied on post_migrate by `register_custom_sql_signal` in `tests/e2e/conftest.py`). The comment was wrong — corrected it. The explicit ids are still needed, but for determinism: the trigger's ids are serial-based, and the test feeds these exact ids into the targeting rule and asserts on them.

nerjan (approved):

ntrncic (approved):

PR #6146, opened 2026-07-01
ES populate delta script

pavlo-mk:
| Env |  T0 | Delta done | Reconciled | Second Delta done | Second Reconciled |                                                                                                           
  |-----|----------------|-----------|-------------|-------------|-------------|                                                                                                             
  | **EPH-1**| -- | &#9744; | &#9744; | &#9744; | &#9744; |         
  | **PROD** | -- | &#9744; | &#9744; | &#9744; | &#9744; |                                                                                                                                                                          
              
`python manage.py es_populate_delta --using v9 --since 2026-07-01T09:00:00Z --parallel --threads 8`
  | Flag | Purpose |                                                                                                                                                                                                                     
  |------|---------|                                                                                                                                                                                                                     
  | `--using v9` | Target ES connection alias (the new/shadow cluster). |                                                                                                                                                                
  | `--since <ISO-8601>` | Rebuild programs changed at/after this time (use copy **start**, UTC). |                                                                                                                                      
  | `--reconcile` | Rebuild programs whose ES-vs-DB counts disagree (catches hard-deletes). |                                                                                                                                            
  | `--program <code\|uuid>` | Limit to one program (sample-first). |                                                                                                                                                                    
  | `--business-area <slug>` | Limit to one business area. |                                                                                                                                                                             
  | `--include-non-active` | Also include closed/finished programs (default: ACTIVE only). |                                                                                                                                             
  | `--parallel --threads N` | Multi-threaded bulk indexing. |                                                                                                                                                                           
  | `--chunk-size N` | Bulk chunk size (default 2000). |                                                                                                                                                                                 
  | `--verify` | After each rebuild, confirm DB/ES counts match. |                                                                                                                                                                       
  | `--dry-run` | Print what would be rebuilt; no writes. |
 

**Step 0** — record T0 before you start the bulk copy (snapshot/restore or full populate). Use UTC:                                                                                                                       
  `date -u +%Y-%m-%dT%H:%M:%SZ      # e.g. 2026-07-01T09:00:00Z  ← this is your T0`                                                                                                                                       
                                                                                                                                                                                                                        
  **Step 1** — first delta pass (after the bulk copy finishes): rebuild every program changed since T0.                                                                                                                     
  `python manage.py es_populate_delta --using v9 --since 2026-07-01T09:00:00Z --parallel --threads 8`                                                                                                                     
                                                                                                                                                                                                                        
  **Step 2** — see what would change before committing (dry run, safe, no writes):                                                                                                                                          
  `python manage.py es_populate_delta --using v9 --since 2026-07-01T09:00:00Z --dry-run`                                                                                                                                  
  Prints each program + the reason (changed / count-mismatch).                                                                                                                                                          
                                                                                                                                                                                                                        
  **Step 3** — second delta pass to shrink the window (use a fresh, later timestamp — the time your first delta started):                                                                                                   
  `python manage.py es_populate_delta --using v9 --since 2026-07-01T11:30:00Z`                                                                                                                                            
  Re-run until it reports Programs to rebuild: 0 — that means the cluster has caught up.                                                                                                                                
                                                                                                                                                                                                                        
  **Step 4** — final convergence check before cutover (the "be sure all data are the same" pass):                                                                                                                           
  `python manage.py es_populate_delta --using v9 --reconcile --verify`                                                                                                                                                    
  This compares ES-vs-DB counts for every active program, rebuilds any mismatch (catches hard-deletes), and --verify re-checks counts after each rebuild.                                                               
                                                                                                                                                                                                                        
  **_Other useful variants_**                                                                                                                                                                                                 
                                                                                                                                                                                                                        
  Combine delta + reconcile in one run:                                                                                                                                                                                 
  `python manage.py es_populate_delta --using v9 --since 2026-07-01T11:30:00Z --reconcile --verify`                                                                                                                       
                                                                                                                                                                                                                        
  Include closed/finished programs (default is ACTIVE only):                                                                                                                                                            
  `python manage.py es_populate_delta --using v9 --reconcile --include-non-active`                                                                                                                                        
                                                                                                                                                                                                                        
  Tune bulk performance for large programs:                                                                                                                                                                             
```
  python manage.py es_populate_delta --using v9 --since 2026-07-01T09:00:00Z \                                                                                                                                          
    --parallel --threads 8 --chunk-size 4000 
```                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                        
  **_Notes / gotchas_**
    - --since = when the copy started (T0), ideally minus ~1 min of buffer. The delta is idempotent, so overlap is harmless — better to re-index a few extra programs than miss one.                                      
  - Pass UTC (...Z). A naive timestamp is interpreted in the server's current timezone; UTC avoids ambiguity.                                                                                                           
  - You must pass --since and/or --reconcile — running with neither errors out on purpose.                                                                                                                              
  - Prerequisite reminder: --using v9 only actually writes to the ES9 cluster once the stash machinery (v9 connection registration + using= threading) is applied and ELASTICSEARCH_HOST_V9 is set; otherwise the alias 
  won't resolve and the command will tell you 'v9' is not registered in settings.ELASTICSEARCH_DSL.


Jeeziorny:
Here is the URL for ES Blue-green deployment: https://unicef.sharepoint.com/teams/EMOPS-HCT-MIS/Dev%20Team/jan/reindex-playbook.md we agreed on a few weeks ago. I'll base my review on this doc, because this management command will be reused in future for other ES migrations

src/hope/apps/household/management/commands/es_populate_delta.py
```diff
@@ -0,0 +1,216 @@
+"""Reconcile a shadow Elasticsearch cluster with Postgres after the bulk copy.
+
+This is the *delta / catch-up* step of the ES 8 -> 9 migration. The bulk copy
+(snapshot+restore of the ES 8 data into the fresh ES 9 cluster, or a full
+``es_shadow_populate`` run) is a point-in-time snapshot. Records keep changing in
+Postgres while that copy runs, so the shadow cluster drifts. This command brings
+it back in sync.
```
Jeeziorny:
```suggestion
Records keep changing in Postgres while the ES Index copy runs, so the shadow cluster drifts. This command brings it back in sync.
```

src/hope/apps/household/management/commands/es_populate_delta.py
```diff
@@ -0,0 +1,216 @@
...
+from django.utils import timezone
+
+from hope.apps.household.services.index_management import (
+    check_program_indexes,
+    rebuild_program_indexes,
+)
+
+
+class Command(BaseCommand):
+    help = "Delta/reconcile per-program ES indexes on a shadow cluster against Postgres."
```
Jeeziorny:
```suggestion
    help = "Delta/reconcile per-program ES indexes on a shadow cluster against Postgres. Shadow cluster is the one that waits to be plugged into prod and replace the one that is currently used."
```

src/hope/apps/household/management/commands/es_populate_delta.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def add_arguments(self, parser: Any) -> None:
```
Jeeziorny, line 75: Can we add --program <id> and --busines-area <slug> params? It would be good if we could rebuild first for smaller sample

src/hope/apps/household/management/commands/es_populate_delta.py
```diff
@@ -0,0 +1,216 @@
...
+        if failures:
+            raise CommandError(f"Done with {failures} failure(s) on cluster '{using}'.")
+        self.stdout.write(self.style.SUCCESS(f"Done. Cluster '{using}' is in sync for the processed programs."))
+
+    @staticmethod
+    def _count_mismatch_ids(scope_id_set: set, using: str) -> set:
+        return {pid for pid in scope_id_set if not check_program_indexes(str(pid), using=using)[0]}
+
+    def _print_dry_run(self, to_rebuild: set, changed_ids: set, mismatch_ids: set, code_by_id: dict) -> None:
+        for pid in sorted(to_rebuild, key=lambda p: code_by_id.get(p, "")):
```
Jeeziorny, line 230: What's the +/- volume of `to_rebuild` on prod? If it's more than 10000, maybe we should print some summary only without printing every pid? It's more of a question than a suggestion. If it write 1000000000 rows on the console I'm just not sure if someone would read it.

src/hope/apps/household/management/commands/es_populate_delta.py
```diff
@@ -0,0 +1,216 @@
...
+            reasons = [r for r, group in (("changed", changed_ids), ("count-mismatch", mismatch_ids)) if pid in group]
+            self.stdout.write(f"  - {code_by_id.get(pid)}  id={pid}  ({', '.join(reasons)})")
+
+    def _rebuild_programs(self, to_rebuild: set, code_by_id: dict, using: str, opts: dict) -> int:
+        failures = 0
```
Jeeziorny: I think it's usable to have information about which pids failed, not only how many of them failed. WDYT?

Jeeziorny (changes requested):

src/hope/apps/household/management/commands/es_populate_delta.py
pavlo-mk: make sense 👍

src/hope/apps/household/management/commands/es_populate_delta.py
pavlo-mk: updating prod DB is huge for sure. make sense to have it more readable and compact

src/hope/apps/household/management/commands/es_populate_delta.py
pavlo-mk: now failure run tells us exactly what to re-target

pavlo-mk: updated

PR #6151, opened 2026-07-03
[326506]  Use autocomplete for all admin ForeignKey fields (register FileTemp, PaymentPlanSplit, FundsCommitmentGroup)

pkujawa:
[AB#326506](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/326506)
Django admin FKs only render as autocomplete when the related model's admin is registered with `search_fields`; otherwise they fall back to a plain `<select>` dropdown. Several models used as FK targets weren't registered, so their FKs (e.g. `WesternUnionData.file`) rendered as dropdowns.                          
                                                                                                                                                                                                                                                                                                                             
  This registers the missing target admins so every FK across the mixin-based admins is now autocomplete.                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                             
  ## Changes                                                                                                                                                                                                                                                                                                                 
  - Register `FileTemp` admin with `search_fields` — fixes the `FileTemp` FK on Western Union Data, Western Union Invoice/Report, Payment Plan (export/import files), Payment Plan Group, Follow-up Instruction, etc.                                                                                                        
  - Register `PaymentPlanSplit` admin — fixes `Payment.parent_split`.                                                                                                                                                                                                                                                        
  - Register `FundsCommitmentGroup` admin — fixes `FundsCommitmentItem.funds_commitment_group` (also enables its existing `AutoCompleteFilter`).                                                                                                                                                                             
  - Consolidate the file **Download** link onto the `FileTemp` admin and remove the duplicated `download_link` methods from the Western Union and Follow-up Instruction admins.                     
  
`FileTemp` has encrypted `password` / `xlsx_password` fields (they decrypt transparently on read). Since `EncryptedTextField` would expose the plaintext to anyone with `view_filetemp`, these two fields are **root-only** (superuser + valid `x-root-token`): excluded from the form for everyone else, readonly for root.


kosior (approved):

nerjan (approved):

src/hope/admin/file_temp.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@admin.register(FileTemp)
+class FileTempAdmin(HOPEModelAdminBase):
+    list_display = ("file", "download_link", "content_type", "object_id", "was_downloaded", "created")
+    list_filter = ("was_downloaded", "content_type")
+    search_fields = ("file", "object_id")
+    ordering = ("-created",)
+
+    def get_readonly_fields(self, request: HttpRequest, obj: FileTemp | None = None) -> tuple[str, ...]:
```
domdinicola, line 17: good!

PR #6152, opened 2026-07-03
320573_Nigeria_Enhance_Aurora_RDI_creation_script

MarekBiczysko:
[AB#320573](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/320573)

Adds Nigeria Aurora RDI duplicate account handling.

- Skips Nigeria Aurora records when their account number already exists in the program, including accounts created earlier in the same import batch.
- Marks skipped records as ignored with a reason and prevents ignored-only imports from remaining stuck in importing status.
- Adds a Nigeria-only Registration admin CSV export for ignored Aurora records.


src/hope/contrib/aurora/services/nigeria_people_registration_service.py
```diff
@@ -30,6 +32,68 @@
...
+            "national_id_no_i_c": "document.doc_national-document_number",
+        },
+    }
+
+    @classmethod
+    def get_mapping(cls, registration_mapping: dict | None) -> dict:
+        return mergedicts(copy.deepcopy(cls.DEFAULT_MAPPING), registration_mapping or {}, [])
+
+    @classmethod
+    def _get_account_number_from_account_data(cls, account_data: dict) -> str:
```
pkujawa: seems it is not used?

johniak (dismissed):

PR #6157, opened 2026-07-06
[323047] Fix RDI Merge Deadlock

kosior:
https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/323047

### Problem

- Production RDI merges intermittently fail with deadlock detected ... while locking tuple in relation "household_household".
- Concurrent transactions (RDI merge, grievance data-change ticket close, recalculate_population_fields) acquire row locks on the same households/individuals in inconsistent order, forming lock cycles:
  - grievance services locked the individual first, then its household - the reverse of recalculate_data (household → individuals).
  - the merge's collision handling wrote to pre-existing (merged) households in collision-encounter order, with locks held for the whole merge transaction.
  - the recalc chunk task held locks from every processed page until the end of the job.

### Solution

- Establish a single lock-ordering convention everywhere: household before individual, pk order within a set - with a consistent order, lock cycles should be impossible and conflicts degrade into plain waits.
- Acquire all collision-target locks up front in one pk-ordered statement, before any collision writes.
- Shorten how long locks are held where possible (per-page transactions in the recalc chunk task).

### Backend

- Grievance data-change services (individual_data_update_service, individual_delete_service, add_individual_service): lock the household before the individual on ticket close; bulk individual locks are pk-ordered.
- RdiMergeTask.execute: households are locked before individuals (same order as recalculate_data).
- RdiMergeTask._process_collisions restructured into detect → lock → write:
  - detection collects only ids
  - one select_for_update().only("pk").order_by("pk") statement acquires every collision-target lock deterministically;
  - collided pending households are processed in chunks of 1000; extra_rdis is set via a single bulk extra_hh_rdis.add(*ids) - net fewer queries than before.
- recalculate_population_fields_chunk_task: one transaction per page instead of one for the whole job, so household/individual locks are released page by page.
- Tests: new unit tests (add-individual HEAD reassignment, lock-order coverage); query-count assertions recalibrated (test_update_individual_data_service: +1 for the household lock; test_cw_dedup_flow_e2e: −1 from the batched collision writes).

---

### Possible follow-ups (not in this PR)

Retry on deadlock

- Lock ordering makes deadlocks between the fixed code paths impossible, but not against paths outside the convention (admin actions, ad-hoc scripts, future code)

Shorten the merge transaction's lock hold

- The merge transaction currently holds household/individual locks across work that doesn't need them: Elasticsearch dedup + needs-adjudication tickets, sanction-list check, biometric dedup, ES index population.
- Move the non-DB / external-service steps out of the atomic block (before it, or into transaction.on_commit), keeping the transaction to just the DB writes: collision processing, status flips, collections.


src/hope/apps/household/celery_tasks.py
```diff
@@ -38,8 +38,8 @@ def recalculate_population_fields_chunk_async_task_action(job: AsyncJob) -> None
 
     with disable_concurrency(Household), disable_concurrency(Individual):
         program = Program.objects.get(id=program_id) if program_id else None
-        with transaction.atomic():
-            for page in paginator.page_range:
+        for page in paginator.page_range:
+            with transaction.atomic():
```
kosior, line 42: I think this should be ok, but please double check.

src/hope/apps/registration_data/tasks/rdi_merge.py
```diff
@@ -59,17 +59,25 @@
...
-            households_chunks = PendingHousehold.objects.filter(id__in=ids)
-            for household in households_chunks:
-                collided_id = obj_hct.program.collision_detector.detect_collision(household)
-                if not collided_id:
-                    continue
-                household_ids_to_exclude.append(household.id)
-                obj_hct.program.collision_detector.update_household(household)
-                updated_household = Household.objects.get(id=collided_id)
-                updated_household.extra_rdis.add(obj_hct)
+            for household in PendingHousehold.objects.filter(id__in=ids):
```
nerjan: maybe worth adding `.only("id", "identification_key")` for better performance

src/hope/apps/registration_data/tasks/rdi_merge.py
```diff
@@ -59,17 +59,25 @@
 
 class RdiMergeTask:
     def _process_collisions(self, obj_hct: RegistrationDataImport, household_ids: list) -> tuple[list, list]:
```
nerjan, line 61: Would be good to add test for that.
We have some in test_rdi_merge_collision.py but we should test with more households.

nerjan (changes requested):
LGTM, would be nice to have some test exercising this deadlock - that it actually not occure

domdinicola:
@kosior pls resolve conflicts, and address review

src/hope/apps/registration_data/tasks/rdi_merge.py
kosior: Added.

src/hope/apps/registration_data/tasks/rdi_merge.py
kosior: Done.

nerjan (dismissed):

johniak (changes requested):
I see that in several places we lock the household first, then the individual, and the convention only lives in comments. Maybe we could add a helper, e.g. src/hope/apps/household/services/locking.py, so the ordering is enforced by code rather than by comments?

```
def lock_household_then_individual(individual: Individual) -> tuple[Household | None, Individual]:
    household = None
    if individual.household_id:
        household = Household.objects.select_for_update().get(id=individual.household_id)
    locked_individual = Individual.objects.select_for_update().get(id=individual.id)
    if locked_individual.household_id != individual.household_id:
        household = (
            Household.objects.select_for_update().get(id=locked_individual.household_id)
            if locked_individual.household_id
            else None
        )
    return household, locked_individual

```

MarekBiczysko (dismissed):

kosior:
> I see that in several places we lock the household first, then the individual, and the convention only lives in comments. Maybe we could add a helper, e.g. src/hope/apps/household/services/locking.py, so the ordering is enforced by code rather than by comments?
> 
> ```
> def lock_household_then_individual(individual: Individual) -> tuple[Household | None, Individual]:
>     household = None
>     if individual.household_id:
>         household = Household.objects.select_for_update().get(id=individual.household_id)
>     locked_individual = Individual.objects.select_for_update().get(id=individual.id)
>     if locked_individual.household_id != individual.household_id:
>         household = (
>             Household.objects.select_for_update().get(id=locked_individual.household_id)
>             if locked_individual.household_id
>             else None
>         )
>     return household, locked_individual
> ```

@johniak I've added the locking service. I didn't add the protection against the reassignment race though - it would violate lock household before individual rule. We could raise exception when happens and retry the action. Can we move an already-merged individual between households? I didn't find that scenario.

nerjan (approved):

johniak (approved):

PR #6187, opened 2026-07-14
fix(frontend): clean up grievances filter styling & console warnings

mmaciekk:
## Summary

Cleans up styling and eliminates React console warnings on the Grievances (and People) filter UI.

## Changes

- **`SelectFilter`** — `borderRadius` is now applied via a transient `$borderRadius` prop instead of leaking through `...otherProps` onto the DOM. Fixes the `React does not recognize the borderRadius prop on a DOM element` warning.
- **`BulkBaseModal`** — replaced the non-tabular `<Table>`/`<TableBody>` wrappers with a `styled.div`, fixing the `<p> cannot appear as a descendant of <p>` DOM-nesting warning when opening the bulk-action modals. `data-cy` hooks are preserved.
- **`SearchTextField`** — migrated the deprecated `inputProps`/`InputProps` to MUI 7 `slotProps` (`htmlInput` / `input`).
- **Corner rounding** — the standalone Search field and the People document-type select are now fully rounded; the Document Type / Document Number pair renders as a proper outward-rounded joined pill.


nerjan (dismissed):

src/frontend/src/components/core/SelectFilter.tsx
```diff
@@ -27,10 +27,13 @@ const SelectWrapper = styled.div`
   max-width: 100%;
 `;
 
-const StyledSelect = styled(Select)`
+const StyledSelect = styled(Select)<{ $borderRadius?: string }>`
   && .MuiOutlinedInput-input {
     padding-right: 10px !important;
   }
+  .MuiOutlinedInput-root {
+    border-radius: ${(props) => props.$borderRadius};
```
johniak, line 35: This rule won't apply. `styled(Select)` puts the generated class directly on the `.MuiOutlinedInput-root` element (MUI merges it there via `clsx(..., className, classes.root)`), and border-radius lives on that same root. A bare `.MuiOutlinedInput-root { ... }` compiles to a descendant selector (`.<generated> .MuiOutlinedInput-root`), so it looks for a nested root that doesn't exist and matches nothing. It works in `SearchTextField` only because `styled(TextField)` lands the class on an outer `FormControl`, making the root a real descendant there.

Consequence: the grievances `DocumentSearchField` Document Type select keeps MUI's default full 4px rounding instead of `4px 0px 0px 4px`, so the joined Type/Number pill this PR targets still has a rounded seam. Target the root itself:

```suggestion
  && {
    border-radius: ${(props) => props.$borderRadius};
  }
```

(This matches how `BusinessAreaSelect`, `KoboProjectSelect`, and the RDI `ComboBox` already style `styled(Select)` roots with `&`/`&&`.)

johniak (changes requested):
The new border-radius rule in SelectFilter targets the wrong element, so the grievances Document Type select never gets its left-only rounding and the joined document pill this PR is meant to fix still renders broken. One-character specificity fix needed.

src/frontend/src/components/core/SelectFilter.tsx
mmaciekk: Good catch, fixed in 19748a0a5f. Swapped the dead `.MuiOutlinedInput-root { … }` descendant rule for `&& { … }` so it targets the `styled(Select)` root directly — matching the `BusinessAreaSelect` precedent you cited. The Document Type select now gets its `4px 0px 0px 4px` left-only rounding and the joined Type/Number pill renders without the seam.

johniak (approved):

PR #6188, opened 2026-07-14
317775: delete rdi endpoint

Jeeziorny:
# Initial Problem:
When something goes wrong during the RDI upload, Country Workspace CAN'T reupload the same RDI because in HOPE country_workspace_id attribute of RDI needs to be UNIQUE. The approach we took is if something goes wrong, instead of UPDATING RDI we REVERTING THE WHOLE OPERATION, so that CW can start with given RDI from scratch: creating, uploading INDs, HHs and COMPLETEing it

# Technical spec that we agreed on:
1. DELETE RDI removes EVERYTHING RELATED TO RDI WITH THE RDI ITSELF. CW would need to start over from CreateRDI endpint. There will be no unique constraint violation.
2. DELETE RDI can remove RDI in ALL STATES EXCEPT MERGED and MERGING.
3. ENDPOINT SIGNATURE: POST rdi/<uuid:rdi>/reset/

# Update (19 Aug)
Initial problem was solved, but after review other problems appeared.
1. In case of big RDIs we need to perform the RDI delete in async. This PR adds async layer to delete logic.
2. Endpoint is called RESET instead of DELETE: from CW POV it's just freeing cw id for RDI. They don't need to know that HOPE removes anything.
3. When HOPE succesfully RESETs (deletes) an RDI it needs to tell the CW this. It'll use callback provided in CW RESET request.
4. @ntrncic requested to leave static files for now and implement the delete logic without it. Static files handling will be changing in following weeks.

# Update (24 Aug)
RDI Delete cascade triggers a lot of ES and REDIS calls. Mechanism with ContextVar and muting signals was used, details at https://github.com/unicef/hope/pull/6326


src/hope/api/endpoints/rdi/base.py
```diff
@@ -226,3 +241,69 @@ def update(self, request: Request, *args: Any, **kwargs: Any) -> Response:
                 {"id": self.selected_rdi.pk, "status": self.selected_rdi.status},
             ]
         )
+
+
+def remove_rdi_cascade(rdi: RegistrationDataImport) -> None:
```
ntrncic: Can we have a docstring what this does.

src/hope/api/endpoints/rdi/base.py
```diff
@@ -226,3 +241,69 @@ def update(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+
+class DeleteRDIView(BusinessAreaIngestCWOnlyMixin, HOPEAPIBusinessAreaView):
+    """Api to hard-delete an RDI (and everything tied to it) for a CW-managed business area.
+
+    Frees the RDI-level unique ``country_workspace_id`` constraint and, via the
+    cascaded individual deletes, the app-level ``/push/people/`` ``country_workspace_id`` guard
+    (re-PUSH). Deletable in every status except ``MERGED`` (terminal -> 409).
+    """
+
+    permission = Grant.API_RDI_CREATE
```
ntrncic: we should have a new permission API_RDI_DELETE and we should log this, i think we already do which token hits what api  endpoint, but please double check and maybe add to test.

ntrncic (changes requested):
One question as well, what if it is in merging state? how do we kill that task and then delete?

src/hope/api/endpoints/rdi/base.py
nerjan: It seems there is partially similar code for refuse/erase of RegistrationDataImportViewSet - there we also check for biometric deduplication, here it is missing.
If biometric deduplication should be here, it would be good idea to extract this as some helper function for this 3 places and use so we do not forget about every step that needs to be taken for such operations.

src/hope/api/endpoints/rdi/base.py
```diff
@@ -226,3 +241,69 @@ def update(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+    def delete(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        with atomic():
+            try:
+                rdi = (
+                    RegistrationDataImport.objects.select_for_update(of=("self",))
+                    .select_related("program")
+                    .get(id=self.kwargs["rdi"], business_area=self.selected_business_area)
+                )
+            except RegistrationDataImport.DoesNotExist:
+                raise Http404
```
nerjan: get_object_or_404

src/hope/api/endpoints/rdi/base.py
```diff
@@ -226,3 +241,69 @@ def update(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+
+            if rdi.status == RegistrationDataImport.MERGED:
+                return Response({"error": "rdi_already_merged"}, status=status.HTTP_409_CONFLICT)
+
+            try:
+                remove_rdi_cascade(rdi)
+            except ProtectedError:
+                set_rollback(True)
+                return Response({"error": "rdi_has_dependents"}, status=status.HTTP_409_CONFLICT)
```
nerjan: I think that set_rollback(True) is overkill and is Low-level APIs (https://docs.djangoproject.com/en/6.0/topics/db/transactions/#low-level-apis). It is better to wrap whole atomic operation in try/except catching ProtectedError - when atomic fails it rollbacks automatically and we catch the error outside atomic like:
```
try:
    with atomic():
        rdi = get_object_or_404(
            RegistrationDataImport.objects.select_for_update(of=("self",)).select_related("program"),
            id=self.kwargs["rdi"],
            business_area=self.selected_business_area,
        )
        if rdi.status == RegistrationDataImport.MERGED:
            return Response({"error": "rdi_already_merged"}, status=status.HTTP_409_CONFLICT)
        remove_rdi_cascade(rdi)
except ProtectedError:
    return Response({"error": "rdi_has_dependents"}, status=status.HTTP_409_CONFLICT)
return Response(status=status.HTTP_204_NO_CONTENT)```

**Review by nerjan — CHANGES_REQUESTED** [review]

**Review by johniak — APPROVED** [review]
Looks good

**Comment by Jeeziorny** [comment]
> One question as well, what if it is in merging state? how do we kill that task and then delete?

Let's go deeper and consider each RDI state and what's the impact of deleting it. 

Country Workspace RDIs can be only in four states:
1. LOADING - mid-upload state. DELETE RDI simply removes HHs / INDs / ES docs -> OK.
2. MERGE_SCHEDULED - It's a state for RDI that means: RDI is ready to be merged, but we're waiting for the worker to take it and merge it. RDI shouldn't be in this state for too long. In practice It differs from LOADING only on RDI status, so cleanup is the same as for LOADING and thus delete is safe.
3. MERGING - RDI is being processed. Different kind of operation are taking place on DB and we SHOULD NOT interrupt it. 
4. MERGED = COMPLETED -> RDIs can't be removed (at least not this way) when they are in MERGED.

FYI @ntrncic

File: src/hope/api/endpoints/rdi/base.py (hunk 1 above)
Comment (Jeeziorny) [reply src/hope/api/endpoints/rdi/base.py]: @nerjan 

> there we also check for biometric deduplication
Biometric is only for CW, CW is enforced by mixin in view, but good catch.

I moved the logic to the removal service.. I'm not sure if this is the best option, but I can't find better one. I feel like keeping this logic between views is worse. @johniak FYI, please check it guys and tell me WDYT.

File: src/hope/api/endpoints/rdi/base.py (hunk 3 above)
Comment (Jeeziorny) [reply src/hope/api/endpoints/rdi/base.py]: Logic has changed, pls re review

**Review by ntrncic — APPROVED** [review]

File: src/hope/api/endpoints/rdi/base.py [hunk 5]
```diff
@@ -226,3 +240,37 @@ def update(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+                )
+                try:
+                    rdi = base_qs.select_for_update(skip_locked=True, of=("self",)).select_related("program").get()
+                except RegistrationDataImport.DoesNotExist:
+                    if base_qs.exists():
+                        return Response({"error": "rdi_merge_in_progress"}, status=status.HTTP_409_CONFLICT)
+                    raise Http404
+
+                if rdi.status == RegistrationDataImport.MERGED:
+                    return Response({"error": "rdi_already_merged"}, status=status.HTTP_409_CONFLICT)
```
nerjan, line 266: Are API users aware of this endpoint returning 409?

src/hope/apps/registration_data/services/rdi_removal.py
```diff
@@ -0,0 +1,64 @@
...
+    A deletable RDI is always non-merged and pending rows are only indexed at merge time, so in the
+    normal recovery flow there is usually nothing to remove here. When a stale ("ghost") document does
+    survive it is never read by RDI create or push — both pure Postgres writes — and it is keyed on the
+    deleted row's random UUID, so a re-upload writes a fresh document instead of colliding with it.
+    Only deduplication reads these documents and can transiently false-match a ghost, but HOPE
+    reconciles every Elasticsearch hit against Postgres, resets the individual to UNIQUE and deletes the
+    stale doc. A leftover document is therefore cosmetic and self-healing, so an Elasticsearch failure
+    here must never roll back the RDI delete.
+    """
+    if not program or program.status != Program.ACTIVE:
```
nerjan: This comment it hard to understand, pls make shorter - check other as well.

src/hope/apps/registration_data/services/rdi_removal.py
nerjan, line 1: test this file

nerjan (changes requested):

src/hope/api/endpoints/rdi/base.py
Jeeziorny: @vitali-yanushchyk-valor are we ok with 409? @ntrncic fyi

vitali-yanushchyk-valor: Using `409` with different error codes looks good to me. I just need to handle these cases explicitly on the CW side, because currently CW treats any non-2xx response from this endpoint as a failure.

Jeeziorny:
Why HOPE did not report 404 when they get it from DE while merging? TBD

Jeeziorny:
RDI.Country workspace id must be read only in admin panel

src/hope/api/endpoints/rdi/base.py
```diff
@@ -226,3 +240,37 @@ def update(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+    (re-PUSH). Deletable in every status except ``MERGED`` (terminal -> 409). An in-flight merge
+    holding the row lock is not waited on -> 409 ``rdi_merge_in_progress`` (retryable).
+    """
+
+    permission = Grant.API_RDI_DELETE
+
+    def delete(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        try:
+            with atomic():
+                base_qs = RegistrationDataImport.objects.filter(
```
johniak, line 253: This matches any RDI in the BA, not just CW-sourced ones. A BA migrated from legacy ingest to CW-only keeps its old UI imports (xlsx/Kobo, e.g. IN_REVIEW with real pending data), and the CW token can hard-delete them even though CW never owned them. Those legacy RDIs are also merged by the legacy task that never locks the RDI row, so the skip_locked guard doesn't cover them. Filter on `country_workspace_id__isnull=False` (or check `rdi.is_coming_from_cw`) and 404/409 otherwise.

src/hope/api/endpoints/rdi/base.py
```diff
@@ -17,12 +19,24 @@
...
+    Program,
+    RegistrationDataImport,
+    User,
+)
 
 if TYPE_CHECKING:
     from hope.models import BusinessArea
 
 
+logger = logging.getLogger(__name__)
```
johniak: `logger` isn't used anywhere in this module — drop it together with the `logging` import.

tests/unit/api/test_delete_rdi.py
```diff
@@ -0,0 +1,356 @@
...
+        dbname=db["NAME"],
+        user=db["USER"],
+        password=db["PASSWORD"],
+        host=db["HOST"],
+        port=db["PORT"],
+    )
+    try:
+        with locker.cursor() as cur:
+            cur.execute(
+                "SELECT id FROM registration_data_registrationdataimport WHERE id = %s FOR UPDATE",
```
johniak: Instead of hardcoding the table name, use `RegistrationDataImport._meta.db_table` in the SQL — if the table is ever renamed, this test would silently lock the wrong (nonexistent) row and fail confusingly.

johniak (changes requested):

src/hope/apps/registration_data/api/views.py
```diff
@@ -160,28 +157,11 @@ def erase(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
-        households_to_remove = list(
-            Household.all_objects.filter(registration_data_import=rdi).values_list("id", flat=True)
-        )
-        hoh_to_remove = list(
-            Household.all_objects.filter(registration_data_import=rdi).values_list("head_of_household_id", flat=True)
-        )
-        Household.all_objects.filter(registration_data_import=rdi).update(head_of_household=None)
-        Individual.all_objects.filter(id__in=hoh_to_remove).delete()
-        Household.all_objects.filter(registration_data_import=rdi).delete()
+        remove_rdi_population(rdi, delete_rdi=False)
```
johniak, line 160: `_remove_rdi_elasticsearch_documents` wraps the ES call in `try/except`, so if Elasticsearch throws, the error is swallowed and never reaches `@transaction.atomic` — the delete commits instead of rolling back. That's what you want for the new DELETE, but erase/refuse go through the same service and used to roll back on an ES failure. Intended, or should the swallow be delete-only?

tests/unit/api/test_delete_rdi.py
```diff
@@ -0,0 +1,366 @@
...
+    other = RegistrationDataImportFactory(
+        business_area=BusinessAreaFactory(name="Ukraine"),
+        status=RegistrationDataImport.LOADING,
+    )
+    response = token_api_client.delete(reverse("api:rdi-delete", args=[user_business_area.slug, str(other.id)]))
+    assert response.status_code == status.HTTP_404_NOT_FOUND
+    assert RegistrationDataImport.objects.filter(id=other.id).exists()
+
+
+def test_delete_cascades_hh_ind(
```
johniak, line 141: None of the new tests pin a query count for the cascade delete. `remove_rdi_population` deletes households/individuals and collects their ids for ES, which is a classic N+1 surface, and right now nothing asserts the number of queries. Wrap at least this cascade test (and the service test) in `django_assert_num_queries` so the expected count is pinned and a per-row regression fails the test instead of slipping through.

johniak (changes requested):

tests/unit/api/test_delete_rdi.py
Jeeziorny: Fixed!

src/hope/apps/registration_data/api/views.py
Jeeziorny: You're right. I preserved the old behaviour in erase / refuse. ES won't rollback on Delete endpoint though

src/hope/apps/registration_data/services/rdi_removal.py
```diff
@@ -0,0 +1,56 @@
...
+        Household.all_objects.filter(registration_data_import=rdi).values_list("id", "head_of_household_id")
+    )
+    individual_ids = list(Individual.all_objects.filter(registration_data_import=rdi).values_list("id", flat=True))
+    household_ids = [row[0] for row in household_rows]
+    hoh_ids = [row[1] for row in household_rows if row[1] is not None]
+
+    Household.all_objects.filter(registration_data_import=rdi).update(head_of_household=None)
+
+    if delete_rdi:
+        rdi.delete()
```
johniak, line 46: Should this hard-delete also remove attached files and photos from storage? Django cascades only delete the database rows, and the existing flows appear to leave FileField/ImageField objects in storage as well. Since this endpoint explicitly deletes the entire RDI population, we should make a conscious decision about beneficiary photos, certificates, document photos, consent signatures, and XlsxUpdateFile files, then document or implement that retention policy.

johniak (approved):

nerjan (approved):

MarekBiczysko (approved):

vitali-yanushchyk-valor (approved):

vitali-yanushchyk-valor:
> * **Callback auth** — how HOPE authenticates to the CW callback endpoint and how CW verifies the call is genuinely from HOPE (shared secret / token), and where that secret lives.

I thought about use signed url for auth

src/hope/api/endpoints/rdi/base.py
```diff
@@ -226,3 +241,64 @@ def update(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
         )
+
+
+class RDIResetSerializer(serializers.Serializer):
+    callback_url = serializers.URLField(required=True)
+    signed_token = serializers.CharField(required=True)
+
+    def validate_callback_url(self, value: str) -> str:
+        parts = urlsplit(value)
+        if parts.scheme not in ("http", "https"):
```
MarekBiczysko: DRF's URLField uses Django URL validation and HTTP/HTTPS are the standard accepted schemes, so probably no need for this explicit check

src/hope/api/endpoints/rdi/base.py
```diff
@@ -226,3 +241,64 @@ def update(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+                logger.info("RDI reset conflict for %s: already MERGED (terminal)", rdi_id)
+                return Response({"error": "rdi_already_merged"}, status=status.HTTP_409_CONFLICT)
+
+            job = remove_rdi_population_async_task(rdi, callback_url=callback_url, signed_token=signed_token)
+            if job is None:
+                logger.info("RDI reset for %s is already in progress (status %s)", rdi_id, rdi.status)
+                return Response({"id": str(rdi.id), "status": rdi.status}, status=status.HTTP_202_ACCEPTED)
+
+            rdi.status = RegistrationDataImport.DELETE_SCHEDULED
+            rdi.save(update_fields=["status"])
```
MarekBiczysko, line 304: Something to consider:


Suppose one program has two RDIs waiting to merge:

  RDI-A — MERGE_SCHEDULED — oldest
  RDI-B — MERGE_SCHEDULED — next

  The dispatcher runs once, selects the oldest RDI-A, and creates a merge job for it:

  Dispatcher → merge job for RDI-A

  Before that merge job starts, Country Workspace resets RDI-A. Its status changes to:

  RDI-A — DELETE_SCHEDULED
  RDI-B — MERGE_SCHEDULED

  The previously created merge job for RDI-A still runs, but it sees that RDI-A is now being deleted. It exits without merging it and returns False.

  The code only runs the dispatcher again after a successful merge. Because RDI-A was not merged, no new dispatcher task is created. Therefore, nobody comes back to select RDI-B, even though it is ready.

  RDI-B will remain waiting until some unrelated action—such as an admin retry—happens to run the dispatcher again.

  The fix is to trigger the dispatcher after RDI-A is changed to DELETE_SCHEDULED. The dispatcher will then ignore RDI-A and select RDI-B.

src/hope/apps/registration_data/api/country_workspace.py
```diff
@@ -0,0 +1,16 @@
...
+    API_AUTHENTICATION_REQUIRED = False  # no `Authorization` header; auth is the signed_token in the body
+
+    class CountryWorkspaceAPIError(Exception):
+        pass
+
+    API_EXCEPTION_CLASS = CountryWorkspaceAPIError
+
+    def notify_rdi_deleted(self, signed_token: str) -> None:
+        """Success-only ping — POST the signed token to the callback_url (== self.api_url)."""
+        self._post(self.api_url, data={"signed_token": signed_token})
```
johniak: `BaseAPI._post` calls requests without a timeout (`core/api/mixins.py:87-92`). That was fine while every subclass talked to a host from settings, but this URL arrives in the request body. A callback host that accepts the connection and never answers pins a Celery worker indefinitely, and the `Retry(total=3)` adapter multiplies it. Pass an explicit timeout on this call.

src/hope/apps/registration_data/tasks/rdi_removal_async.py
```diff
@@ -0,0 +1,74 @@
...
+      - ProtectedError (dependents) → DELETE_FAILED, no retry.
+      - any other error             → propagates; async_retry_job_task retries and, once
+                                      spent, the on_failure hook marks DELETE_FAILED.
+    """
+
+    def execute(self, rdi_id: str, callback_url: str, signed_token: str) -> None:
+        from hope.apps.registration_data.celery_tasks import locked_cache, notify_rdi_deleted_async_task
+
+        logger.info("RDI reset job started for %s", rdi_id)
+        with locked_cache(key=f"merge_registration_data_import_async_task-{rdi_id}") as acquired:
```
johniak: This is the key held by the legacy UI merge (`celery_tasks.py:316`). CW RDIs — the only ones this endpoint accepts — merge through `fetch_findings_and_merge_rdi_action`, which locks `fetch_findings_and_merge_rdi-program-{program_id}` (`celery_tasks.py:495`). So for the RDIs this job actually handles the cache lock excludes nothing, and both tests that prove the exclusion drive the legacy task. The row lock still keeps this correct, but either add the program key or drop the cache lock and say so.

src/hope/apps/registration_data/services/rdi_removal.py
```diff
@@ -0,0 +1,65 @@
...
+        "Removing RDI %s population: %d households, %d individuals (delete_rdi=%s)",
+        rdi.id,
+        len(household_ids),
+        len(individual_ids),
+        delete_rdi,
+    )
+    Household.all_objects.filter(registration_data_import=rdi).update(head_of_household=None)
+
+    if delete_rdi:
+        rdi.delete()
```
johniak: The cascade fires the per-row `pre_delete` receivers in `apps/household/signals.py:23-37`, each registering an on_commit cache increment per Household and per Individual — measured on this branch: 6 callbacks for 1 household, 26 for 5. With ES on, `remove_individual_from_elasticsearch` adds a delete-by-query plus a `program` lookup per individual, duplicating the bulk ES cleanup done a few lines below. On the big RDIs this endpoint exists for that is tens of thousands of redundant round trips.

src/hope/apps/registration_data/tasks/rdi_removal_async.py
```diff
@@ -0,0 +1,74 @@
...
+            if rdi.status == RegistrationDataImport.MERGED:
+                raise NonRetriableTaskError("rdi_already_merged")
+            remove_rdi_population(rdi, delete_rdi=True, swallow_es_errors=True)
+            return True
+
+    @staticmethod
+    def mark_failed(rdi_id: str, *, reason: str) -> None:
+        # Internal use only, CW is not told about failure.
+        with transaction.atomic():
+            RegistrationDataImport.objects.filter(id=rdi_id).update(
```
johniak, line 77: `.update()` bypasses `post_save`, so the RDI list cache is never invalidated and the UI keeps showing DELETE_SCHEDULED after the row flipped to DELETE_FAILED. `invalidate_rdi_cache` (`registration_data/signals.py:11-24`) exists for exactly this — its docstring says to call it after `RegistrationDataImport.objects.filter(...).update(...)`. It needs the BA slug and program code, so `test_fail_marks_status_keyed_on_id_without_a_select` needs to relax.

src/hope/api/endpoints/rdi/base.py
```diff
@@ -226,3 +242,67 @@ def update(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+class ResetRDIView(BusinessAreaIngestCWOnlyMixin, HOPEAPIBusinessAreaView):
+    """Async reset (hard-delete) of a CW-managed RDI.
+
+    RDI is resettable in every status except ``MERGED`` (terminal -> 409). An in-flight
+    merge holding the row lock is not waited on -> 409 ``rdi_merge_in_progress`` (retryable).
+    """
+
+    permission = Grant.API_RDI_DELETE
+
+    def post(self, request: Request, *args: Any, **kwargs: Any) -> Response:
```
johniak, line 271: Every sibling POST endpoint that takes a body declares its schema (`delegate_people.py:42`, `program.py:54`, `upload.py:301`, `lax.py:468`), but this one doesn't. drf-spectacular has no serializer to work from here, so `POST rdi/{id}/reset/` shows up in Swagger with no request body at all — `callback_url` and `signed_token` are invisible to anyone reading the docs. Add `@extend_schema(request=RDIResetSerializer)`.

src/hope/apps/registration_data/celery_tasks.py
```diff
@@ -530,11 +530,62 @@ def rdi_dispatcher_task_action(job: AsyncRetryJob) -> bool:
...
+
+def remove_rdi_population_on_failure(job: AsyncRetryJob, exc: Exception) -> None:
+    """on_failure_action target — fires only when transient retries are exhausted (REWORK #3)."""
+    from hope.apps.registration_data.tasks.rdi_removal_async import RdiPopulationRemoval
+
+    RdiPopulationRemoval.mark_failed(job.config["registration_data_import_id"], reason=str(exc))
+
+
+def notify_rdi_deleted_async_task(callback_url: str, signed_token: str) -> None:
+    AsyncRetryJob.queue_task(
```
johniak, line 574: This is the only `AsyncRetryJob.queue_task` in the module without `instance=`/`program=`, so the callback job can't be filtered by program in the admin and Sentry has nothing to tag it with. The RDI is gone by then, but the program is known before the delete and could be threaded through. There's also no `on_failure_action`, so once retries are spent nothing records that CW was never told.

tests/unit/apps/registration_data/test_rdi_removal.py
```diff
@@ -0,0 +1,261 @@
...
+def draft_household(
+    business_area: BusinessArea, draft_program: Program, draft_rdi: RegistrationDataImport
+) -> Household:
+    return PendingHouseholdFactory(
+        business_area=business_area, program=draft_program, registration_data_import=draft_rdi
+    )
+
+
+@pytest.fixture
+def program_less_rdi(business_area: BusinessArea) -> RegistrationDataImport:
```
johniak: `program_less_rdi` and `program_less_household` are never requested by any test here, and they can't be: `RegistrationDataImportFactory(program=None)` blows up on `post_save` in `registration_data/signals.py:31` (`instance.program.code` on None). Either drop both, or null-guard the signal and add the test they were written for — `rdi.delete()` on a programless RDI hits the same crash in `pre_delete`.

src/hope/apps/registration_data/celery_tasks.py
```diff
@@ -530,11 +530,62 @@ def rdi_dispatcher_task_action(job: AsyncRetryJob) -> bool:
...
+            "signed_token": signed_token,
+            "on_failure_action": "hope.apps.registration_data.celery_tasks.remove_rdi_population_on_failure",
+        },
+        group_key="registration_data",
+        description=f"Delete RDI population {registration_data_import.id}",
+    )
+
+
+def remove_rdi_population_async_task_action(job: AsyncRetryJob) -> None:
+    """Retriable wipe worker (REWORK #3): wipe the population, then success-callback CW on success only."""
```
johniak: "REWORK #3" here and on line 563 is a review-round note, not something the next reader can act on. Drop the marker and keep the rest of the sentence.

tests/unit/apps/registration_data/api/test_country_workspace.py
```diff
@@ -0,0 +1,46 @@
...
+    assert "Authorization" not in responses.calls[0].request.headers
+
+
+@responses.activate
+def test_notify_reads_no_body() -> None:
+    responses.add(responses.POST, CALLBACK_URL, status=200)
+
+    CountryWorkspaceAPI(api_url=CALLBACK_URL).notify_rdi_deleted(SIGNED_TOKEN)
+
+    assert len(responses.calls) == 1
```
johniak: This asserts nothing about reading a body, and the call it makes is identical to `test_notify_posts_signed_token_in_body`, which already checks the same thing. Same for the no-auth-header test — that assertion could just move into the first test.

johniak (changes requested):

src/hope/apps/registration_data/services/rdi_removal.py
Jeeziorny: wow, this would have kill us. Good catch!

Jeeziorny:
BLOCKED BY https://github.com/unicef/hope/pull/6326

johniak (approved):

PR #6201, opened 2026-07-16
328310 payment admin readonly fields

MarekBiczysko:
[AB#328310](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/328310)


src/hope/admin/payment_plan.py
```diff
@@ -174,6 +174,60 @@ class PaymentPlanAdmin(HOPEModelAdminBase, PaymentPlanCeleryTasksMixin):
         "export_pdf_file_summary",
         "reconciliation_import_file",
     )
+    readonly_fields = (
```
pkujawa, line 177: Would be good to add `plan_type` to readonly

src/hope/admin/follow_up_instruction.py
```diff
@@ -21,15 +23,20 @@ class FollowUpInstructionAdmin(HOPEModelAdminBase):
...
+        "created_by",
+        "status",
     )
 
     @button(permission="payment.view_followupinstruction")
     def payment_plans(self, request: HttpRequest, pk: str) -> HttpResponseRedirect:
         url = reverse("admin:payment_paymentplan_changelist")
         return HttpResponseRedirect(f"{url}?follow_up_instruction__id__exact={pk}")
+
+    def has_add_permission(self: Any, request: Any, obj: Any = None) -> bool:
```
pkujawa: it's a ModelAdmin, no an inline, so `obj` is not passed. (Same for other `has_add_permission` added in this PR)

pkujawa (approved):

src/hope/admin/account.py
```diff
@@ -56,13 +68,29 @@ def get_business_area(self, obj: Account) -> BusinessArea:
     def get_program(self, obj: Account) -> Program:
         return obj.individual.program
 
+    def has_add_permission(self, request: HttpRequest) -> bool:
+        return False
+
```
nerjan, line 73: Idk if thats an issue, but deleting is still possible.

nerjan (approved):

src/hope/admin/account.py
domdinicola: maybe it would be a good idea to have a mixin @MarekBiczysko with add, edit, and delete only for root users

PR #6206, opened 2026-07-16
[DO NOT MERGE] [ONE-TIME-SCRIPT] Fix Follow Up PP Group assignment

pkujawa:
fix_payment_plan_groups — one-time data fix for the migration-0069 group backfill                           
  
  Corrects two gaps:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
  1. Move each follow-up into its source plan's group (matching the flow of newly created ones) and deletes the now-empty group.                                                                                                           
  2. Default Group per cycle. New cycles get a "Default Group" via signal, but pre-existing cycles never did. This creates one for every cycle that's missing it. (empty group since all PPs got their own but matching the structure of new ones)                        

```
  from hope.one_time_scripts.fix_payment_plan_groups import fix_payment_plan_groups                                                                                                                                                                                                                                          
  fix_payment_plan_groups()    
```             

 Do not merge — opened only to run CI and keep a record of this change

### DIFF
--- a/src/hope/one_time_scripts/fix_payment_plan_groups.py
+++ b/src/hope/one_time_scripts/fix_payment_plan_groups.py
@@ -1,0 +1,78 @@
+"""Fix the PaymentPlanGroup backfill from migration 0069.
+
+1) Move follow-up Payment Plans into their source plan's group and drop the group that
+   0069 created for them.
+2) Ensure every Programme Cycle has a Default Group.
+
+Run from Django shell:
+    from hope.one_time_scripts.fix_payment_plan_groups import fix_payment_plan_groups
+    fix_payment_plan_groups()
+"""
+from django.db import transaction
+
+from hope.models import PaymentPlan, PaymentPlanGroup, ProgramCycle
+
+BATCH_SIZE = 2000
+DEFAULT_GROUP_NAME = "Default Group"
+
+
+def fix_payment_plan_groups() -> None:
+    with transaction.atomic():
+        print("Step 1/2: ensuring a Default Group in every cycle...")
+        created_defaults = _ensure_default_group_per_cycle()
+        print(f"  created {created_defaults} Default Group(s)")
+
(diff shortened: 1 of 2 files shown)


MarekBiczysko (approved):

src/hope/one_time_scripts/fix_payment_plan_groups.py
```diff
@@ -0,0 +1,78 @@
...
+
+def _ensure_default_group_per_cycle() -> int:
+    """Create a Default Group for every cycle that lacks one."""
+    cycle_ids = (
+        ProgramCycle.objects.exclude(payment_plan_groups__name=DEFAULT_GROUP_NAME)
+        .distinct()
+        .values_list("id", flat=True)
+    )
+    groups = [PaymentPlanGroup(cycle_id=cid, name=DEFAULT_GROUP_NAME) for cid in cycle_ids]
+    PaymentPlanGroup.objects.bulk_create(groups, batch_size=BATCH_SIZE, ignore_conflicts=True)
```
johniak, line 40: bulk_create doesn't fire post_save, so the create_default_payment_plan_group cache-increment signal that bumps the program's payment_plan_groups_list version never runs for these Default Groups. Step 2's orphaned-group deletes go through QuerySet.delete, which does fire post_delete (the receiver prevents a fast delete) and self-invalidates that program's cache. But a program that only gained a Default Group here and had no orphaned group deleted keeps a stale cached group list until the next group write. Consider bumping the group-list cache version for affected programs after the run (or note it in the run instructions).

johniak (approved):

src/hope/one_time_scripts/fix_payment_plan_groups.py
pkujawa: Yes, we will bump the group-list cache version after run

pkujawa:
The script was run and data was checked.

PR #6214, opened 2026-07-21
Add Known Affected Beneficiaries (KAB) household counters

johniak:
[AB#326718](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/326718)

## What and how

Households get a new set of 30 nullable `kab_*` counters ("known affected beneficiaries"), parallel to the existing composition fields. They answer "how many people do we know are affected" for every data collecting type. If the stored age and gender disaggregation is populated, KAB mirrors it; if it is empty and the DCT declares it registers individual members (new `collects_individual_data` flag, auto enabled where `recalculate_composition=True`, editable in the admin), KAB is counted from linked individuals with the same rules as the existing recalculation; otherwise it stays NULL, so "not collected" never turns into a false zero.

The counting rules live in a single shared filter builder (`_composition_count_filters`), used both by the per-household recalculation and by the backfill's set-based batch aggregation, so the two paths cannot drift apart.

Behavioural changes to the existing recalculation:
- the recalculation tasks no longer skip households of non recalculating DCTs, so their KAB gets refreshed too.

Fixes of pre-existing issues (surfaced during review, present on develop before this PR):
- `other_sex_group_count` and `unknown_sex_group_count` now apply the same active-beneficiary filters as every other counter — previously they also counted non-beneficiaries, withdrawn and duplicates;
- bulk individual updates (XLSX individual update) now schedule the population-fields recalculation for affected households whenever recalc-relevant columns change — previously these writes left the stored composition (and now KAB) stale indefinitely.

## API and UI

The new fields are returned by the household detail endpoint. Since `HouseholdDetailSerializer` is nested in `PaymentDetailSerializer`, they are returned in payment detail as well, next to the composition fields they mirror — purely additively, nothing existing changed.

Nothing consumes them yet. No screen in the frontend renders KAB: the household details page still shows the household composition only, and dashboards, exports and reporting are untouched. KAB is available to API consumers only. Presenting it in the UI is follow up work, and whichever screen picks it up first has to distinguish an empty counter (unknown) from a genuine `0`.

Documentation for all of the above is included in this PR — user manual, Data Collection Type glossary term, an administrative Maintenance page for `backfill_kab` and a developer page on the recalculation pipeline.

## Performance

The backfill is designed for production scale (~20M households): the copy phase is a set-based `UPDATE` per batch, and the compute phase runs one grouped aggregate over individuals plus one `bulk_update` per batch of 5000 households — constant queries per batch, no per-household round trips. Query counts on the backfill and async chunk paths are pinned with `django_assert_num_queries`.

## Action required after deploy

On every environment run `python manage.py backfill_kab` once after the release, otherwise existing households keep NULL KAB. The command works program by program, logs progress and can be safely restarted.

Partial DCTs that do collect individuals need `collects_individual_data` enabled in the admin.

## Decisions to confirm during review

Plain columns were chosen over a JSONField for reporting ergonomics and consistency with existing composition columns. The `kab_` prefix abbreviates "known affected beneficiaries". The fallback is all or nothing: individuals are counted only when the entire disaggregation is NULL. `child_hoh`/`fchild_hoh` are not mirrored since they are flags, not counts.

Open question: where should KAB be consumed next (household details page, dashboard, power query reports)? Please advise so follow up tickets can be created.


Jeeziorny:
Maybe it is good idea: https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/326718#6180400

Jeeziorny (dismissed):

src/hope/apps/household/celery_tasks.py
pkujawa: This loops per household - `recalculate_data` re-fetches and re-counts each one individually.
`aggregate_composition_by_household_id` already does this in one query for the backfill - could wwe reuse it here?

src/hope/apps/household/celery_tasks.py
```diff
@@ -84,13 +84,9 @@ def recalculate_population_fields_async_task_action(job: AsyncJob) -> None:
     params = {}
     if household_ids:
         params["pk__in"] = household_ids
-    recalculate_composition = None
-    if program_id:
-        program = Program.objects.get(id=program_id)
-        recalculate_composition = program.data_collecting_type.recalculate_composition
+    # No gate on recalculate_composition: recalculate_data itself decides what to compute
+    # (composition only when the flag is on, KAB always), so all selected households flow through.
     queryset = Household.objects.filter(**params).only("pk").order_by("pk")
```
pkujawa: If household_ids is empty, this filters on nothing and picks up every household in the DB (program_id isn't applied to the queryset).                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                             
The two new call sites guard against this, but rdi_merge.py doesn't - households_to_merge_ids could be empty when every household in the RDI collides.                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                             
  Could we guard here rather than in each caller?  `if not household_ids: return`

src/hope/models/data_collecting_type.py
```diff
@@ -27,6 +27,11 @@ def get_choices() -> list[tuple[str, str]]:
     individual_filters_available = models.BooleanField(default=False)
     household_filters_available = models.BooleanField(default=True)
     recalculate_composition = models.BooleanField(default=False)
+    collects_individual_data = models.BooleanField(
+        default=False,
+        help_text="Whether this data collecting type registers individual members data; "
+        "when enabled, known affected beneficiaries may be derived by counting linked individuals",
+    )
     weight = models.PositiveSmallIntegerField(default=0)
```
pkujawa: as already raised, help_text for `weight` and `collects_individual_data` missing, while requested in the ticket

src/hope/apps/household/api/serializers/household.py
```diff
@@ -305,6 +305,36 @@ class Meta:
...
+            "kab_female_age_group_12_17_count",
+            "kab_female_age_group_18_59_count",
+            "kab_female_age_group_60_count",
+            "kab_male_age_group_0_5_count",
+            "kab_male_age_group_6_11_count",
+            "kab_male_age_group_12_17_count",
+            "kab_male_age_group_18_59_count",
+            "kab_male_age_group_60_count",
+            "kab_female_age_group_0_5_disabled_count",
+            "kab_female_age_group_6_11_disabled_count",
```
pkujawa, line 319: as per description, "The new fields are exposed on the household detail endpoint only"
but HouseholdDetailSerializer is nested in PaymentDetailSerializer.
Generally would be good to use a smaller serializer in payment, since this one was already huge and growing.

src/hope/apps/household/management/commands/backfill_kab.py
```diff
@@ -0,0 +1,97 @@
...
+                program_copied += Household.objects.filter(pk__in=pks).update(**copy)
+                self.stdout.write(
+                    f"{prefix}: copied batch of {len(pks)} in {time.monotonic() - batch_start:.1f}s,"
+                    f" total copied {program_copied}"
+                )
+            if collects_individual_data:
+                to_compute = households.filter(~COMPOSITION_PRESENT, kab_size__isnull=True)
+                for pks in _pk_batches(to_compute, batch_size):
+                    batch_start = time.monotonic()
+                    # ponytail: no per-row lock/transaction — the backfill is idempotent and any
```
pkujawa: I don't recognize this prefix - leftover?

tests/unit/apps/household/test_recalculate_kab.py
```diff
@@ -0,0 +1,370 @@
...
+    # Gate removal: previously the action emptied the queryset for recalculate_composition=False
+    # DCTs, so nothing ran. Now the household flows to the chunk task and KAB is computed.
+    household = household_from_individuals(recalculate_composition=False, collects_individual_data=True)
+    job = AsyncJob.objects.create(
+        type="JOB_TASK",
+        action="hope.apps.household.celery_tasks.recalculate_population_fields_async_task_action",
+        config={"household_ids": [str(household.pk)], "program_id": str(household.program_id)},
+    )
+
+    from hope.apps.household.celery_tasks import recalculate_population_fields_async_task_action
```
pkujawa: can we move the import to the top?

tests/unit/apps/household/test_recalculate_kab.py
```diff
@@ -0,0 +1,370 @@
...
+    django_assert_num_queries: Any,
+) -> None:
+    household = household_from_individuals(recalculate_composition=False, collects_individual_data=True)
+    job = AsyncJob.objects.create(
+        type="JOB_TASK",
+        action="hope.apps.household.celery_tasks.recalculate_population_fields_chunk_async_task_action",
+        config={"households_ids": [str(household.pk)], "program_id": str(household.program_id)},
+    )
+
+    from hope.apps.household.celery_tasks import recalculate_population_fields_chunk_async_task_action
```
pkujawa: move import to the top

src/hope/apps/household/celery_tasks.py
pkujawa, line 138: This list can contain "None" — individuals without a household (representatives) become the string "None" and blow up the pk__in filter. Fails on develop too, so pre-existing, but worth fixing alongside the guard:                                                                                                                                                                                                                                                                        
`.exclude(household_id__isnull=True)`                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                             
Actually, after this PR, the interval job will actually work for the first time it seems (it currently either crashes or no-ops), so it's worth a look at the per-household query cost.

(It did not pass the program so it went into the condition you removed 
`    if not recalculate_composition:
        queryset = queryset.none()
`
and didn't trigger before.
While, if any individual had no household, it crashed earlier at  
`queryset = Household.objects.filter(**params).only("pk").order_by("pk")`             
)

pkujawa (changes requested):

pkujawa:
```
Action required after deploy

On every environment run python manage.py backfill_kab once after the release, otherwise existing households keep NULL KAB. The command works program by program, logs progress and can be safely restarted.

Partial DCTs that do collect individuals need collects_individual_data enabled in the admin.
```
So we also need to rerun the backfill_kab after enabling the flag on DCTs manually because these will be skipped.

src/hope/apps/household/api/serializers/household.py
johniak: Good catch on the description — KAB also surfaces in payment detail through the nested HouseholdDetailSerializer; I've corrected the wording. That nesting already exposes the full composition fields (female_age_group_* etc.), so KAB landing next to its mirrors there is consistent — and purely additive, nothing existing changed.

Jeeziorny (dismissed):

PR #6222, opened 2026-07-21
[329150] Email notification for new step: ready for closure + send back

pkujawa:
[AB#329150](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/329150)

Payment plans now send a notification email when they are marked ready for                                                                                                                                                                                                                                                 
  closure and when they are sent back to finished. In both cases the recipients                                                                                                                                                                                                                                              
  are the users holding the permission required for the next action in the flow - `PM_CLOSE_FINISHED`                                                                                                                                                                                                                                            
  are notified after mark-ready ("Payment pending for Closure"); users with                                                                                                                                                                                                                                                     
  `PM_MARK_READY_FOR_CLOSURE` are notified on send-back ("Payment sent back to                                                                                                                                                                                                                                               
  Finished").                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                             
  ### Changes                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                             
  - Added `MARK_READY_FOR_CLOSURE` and `SEND_BACK_TO_FINISHED` to                                                                                                                                                                                                                                                            
    `PaymentPlan.Action`, with recipient permissions, subjects and email bodies in                                                                                                                                                                                                                                           
    `PaymentNotification`.                                                                                                                                                                                                                                                   
  - `ready_for_closure(user, *, notify=True)` and `send_back_to_finished(user)`                                                                                                                                                                                                                                              
    queue the notification email.                                                                                                                                                                                                                                                                                            
  - Follow-up-instruction bulk close passes `notify=False`: it auto-advances                                                                                                                                                                                                                                                 
    `FINISHED → READY_FOR_CLOSURE → CLOSED` without pause, so the intermediate                                                                                                                                                                                                                                               
    email would reach every closer once per child plan.

### DIFF
--- a/src/hope/apps/payment/notifications.py
+++ b/src/hope/apps/payment/notifications.py
@@ -18,12 +18,16 @@ class PaymentNotification:
     ACTION_APPROVE = PaymentPlan.Action.APPROVE.name
     ACTION_AUTHORIZE = PaymentPlan.Action.AUTHORIZE.name
     ACTION_REVIEW = PaymentPlan.Action.REVIEW.name  # payment plan release
+    ACTION_MARK_READY_FOR_CLOSURE = PaymentPlan.Action.MARK_READY_FOR_CLOSURE.name
+    ACTION_SEND_BACK_TO_FINISHED = PaymentPlan.Action.SEND_BACK_TO_FINISHED.name
 
     ACTION_TO_RECIPIENTS_PERMISSIONS_MAP = {
         ACTION_SEND_FOR_APPROVAL: Permissions.PM_ACCEPTANCE_PROCESS_APPROVE.name,
         ACTION_APPROVE: Permissions.PM_ACCEPTANCE_PROCESS_AUTHORIZE.name,
         ACTION_AUTHORIZE: Permissions.PM_ACCEPTANCE_PROCESS_FINANCIAL_REVIEW.name,
         ACTION_REVIEW: Permissions.PM_DOWNLOAD_XLSX_FOR_FSP.name,
+        ACTION_MARK_READY_FOR_CLOSURE: Permissions.PM_CLOSE_FINISHED.name,
+        ACTION_SEND_BACK_TO_FINISHED: Permissions.PM_MARK_READY_FOR_CLOSURE.name,
     }
 
     ACTION_PREPARE_EMAIL_BODIES_MAP = {
@@ -45,6 +49,16 @@ class PaymentNotification:
         ACTION_REVIEW: {
             "action_name": "released",
             "subject": "Payment is Released",
+            "recipient_title": "Reviewer",
+        },
+        ACTION_MARK_READY_FOR_CLOSURE: {
+            "action_name": "marked as ready for closure",
+            "subject": "Payment pending for Closure",
+            "recipient_title": "Reviewer",
+        },
+        ACTION_SEND_BACK_TO_FINISHED: {
+            "action_name": "sent back to finished",
+            "subject": "Payment sent back to Finished",
             "recipient_title": "Reviewer",
         },
     }
--- a/src/hope/apps/payment/services/payment_plan_services.py
+++ b/src/hope/apps/payment/services/payment_plan_services.py
@@ -1313,7 +1313,7 @@ class PaymentPlanService:
                     ind_filter.individuals_filters_block = ind_filter_block_copy
                     ind_filter.save()
 
-    def ready_for_closure(self) -> PaymentPlan:
+    def ready_for_closure(self, user: "User", *, notify: bool = True) -> PaymentPlan:
         if self.payment_plan.status != PaymentPlan.Status.FINISHED:
             raise ValidationError(
                 f"Mark as Ready for Closure is possible only within Status {PaymentPlan.Status.FINISHED}"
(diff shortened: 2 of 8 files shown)


nerjan (approved):

MarekBiczysko (approved):

src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -1322,15 +1322,28 @@ def ready_for_closure(self) -> PaymentPlan:
         flow.status_ready_for_closure()
         self.payment_plan.save(update_fields=("status", "status_date", "updated_at"))
         self.payment_plan.refresh_from_db(fields=["status", "status_date", "updated_at"])
+        if notify:
+            send_payment_notification_emails_async_task(
```
johniak, line 1326: These notification jobs are created after a status check on an unlocked `PaymentPlan`. Two concurrent requests can both observe `FINISHED`, both transition it, and both enqueue the same email; `send_back_to_finished()` has the same race. Lock the row before checking the status, or use a conditional update whose affected-row count gates the enqueue, so only one transition can produce a notification.

src/hope/apps/payment/notifications.py
```diff
@@ -18,12 +18,16 @@ class PaymentNotification:
...
+    ACTION_MARK_READY_FOR_CLOSURE = PaymentPlan.Action.MARK_READY_FOR_CLOSURE.name
+    ACTION_SEND_BACK_TO_FINISHED = PaymentPlan.Action.SEND_BACK_TO_FINISHED.name
 
     ACTION_TO_RECIPIENTS_PERMISSIONS_MAP = {
         ACTION_SEND_FOR_APPROVAL: Permissions.PM_ACCEPTANCE_PROCESS_APPROVE.name,
         ACTION_APPROVE: Permissions.PM_ACCEPTANCE_PROCESS_AUTHORIZE.name,
         ACTION_AUTHORIZE: Permissions.PM_ACCEPTANCE_PROCESS_FINANCIAL_REVIEW.name,
         ACTION_REVIEW: Permissions.PM_DOWNLOAD_XLSX_FOR_FSP.name,
+        ACTION_MARK_READY_FOR_CLOSURE: Permissions.PM_CLOSE_FINISHED.name,
+        ACTION_SEND_BACK_TO_FINISHED: Permissions.PM_MARK_READY_FOR_CLOSURE.name,
```
johniak, line 30: `send_back_to_finished` is performed with `PM_MARK_READY_FOR_CLOSURE`, and this mapping targets that same permission. Because `_prepare_user_recipients()` removes the action user, a business area with a single holder produces an empty `To` list; `MailjetClient` then iterates zero batches, so no email, including the intended CC to the actor, is sent. Keep the actor in `To` for this action or provide another non-empty fallback, and cover the single-holder case.

src/hope/models/payment_plan.py
```diff
@@ -274,6 +274,8 @@ class Action(models.TextChoices):
         REVIEW = "REVIEW", "Review"
         REJECT = "REJECT", "Reject"
         FINISH = "FINISH", "Finish"
+        MARK_READY_FOR_CLOSURE = "MARK_READY_FOR_CLOSURE", "Mark Ready for Closure"
```
johniak, line 277: `PaymentPlan.Action.choices` feeds `PaymentPlanBulkActionSerializer`, so these additions make the managerial bulk endpoint advertise and accept both closure actions. `_get_action_permission()` and `PaymentPlanService.actions_map` support neither, so requests pass serializer validation and then fail instead of executing. Keep notification-only actions out of this shared enum, or narrow the bulk serializer to its three supported actions.

src/hope/apps/payment/notifications.py
```diff
@@ -47,6 +51,16 @@ class PaymentNotification:
             "subject": "Payment is Released",
             "recipient_title": "Reviewer",
         },
+        ACTION_MARK_READY_FOR_CLOSURE: {
+            "action_name": "marked as ready for closure",
+            "subject": "Payment pending for Closure",
+            "recipient_title": "Reviewer",
```
johniak, line 57: This email goes to `PM_CLOSE_FINISHED` holders—the Deputy Representative/closer described in the ticket—but the template variables address them as `Payment Plan Reviewer`. Use the closure role label here so the new message does not misidentify its recipient.

tests/unit/apps/payment/test_payment_notification.py
```diff
@@ -605,10 +631,158 @@ def test_prepare_user_recipients_for_release(notification_setup: dict) -> None:
...
+    payment_notification = PaymentNotification(
+        notification_setup["payment_plan"],
+        PaymentPlan.Action.MARK_READY_FOR_CLOSURE.name,
+        notification_setup["user_action_user"],
+        f"{timezone.now():%-d %B %Y}",
+    )
+
+    users = notification_setup["users"]
+    assert payment_notification.user_recipients.count() == 11
+
```
johniak, line 650: The project's test rules prohibit `for` loops in test bodies. Parameterize the included and excluded recipient cases instead; the send-back test below has the same issue. That preserves one scenario per test and makes failures identify the exact recipient case.

johniak (changes requested):
None of the new database tests asserts its query count, although the repository requires query-count checks. Add `django_assert_num_queries` around the service and recipient operations so changes to recipient resolution or transition side effects cannot introduce N+1 regressions unnoticed.

src/hope/apps/payment/services/payment_plan_services.py
pkujawa: It's pre-existing for all the acceptance process steps.
I will apply the lock in the separate PR

src/hope/apps/payment/notifications.py
pkujawa: If there is no user with permission to perform the next step in the flow, other than the action user, we do not want to send any emails. That's the goal for all the actions. Producing an email to action user just so they get a CC of this email is not something we want.

src/hope/models/payment_plan.py
pkujawa: The bulk endpoint already advertises all 18 Action values while _get_action_permission() covers 3 (APPROVE, AUTHORIZE, REVIEW). FINISH and SEND_XLSX_PASSWORD sit in exactly the position you describe today - in the enum, in neither actions_map nor the permission map - and 13 further values are in actions_map but 
still get None from the permission map, so they 403 for any non-superuser. Unsupported values are rejected cleanly (403, or 400 from validate_action()) before anything executes, inside transaction.atomic, so nothing partially applies. These two actions have to stay in the enum because PaymentNotification keys   
off PaymentPlan.Action.*.name, same as the existing acceptance-process notifications. 

Narrowing the bulk serializer to its three genuinely supported actions is something I can apply in a separate PR, since it changes the advertised contract for 15 pre-existing values.

src/hope/apps/payment/notifications.py
pkujawa: I did not like the title Dear `Payment Plan Closer`, and even less `Payment Plan Person Marking Ready For Closure`.
`Reviewer` is already the generic salutation rather than a role identifier — we use it both for AUTHORIZE recipients (PM_ACCEPTANCE_PROCESS_FINANCIAL_REVIEW) and for REVIEW recipients (PM_DOWNLOAD_XLSX_FOR_FSP), so it doesn't map to a single permission today.

It's a preference issue, but I feel like `Reviewer` is completely fine as a title, rather than making these weird descriptive titles sometimes.

PR #6232, opened 2026-07-22
[Grievance-revamp] sensitive grievance notifications

nerjan:
[AB#175674](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/175674)

**TL;DR**

- **Recipients come from permissions, not hardcoded role names.**
- **One daily digest replaces the per-event assignment and "ticket updated" emails** — 06:00 UTC, one email per user per business area, covering the previous day.
- **Three new columns on `GrievanceTicket`** record when a ticket was assigned, by whom, and who last edited it. One migration, index built concurrently.
- **Nobody is emailed about their own action**, and deactivated users or users without an address are never mailed at all.
- **Sensitive tickets never carry a link**, in the digest or anywhere else.

**Summary**
Revamps grievance & feedback email notifications so recipients are resolved by permission, not hardcoded role names, and closes several gaps where notifications were silently sent to nobody or to the wrong people. Replaces the per-event assignment and "ticket updated" emails with one daily digest per user per business area, and ensures the person performing an action never gets emailed about their own action.

**Changes**

- **Recipients are permission-based, not role-name based.** The old code matched hardcoded role names ("Approver", "Adjudicator", "Releaser", "Senior Management"), which are fragile and don't reflect who can actually act. Recipients are now resolved from the relevant grievance permissions, scoped to the ticket's program(s), and filtered to mailable users (active + has email; superusers/staff excluded in prod). Partner-derived role assignments are included, matching the payment-notification pattern.

- **AS_OWNER / AS_CREATOR handled on approval notifications.** `_prepare_for_approval_recipients` now notifies permission holders (previously role holders) plus the ticket creator (if they hold the `_AS_CREATOR` permission) and the assignee (if they hold the `_AS_OWNER` permission), per ticket category.

- Sensitive & complaint tickets have no "Approve" step - it directly goes to "Close" action, but notification was sent anyway to "Approver" role. Now it is send to user with GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK (+ creator/owner variants) permissions (mirroring view permissions)

- When Payment Verifivation Grievance Ticket is created notifies only users who have both permissions: one of grievance-view permission and one of payment-verification-view permission

- **Replace the assignment and "ticket updated" emails with one daily digest.** At constant hour there is queued one job per business area that has email notifications enabled. Each job mails every user who had a ticket assigned to them, or a ticket they created or own edited, on the previous UTC day — one email per user per business area instead of one per event. `ACTION_ASSIGNMENT_CHANGED`, `ACTION_TICKET_UPDATED` and the bulk-assign notification task are removed along with their templates.

- **Record when a ticket was assigned, by whom, and who last edited it.** Nothing stored this, so the digest had nothing to read. `assigned_at` / `assigned_by` are written wherever the assignee actually changes — update endpoint, bulk assign, creation with an assignee, self-assign on status change, and the beneficiary ticket API. `user_modified_by` is written alongside the existing `user_modified`. The columns record what happened and nothing else; who gets told is decided in the digest query.

- **Add `user_modified_by`.** It tracks that the ticket was modified and by whom. Ticket creation leaves it empty.

- **Never list a change the recipient made themselves.** Self-assignments and self-edits are dropped in the digest query.
- 
- **Stamp only the tickets whose assignee actually changed on bulk assign.** `bulk_assign` updates every selected non-closed ticket, including ones already assigned to that user; those must not turn up in the next digest.

- **Cap the digest at 50 rows per section.** A thousand-ticket bulk assign would otherwise produce a thousand-row Mailjet payload. The remainder is summarised as "and N more". 

- **Isolate each recipient's send, and record a run only once every recipient was reached.** A finished run writes a `completed_for` key into the job config, so a redelivery or a recovery re-queue does not mail the same day again; a run that failed for anyone leaves the day unmarked and retriable.

- Sensitive tickets never expose URLs in emails. Sensitive grievances shouldn't leak a linkable ticket URL; templates fall back to the ticket id when no URL is present. Digest rows follow the same rule.

- The user performing the action is excluded from their own notifications. You shouldn't get emailed about an action you just performed (self-assign, sending for approval, editing).

- Emails sent with transaction on commit. Notification sends are wrapped in transaction.on_commit so emails only go out if the transaction succeeds.

- Drop the dead status-change assignment notification. The STATUS_ASSIGNED branch in status_change built an ASSIGNMENT_CHANGED notification at a point where assigned_to is always unset, so it never had any recipients and never sent. Removed it (and the now-unused notifications argument on _validate_status_change_preconditions).

- Skip unmailable assignees. A deactivated user or one with a blank email are now excluded.

- Send notification emails on commit. create, status_change, and create_note sent their emails inline, so a failure later in the same request rolled back the ticket after the mail had already gone out. These now go through transaction.on_commit.


src/hope/apps/grievance/notifications.py
```diff
@@ -73,47 +103,153 @@ def send_email_notification(self) -> None:
...
+                Permissions.GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK_AS_OWNER,
+            ),
+            # Referral is a feedback-type ticket, so closing/approving is gated by the feedback permission.
+            GrievanceTicket.CATEGORY_REFERRAL: (
+                Permissions.GRIEVANCES_CLOSE_TICKET_FEEDBACK,
+                Permissions.GRIEVANCES_CLOSE_TICKET_FEEDBACK_AS_CREATOR,
+                Permissions.GRIEVANCES_CLOSE_TICKET_FEEDBACK_AS_OWNER,
+            ),
+        }
+        permissions = approve_permissions_by_category.get(self.grievance_ticket.category)
```
johniak: `CATEGORY_BENEFICIARY` is missing from this map, so beneficiary tickets fall into the `None` branch and nobody gets the approval email — previously every Approver did. `BENEFICIARY_STATUS_FLOW` (models.py) does allow `IN_PROGRESS`/`ON_HOLD` → `FOR_APPROVAL`, so this is reachable. Closing those tickets is gated by `GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK`, so they need the same triple as complaint/sensitive. A fallback keyed on `ticket.is_feedback` would stop the next new category hitting this too.

src/hope/apps/grievance/api/views.py
```diff
@@ -742,11 +742,25 @@ def partial_update(self, request: Request, *args: Any, **kwargs: Any) -> Respons
             old_object=old_grievance_ticket,
             new_object=grievance_ticket,
         )
+        # The new assignee already receives a dedicated assignment email.
+        assignment_changed = old_grievance_ticket.assigned_to_id != grievance_ticket.assigned_to_id
```
johniak: The ticket-updated mail goes out on every successful PATCH with no check that anything actually changed. Since `_apply_ticket_field_updates` only writes fields whose current value is falsy, a lot of "Save" clicks are no-ops that still email the creator and the assignee. You already have `old_grievance_ticket` and `grievance_ticket` here for `log_create` — gate the notification on a real diff of the logged fields.

src/hope/apps/grievance/celery_tasks.py
```diff
@@ -66,6 +66,35 @@ def deduplicate_and_check_against_sanctions_list_task_single_individual_async_ta
     )
 
 
+def bulk_assign_notifications_async_task_action(job: AsyncJob) -> None:
+    ticket_ids = job.config["ticket_ids"]
+    action_user_id = job.config.get("action_user_id")
+    action_user = User.objects.filter(id=action_user_id).first() if action_user_id else None
+    GrievanceNotification.send_all_notifications(
+        [
+            GrievanceNotification(ticket, GrievanceNotification.ACTION_ASSIGNMENT_CHANGED, editor=action_user)
```
johniak: Worth thinking through before this ships: one notification per ticket means bulk-assigning 50 tickets lands 50 separate emails on the same person, plus 50 renders and 50 Mailjet calls. A single "N tickets assigned to you" mail would solve that, though it does lose the per-ticket subject people may be filtering on. Is per-ticket the deliberate choice here?

src/hope/apps/grievance/notifications.py
```diff
@@ -73,47 +103,153 @@ def send_email_notification(self) -> None:
...
+    def _editor(self) -> "User | None":
+        # The user who performed the action; never notified about their own action.
+        return self.extra_data.get("editor")
+
+    @cached_property
+    def _program_scope(self) -> Q:
+        program_ids = list(self.grievance_ticket.programs.values_list("pk", flat=True))
+        return Q() if not program_ids else Q(program__isnull=True) | Q(program__in=program_ids)
+
+    def _users_with_permissions(self, permissions: list[Permissions]) -> "QuerySet[User]":
```
johniak, line 172: This is now the third copy of "users with permission X in this BA/program, plus partner-derived, minus staff on prod" — `payment/notifications.py` and `periodic_data_update/notifications.py` have the other two, and they already drift (`permissions__contains` there vs `permissions__overlap` here). A shared `users_with_permissions(business_area, programs, permissions)` helper would keep them in sync.

src/hope/apps/grievance/notifications.py
```diff
@@ -73,47 +103,153 @@ def send_email_notification(self) -> None:
...
-        user_roles = RoleAssignment.objects.filter(
-            role__name="Approver",
-            business_area=self.grievance_ticket.business_area,
-        ).exclude(expiry_date__lt=timezone.now())
-        queryset = User.objects.filter(role_assignments__in=user_roles).distinct()
-        if self.grievance_ticket.assigned_to:
-            queryset = queryset.exclude(id=self.grievance_ticket.assigned_to.id)
-        return queryset.all()
+        # Approval permission depends on the ticket type
+        approve_permissions_by_category = {
```
johniak: This dict is a constant rebuilt on every call. Move it up next to `ACTION_VIEW_PERMISSIONS` as a class attribute — it also makes it obvious at a glance which categories are covered and which aren't.

src/hope/apps/grievance/notifications.py
```diff
@@ -213,8 +362,13 @@ def prepare_notification_for_ticket_creation(
     ) -> list["GrievanceNotification"]:
         notifications = []
         if grievance_ticket.assigned_to:
+            # Skip the assignment email when the creator assigns the ticket to themselves.
```
johniak: The comment sits above an `append` that always appends — the skipping actually happens later, through the `editor` exclusion in `_prepare_assigned_to_recipient`. Reword to something like "the editor exclusion drops the recipient when the creator self-assigns".

tests/unit/apps/grievance/test_bulk_action_service.py
```diff
@@ -126,6 +135,150 @@ def test_bulk_update_assignee(grievance_context: dict[str, Any]) -> None:
...
+    assert all(notification.user_recipients == [user_two] for notification in sent_notifications)
+
+
+def test_bulk_assign_notifications_task_action_does_not_scale_queries_per_ticket(
+    grievance_context: dict[str, Any],
+) -> None:
+    user = grievance_context["users"]["user"]
+    user_two = grievance_context["users"]["user_two"]
+    ticket1, ticket2, ticket3, _ = grievance_context["grievance_tickets"]
+    for ticket in (ticket1, ticket2, ticket3):
```
johniak: `pytest.mark.parametrize` fits this file better than loops and inline setup. The two task-action tests just above differ only in whether `action_user_id` is set, so they collapse into a single parametrized test, and this three-ticket setup reads better as a fixture that returns tickets already assigned to `user_two`.

johniak (changes requested):

src/hope/apps/grievance/celery_tasks.py
nerjan: pending question to business

src/hope/apps/grievance/api/views.py
nerjan: I decided that whenever someone clicks save then this person is responsible for that. Tracking if anything actually changed can be tricky, especially that log_create is not catching everything - I assume its better to send more emails that miss some. The fields are priority, urgency for example and all changes to the support_documents.

src/hope/apps/grievance/celery_tasks.py
nerjan: refactored into one daily digest email that gather all edits and assignments made this date. All other assignment/edit notifications are dropped.

src/hope/apps/grievance/notifications.py
```diff
@@ -73,47 +139,80 @@ def send_email_notification(self) -> None:
...
     def _prepare_for_approval_recipients(self) -> "QuerySet[User]":
-        user_roles = RoleAssignment.objects.filter(
-            role__name="Approver",
-            business_area=self.grievance_ticket.business_area,
-        ).exclude(expiry_date__lt=timezone.now())
-        queryset = User.objects.filter(role_assignments__in=user_roles).distinct()
-        if self.grievance_ticket.assigned_to:
-            queryset = queryset.exclude(id=self.grievance_ticket.assigned_to.id)
-        return queryset.all()
+        permissions = GrievanceNotification.CATEGORY_APPROVE_PERMISSIONS.get(self.grievance_ticket.category)
```
johniak, line 193: `CATEGORY_NEGATIVE_FEEDBACK` and `CATEGORY_POSITIVE_FEEDBACK` aren't in the map, so this returns `none()` and nobody is told those tickets need approval — every Approver used to be. Both reach `FOR_APPROVAL` through `FEEDBACK_STATUS_FLOW`, and `CATEGORY_REFERRAL` (same `is_feedback` family) is in the map, so it looks like an oversight. `test_for_approval_recipients_empty_for_category_without_approval` currently locks the empty result in. An `is_feedback` fallback would also cover the next new category.

johniak (changes requested):

src/hope/apps/grievance/notifications.py
nerjan: but there were never notifications for feedbacks because this flow excluded feedbacks 
```
if grievance_ticket.category in (
            GrievanceTicket.CATEGORY_NEGATIVE_FEEDBACK,
            GrievanceTicket.CATEGORY_POSITIVE_FEEDBACK,
        ):
            raise ValidationError("Feedback tickets are not allowed to be created through this mutation.")
```

src/hope/apps/grievance/templates/overdue_notification_email.txt
```diff
@@ -1,5 +1,5 @@
 {% extends 'base_notification_email.txt' %}
 
 {% block content %}
-    A Sensitive Grievance ticket {{ ticket_url }} created over {{ days_ago }} days ago requires action and attention
+    A Sensitive Grievance ticket {% if ticket_url %}{{ ticket_url }}{% else %}{{ ticket_id }}{% endif %} created over {{ days_ago }} days ago requires action and attention
```
pkujawa: It is pre-existing, but the tickets actually are not sensitive here. 
`.exclude(category=GrievanceTicket.CATEGORY_SENSITIVE_GRIEVANCE)` (`celery_tasks.py:139`)
(Sensitive version is in `sensitive_reminder_notification_email.html` and the text here was incorrectly the same)

src/hope/apps/utils/recipients.py
```diff
@@ -0,0 +1,62 @@
...
+        .distinct()
+    )
+    users = (
+        User.objects.filter(
+            Q(role_assignments__in=role_assignments) | Q(partner__role_assignments__in=role_assignments)
+        )
+        .filter(is_active=True)
+        .exclude(email="")
+    )
+    if settings.ENV == "prod":
```
pkujawa: since you moved it and in the meantime it was changed in https://github.com/unicef/hope/pull/6291, it will not show as conflict, but needs the same change:
```suggestion
    if not config.NOTIFY_INTERNAL_USERS::
```

src/hope/apps/utils/recipients.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+        .exclude(expiry_date__lt=timezone.now())
+        .distinct()
+    )
+    users = (
+        User.objects.filter(
+            Q(role_assignments__in=role_assignments) | Q(partner__role_assignments__in=role_assignments)
+        )
+        .filter(is_active=True)
+        .exclude(email="")
```
pkujawa, line 54: 👍 good addition

src/hope/apps/grievance/services/daily_digest_service.py
```diff
@@ -0,0 +1,170 @@
...
+@dataclass
+class RecipientDigest:
+    user: "User"
+    assigned: list[GrievanceTicket] = field(default_factory=list)
+    edited: list[GrievanceTicket] = field(default_factory=list)
+    assigned_total: int = 0
+    edited_total: int = 0
+
+
+class DailyDigestService:
```
pkujawa, line 38: Just a note:

I think the ticket creation for `system` category could also be handled daily, not instant.
For example after RDI merge, there might be multiple NA tickets created and user would get only 1 email daily instead of 1 per NA created (per every NA created, users with permissions are checked now, quite heavy)
(Other instant ones are: creating sensitive tickets, adding notes, sending for approval and sending back to in progress - which makes sense as they are produced not in bulk)

Also daily `sensitive reminders` and monthly `overdue` - they could be put together in 1 email, instead of 1 per ticket.

Out of scope for this ticket, this is all pre-existing. But worth discussing and probably applying in the future.

johniak (approved):

PR #6233, opened 2026-07-23
Rename "Duplicate in batch" to "Duplicate in RDI" in RDI tab view

Jeeziorny:
On 22 Jul demo for new deduplication flow @ntrncic point out that `duplicate in batch` should be renamed to `duplicate in RDI`. See screenshot below:
image

This PR addresses this comment.


MarekBiczysko (dismissed):

Jeeziorny:
After:
image

domdinicola:
@Jeeziorny should this go in the long term or is it independent?

Jeeziorny:
While it was caught on demo related to new deduplication flow, the reported enhancement is not related to the deduplication flow adjustments and can be applied directly on develop branch. That way it won't extend the new dedup PR and can be released faster in the same time (win-win). That's why I opened PR against dev @domdinicola

PR #6235, opened 2026-07-23
[LONG-TERM]-[305854]-Needs adjudication management

mmaciekk:
[AB#305854](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/305854)

## What changed

Adds a dedicated **NA Tickets Management** page for triaging Needs Adjudication grievance tickets side by side, instead of working them one at a time in the regular grievance details view.

  - New page on the na-tickets-management route, opened from a new NA Tickets Management button in the Grievances table toolbar.                                                                                                                                                                                             
  - Filterable, paginated list of Needs Adjudication tickets on the left, comparison panel on the right. Filters are collapsed by default with a show/hide toggle                        
  - The panel compares the golden record against a possible duplicate side by side — name, date of birth, sex, documents, address, photo, account, similarity score — highlighting fields that differ.                                                                                                                       
  - A ticket with several possible duplicates is stepped through one pair at a time, with Previous/Next and a "1 of 3 decided" counter.                                                                                                                                                                                      
  - For each pair the operator can Withdraw either individual, mark them Not Duplicates, or Clear the mark.                                                                                                                                                                                                                  
  - Marks stay in the browser until Finalize. The list shows which tickets are managed, which still have a duplicate to decide, and which need a role handed over.                                                                                                                                                           
  - Finalize sends every decision in one request. Each ticket is resolved and closed in one step — there is no separate approval.                                                                                                                                                                                            
  - If withdrawing someone would leave their household without a head or a collector, the page asks who takes the role over first. Finalize stays disabled until every ticket is fully decided and every handover is chosen.                                                                                                 
  - Changing page or filters with decisions pending asks for confirmation first; the decisions are kept either way.                                                                                                                                                                                                          
  - Tickets someone else closed in the meantime are skipped and named in the result. The rest still go through.                                                                                                                                                                                                              
  - The list shows only open tickets from the selected programme.                                                                                                                                                                                                                                                            
  - The page needs both the "approve flag and dedupe" and "close ticket" permissions, re-checked per ticket on the server.                                                                                                                                                                                                   
  - Demo data (demo_data/grievance.py): adds six Needs Adjudication scenarios covering every role-handover case — head only, primary only, both on one person, alternate only, head of a one-person household, and primary in two households — each tagged in its description with what the screen should show. Individuals  
  get a generated placeholder photo and a document so the comparison rows have content.                        
                                                                                                                                                                                                                                                                                                                             
  API                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                             
  - New endpoint: POST .../grievance-tickets/bulk-needs-adjudication/, returning {resolved, skipped_closed} — ids only, not full ticket payloads.                                                                                                                                                                            
  - Ticket detail now includes, for the compared people: similarity score, accounts, roles in households, and programme id.


  Fixes along the way                                                                                                                                                                                                                                                                                                        
              
  - Role reassignment was broken on the existing single-ticket NA screen. Withdrawing primary collector or head, failed at close time with an error the operator had no way to resolve.                                                                                                                                                                                                                               
  - The bulk endpoint now applies the same per-individual checks as the single-ticket flow: partner area access, and no marking of already-withdrawn people.                                                                                                                                                                                                                                                                                                                                                     
  - Reassign-role picker, existing single-ticket flow: it had no programme filter at all — under All Programmes it listed everyone in the business area, and picking someone from another programme failed on save with no explanation. It now offers only people from the programme of the individual losing the role.                                                                                                                                                                                                                           
  - Underlying cause of both: the shared individual lookup sent a programId parameter that neither endpoint accepts, so the programme filter did nothing whenever no programme was in the URL. It now sends program, and the ticket detail payload carries each individual's programme id so the picker can use it.


src/hope/apps/grievance/api/serializers/ticket_detail.py
```diff
@@ -194,30 +201,154 @@ class DeduplicationEngineSimilarityPairSerializer(serializers.Serializer):
...
+            "roles_in_households",
+        )
+
+    def get_role(self, obj: Individual) -> str | None:
+        role = obj.households_and_roles(manager="all_objects").filter(household=obj.household).first()
+        return role.role if role else None
+
+    @extend_schema_field(IndividualRoleInHouseholdSerializer(many=True))
+    def get_roles_in_households(self, obj: Individual) -> ReturnDict:
+        return IndividualRoleInHouseholdSerializer(obj.households_and_roles.all(), many=True).data
```
nerjan: This can be field next to roles_in_households 
`roles_in_households = IndividualRoleInHouseholdSerializer(source="households_and_roles", many=True)`

nerjan (approved):

src/hope/apps/grievance/api/serializers/ticket_detail.py
```diff
@@ -194,30 +201,147 @@ class DeduplicationEngineSimilarityPairSerializer(serializers.Serializer):
     status_code = serializers.CharField()
 
 
+def can_view_biometric_results(context: Mapping[str, Any]) -> bool:
```
johniak, line 204: This now always checks the permission against the business area. The old code checked it against the programme whenever `program_code` was in the URL, and those two are not the same. A business-area check counts the user's roles from every programme in that business area, while a programme check only counts roles for that one programme. So a user who is allowed to see biometric results in programme A can now see them on a ticket from programme B, including the new `similarity_score` field. Please bring back the programme branch and fall back to the business area only when there is no `program_code`.

src/hope/apps/grievance/services/bulk_action_service.py
```diff
@@ -149,6 +153,102 @@ def bulk_close(
...
+        resolutions_by_id = {str(resolution["ticket_id"]): resolution for resolution in resolutions}
+        ticket_ids = list(resolutions_by_id.keys())
+        tickets = list(
+            GrievanceTicket.objects.filter(
+                ~Q(status=GrievanceTicket.STATUS_CLOSED),
+                category=GrievanceTicket.CATEGORY_NEEDS_ADJUDICATION,
+                business_area__slug=business_area_slug,
+                id__in=ticket_ids,
+            )
+            .select_for_update(of=("self",))
```
johniak, line 181: No deterministic ordering before `select_for_update`, so two operators finalizing overlapping ticket sets can take the row locks in opposite orders and deadlock. That's much more likely here than in `bulk_close` above, because two people triaging the same NA queue naturally pick the same tickets. The repo convention is pk-ordered locking (see `household/services/household_recalculate_data.py`) — add `.order_by("pk")` before `select_for_update`.

src/hope/apps/grievance/services/bulk_action_service.py
```diff
@@ -149,6 +153,102 @@ def bulk_close(
...
+            )
+            if len(skipped_closed) != len(missing_ids):
+                raise ValidationError("Some selected tickets do not exist or are not Needs Adjudication tickets.")
+
+        required = [
+            Permissions.GRIEVANCES_APPROVE_FLAG_AND_DEDUPE.value,
+            Permissions.GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK.value,
+        ]
+        for ticket in tickets:
+            scope = ticket.programs.first() or ticket.business_area
```
johniak, line 204: `programs` is m2m and `first()` has no ordering, so for a ticket linked to more than one programme the permission check lands on an arbitrary one. The same request can pass or fail depending on row order. Either check them all (`all(user.has_perm(p, program) for program in ticket.programs.all())`) or make it explicit why one is enough.

src/hope/apps/grievance/api/serializers/ticket_detail.py
```diff
@@ -194,30 +201,147 @@ class DeduplicationEngineSimilarityPairSerializer(serializers.Serializer):
...
-        else:
-            scope = BusinessArea.objects.filter(slug=business_area_slug).first()
-        if self.context["request"].user.has_perm(Permissions.GRIEVANCES_VIEW_BIOMETRIC_RESULTS.value, scope):
+        if self.context["na_can_view_biometric_results"]:
             return DeduplicationEngineSimilarityPairSerializer(obj.get("dedup_engine_similarity_pair")).data
         return {}
 
 
+class NaRoleHouseholdSerializer(serializers.ModelSerializer):
+    active_individuals_count = serializers.IntegerField(source="active_individuals.count", read_only=True)
```
johniak: `source="active_individuals.count"` costs one COUNT query per household — DRF's `get_attribute` walks to `active_individuals` and then calls the bound `count` method. It fires per role, per individual, plus once more for the synthetic HEAD entry in `get_roles_in_households`, and no `prefetch_related` can fix it. That's a good chunk of the 95 queries your new test pins in test_grievance_detail.py:2979, on the endpoint the comparison panel calls on every ticket click. Annotate it on the queryset instead, e.g. `Count("individuals", filter=Q(individuals__withdrawn=False, individuals__duplicate=False))` in the retrieve prefetch.

src/hope/apps/grievance/api/serializers/grievance_ticket.py
```diff
@@ -813,3 +813,51 @@ class BulkCloseGrievanceTicketsSerializer(serializers.Serializer):
...
+
+class BulkNeedsAdjudicationResultSerializer(serializers.Serializer):
+    """What the batch did: the tickets it resolved, and the ones it skipped as already closed."""
+
+    resolved = GrievanceTicketSimpleSerializer(many=True)
+    skipped_closed = GrievanceTicketSimpleSerializer(many=True)
+
+
+class BulkNeedsAdjudicationSerializer(serializers.Serializer):
+    tickets = NeedsAdjudicationResolutionSerializer(many=True, allow_empty=False)
```
johniak, line 853: No upper bound on the batch. Decisions survive paging and filtering by design, so one long triage session can accumulate hundreds of tickets, and the whole thing then runs in a single transaction holding `SELECT ... FOR UPDATE` on every row. Add a `max_length` here so the lock set and transaction length stay predictable.

johniak (changes requested):

src/hope/apps/grievance/api/serializers/ticket_detail.py
pkujawa: The programme branch was unreachable - ticket details are only rendered by `GrievanceTicketGlobalViewSet`, which is registered exclusively on the business-area router; `program_code` was never in the URL kwargs here and scope was always  the business area. Behaviour is unchanged from develop, and similarity_score is gated by the same flag the existing dedup_engine_similarity_pair already used (which already carried that score).

src/hope/apps/grievance/services/bulk_action_service.py
pkujawa: `Program.Meta.ordering = ("id",)` (models/program.py:357-359), and `QuerySet.first()` falls back to `order_by("pk")` when a queryset is unordered, so this isn't row-order dependent — it resolves to the same programme every time.                                                                                           

Beyond that, we do not manage `cross-programme` tickets in our code.
programs.first() is the existing convention for grievance ticket permission scoping - `check_creator_or_owner_permission` and `has_creator_or_owner_permission` both do exactly this, as does views.py:756.                      

Checking all programmes here would make this the one call site that scopes differently from every other grievance permission check. If we want to revisit multi-program scoping, it should be a change to the shared helpers, not to this one. It will be implemented when we actually implement multi-program tickets

src/hope/apps/grievance/services/bulk_action_service.py
pkujawa: There is deterministic ordering — `GrievanceTicket.Meta.ordering = ("status", "created_at")` 

Although, I will add .`order_by("pk")` for consistency with those.

src/hope/apps/grievance/api/serializers/grievance_ticket.py
pkujawa: added limit of 50

src/hope/apps/grievance/api/serializers/ticket_detail.py
pkujawa: Good catch on the cause. `active_individuals` is a property that builds a new filtered queryset every time it's read, so DRF runs a separate COUNT for every household it renders, and you're right that no prefetch can reach it.                                                                                           
I've moved the count into the prefetch queries that already load those households, so it costs nothing extra now. I covered `possible_duplicate` too.                       
                                                                                                                                                                                                                                                                                                                             
On the size of it, I measured before changing anything by swapping the field for a fixed number and re-running the test. It was 2 queries out of the 95, so smaller than it looked, but it grows with the number of roles on a ticket so it's still worth fixing. We're at 93 now.                                         
                                                                                                                                                                                                                                                                                                                             
I also added a test for the value itself. Nothing was checking that number before, and the frontend uses it to decide when to ask for a role reassignment.

johniak (approved):

PR #6240, opened 2026-07-23
[HOTFIX] 331115&330996: email url and group import fix

nerjan:
[AB#331115](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/331115)
[AB#330996](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/330996)
Fix Payment Plan Group emails now have full path url instead of relative path.

Import on payment plan group delivery was crashing and rolling back, leaving the export file deleted but the FileTemp row orphaned.
Real cause: after we delete the shared export FileTemp, the activity-log diff reads the old snapshot that still points at it, and nested_getattr only caught AttributeError — so the DoesNotExist killed the import. File delete isn't transactional, so it was already gone by then.
Fix: nested_getattr now treats a deleted relation as absent, and file deletes are deferred to on_commit so row and file stay in sync on rollback.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [ "setuptools", "wheel" ]
 
 [project]
 name = "hope"
-version = "4.17.0"
+version = "4.17.1"
 description = "HCT MIS is UNICEF's humanitarian cash transfer platform."
 readme = "README.md"
 license = { text = "None" }
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "frontend",
-  "version": "4.17.0",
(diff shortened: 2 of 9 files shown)


johniak (approved):
Both fixes look correct. Main open point: widening `nested_getattr` to swallow `ObjectDoesNotExist` also changes behavior in `SignatureMixin.update_signature_hash`, where a deleted relation now silently hashes as "None" instead of raising. Also worth adding a rollback test that pins the on_commit deferral — the current tests would pass without it.

Jeeziorny (approved):

PR #6243, opened 2026-07-24
[AB#331149]: Replace hardcoded frontend choices with API-driven choices

mmaciekk:
[AB#331149](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/331149)

## Summary

Replaces hardcoded choice lists in the frontend with the existing API-driven choices pattern (`RestService.restChoices*List()` via react-query, shape `{ name, value }`), and dedups the remaining UI-only filter dropdowns into shared constants. This removes duplication and keeps option lists in sync with the backend enums.

## Backend

- **New choices endpoint** `GET /api/rest/choices/payment-verification-plan-channel/` — added a `payment_verification_plan_channel` action to `ChoicesViewSet` sourcing `PaymentVerificationPlan.VERIFICATION_CHANNEL_CHOICES` (mirrors the existing status/sampling actions).
- **New choices endpoint** `GET /api/rest/choices/sex/` — added a global `sex` action to `ChoicesViewSet` sourcing `SEX_CHOICE`, so `useSexChoices` no longer needs the business-area-scoped `individuals/choices/` endpoint.
- **Payment verification status labels re-cased at the source** — `PaymentVerification.STATUS_CHOICES` labels changed from all-caps (`"NOT RECEIVED"`, `"RECEIVED"`, ...) to title case (`"Not Received"`, `"Received"`, ...), so the `payment-verification-status` choices endpoint returns properly cased names instead of relying on a frontend re-case that was a no-op on all-caps input. No migration needed — the field uses a callable (`choices=get_status_choices`). Status chips elsewhere are unaffected (they render the raw value, not the label).
- All new/renamed actions covered by contract tests in `tests/unit/api_contract/test_lookups.py` (+ generated response snapshots) and `tests/unit/apps/core/test_generate_openapi_command.py`.

## Frontend — API migration (high-priority, genuinely duplicated choices)

- **Gender/sex (4 files)** — shared hook `useSexChoices.ts` fetches the new global `sex` choices endpoint (react-query dedupes across consumers). Applied in Create/Edit Verification Plan, Create Communication, and Create Survey. Removes the identical inline `FEMALE/MALE/OTHER/NOT_COLLECTED/NOT_ANSWERED` array duplicated in all four.
- **Verification channel (3 files)** — shared hook `useVerificationChannelChoices.ts` consumes the new `payment-verification-plan-channel` endpoint, used in the Verification Records filter and the Create/Edit Verification Plan radios (both the tab-specific radios and the FULL LIST tab). Radio groups re-attach their `data-cy` (`radio-manual/rapidpro/xlsx`) from the choice value so Selenium selectors are preserved.
- **Payment verification status (2 files)** — shared hook `useVerificationStatusChoices.ts` (with loading/error handling) feeds VerifyManual and VerifyPaymentGrievance, **filtered client-side to `RECEIVED`/`NOT_RECEIVED`** (the business-rule subset for manual/grievance verification). Labels now come pre-cased from the backend (see above), so the frontend no longer re-cases them. `data-cy` preserved on VerifyManual.

## Frontend — filter-dropdown dedup (no API; booleans / UI toggles with no backend enum)

Extracted into `src/frontend/src/utils/constants.ts` and reused:

- `YES_NO_CHOICES` — the Yes/No dropdowns duplicated across `FilterIndividualsOnline` / `FilterIndividualsOffline`.
- `PROGRAM_STATE_CHOICES` — the Active/All programme-state dropdowns in `HouseholdFilter` / `PeopleFilter`.
- `AUTHORIZED_USER_ROLE_CHOICES` — the canEdit/canApprove/canMerge menu items **and** the render-value label ternary, duplicated across `AuthorizedUsersOnlineListCreate` / `...Edit`.

## Deliberately out of scope (surfaced during review)

- **Household withdrawn status filter** — single-use, so left inline rather than hoisted to a global constant (avoids over-abstraction).
- **Import-type & area-access radios** — single-use, frontend-only control-flow / string conventions with no backend enum.
- **`utils.ts` `sexToCapitalize` map** — a synchronous display helper that can't source from the async API without a broad refactor.
- **Other payment-verification enum label casing** (e.g. `VerificationPlanDetails`, `VerificationPlanChart`, the channel detail page, and the record status chips) — those still display raw/uppercase values by design; only `PaymentVerification.STATUS_CHOICES` (the manual/grievance verify status) was re-cased, since that's the one whose label reached the UI through this migration. Widening the re-casing further is out of scope here.


src/frontend/src/components/payments/VerifyManual.tsx
```diff
@@ -41,6 +42,17 @@ export function VerifyManual({
   const { showMessage } = useSnackbar();
   const queryClient = useQueryClient();
   const { programCode, businessAreaSlug } = useBaseUrl();
+  const { data: verificationStatusChoices } = useQuery<Array<Choice>>({
+    queryKey: ['verificationStatusChoices'],
+    queryFn: () => RestService.restChoicesPaymentVerificationStatusList(),
+  });
+  // Manual verification only allows marking a payment received / not received.
+  const statusChoices = (verificationStatusChoices ?? [])
```
johniak: While the query is in flight or after it fails, `statusChoices` is `[]` and `FormikRadioGroup` renders the Status label with zero radios — the user gets a dialog with a comment box and a Save button and no way to mark the payment, with nothing saying why. Please handle the states: spinner (or disabled field) while loading, and surface the error via `showApiErrorMessages`, which is already imported here.

src/frontend/src/components/grievances/PaymentGrievance/VerifyPaymentGrievance/VerifyPaymentGrievance.tsx
```diff
@@ -29,6 +30,14 @@ export function VerifyPaymentGrievance({
   const { showMessage } = useSnackbar();
   const queryClient = useQueryClient();
   const { businessArea } = useBaseUrl();
+  const { data: verificationStatusChoices } = useQuery<Array<Choice>>({
+    queryKey: ['verificationStatusChoices'],
+    queryFn: () => RestService.restChoicesPaymentVerificationStatusList(),
+  });
+  // Grievance verification only allows marking a payment received / not received.
+  const statusChoices = (verificationStatusChoices ?? []).filter((choice) =>
```
johniak: Same as in `VerifyManual`: an empty `statusChoices` renders the Status group with no radios at all, so a failed or pending choices request silently blocks the grievance verification with no error message. Add a loading state and show the API error instead of falling back to an empty list.

src/frontend/src/components/payments/CreateVerificationPlan.tsx
```diff
@@ -634,19 +633,7 @@ export const CreateVerificationPlan = ({
...
-                            dataCy: 'radio-rapidpro',
-                          },
-                          { value: 'XLSX', name: 'XLSX', dataCy: 'radio-xlsx' },
-                          {
-                            value: 'MANUAL',
-                            name: 'MANUAL',
-                            dataCy: 'radio-manual',
-                          },
-                        ]}
+                        choices={channelChoices}
```
johniak: Only the RANDOM SAMPLING tab was migrated — the FULL LIST tab at L461-481 still has the inline array. The same dialog now shows RAPIDPRO/XLSX/MANUAL on one tab and MANUAL/RAPIDPRO/XLSX (backend order) on the other, both bound to `verificationChannel`, and a new channel would appear on one tab only. Migrate both groups.

src/frontend/src/components/payments/EditVerificationPlan.tsx
```diff
@@ -661,19 +660,7 @@ export const EditVerificationPlan = ({
...
-                            dataCy: 'radio-rapidpro',
-                          },
-                          { value: 'XLSX', name: 'XLSX', dataCy: 'radio-xlsx' },
-                          {
-                            value: 'MANUAL',
-                            name: 'MANUAL',
-                            dataCy: 'radio-manual',
-                          },
-                        ]}
+                        choices={channelChoices}
```
johniak: Same half-migration here — the channel radios in the first tab panel (L479-499) keep the inline RAPIDPRO/XLSX/MANUAL array while this one comes from the API in MANUAL-first order. Both write to the same field, so the dialog contradicts itself depending on which tab is open.

src/frontend/src/components/payments/CreateVerificationPlan.tsx
```diff
@@ -164,6 +166,15 @@ export const CreateVerificationPlan = ({
   const { businessArea, baseUrl, programId: programCode } = useBaseUrl();
   const { isActiveProgram, isSocialDctType } = useProgramContext();
   const queryClient = useQueryClient();
+  const sexChoices = useSexChoices();
+  const { data: verificationChannelChoices } = useQuery<Array<Choice>>({
```
johniak: This query is copy-pasted into three files (here, `EditVerificationPlan.tsx:123`, `VerificationRecordsFilters.tsx:54`) and the `dataCy` map into two of them verbatim. The status filter is duplicated the same way in `VerifyManual` and `VerifyPaymentGrievance`. You already wrote `useSexChoices` for exactly this — `useVerificationChannelChoices` is the same five lines and matches the PR's goal.

src/hope/apps/core/api/views.py
```diff
@@ -192,6 +192,12 @@ def payment_verification_plan_sampling(self, request: Request) -> Response:
         resp = ChoiceSerializer(to_choice_object(PaymentVerificationPlan.SAMPLING_CHOICES), many=True).data
         return Response(resp)
 
+    @extend_schema(responses={200: ChoiceSerializer(many=True)})
+    @action(detail=False, methods=["get"], url_path="payment-verification-channel")
```
johniak: `verification_channel` is a `PaymentVerificationPlan` field, and every other action on that model is named `payment-verification-plan-*` (`-status`, `-sampling`), while `payment-verification-*` belongs to `PaymentVerification`. `payment-verification-plan-channel` would read correctly. Worth renaming now — once the client method and `PaymentVerificationChannelEnum` are generated and used, it's a breaking rename.

johniak (changes requested):

src/frontend/src/components/payments/EditVerificationPlan.tsx
nerjan: choices={channelChoices}

src/frontend/src/components/payments/CreateVerificationPlan.tsx
nerjan: choices={channelChoices}

src/frontend/src/components/payments/VerifyManual.tsx
```diff
@@ -41,6 +42,17 @@ export function VerifyManual({
...
+    .filter((choice) => ['RECEIVED', 'NOT_RECEIVED'].includes(choice.value))
+    .map((choice) => ({
+      ...choice,
+      dataCy: `choice-${choice.value.toLowerCase().replace(/_/g, '-')}`,
+    }));
```
nerjan: still hardcoded values, and it would render "RECEIVED" not Received I think.

src/frontend/src/components/grievances/PaymentGrievance/VerifyPaymentGrievance/VerifyPaymentGrievance.tsx
```diff
@@ -29,6 +30,14 @@ export function VerifyPaymentGrievance({
...
+    ['RECEIVED', 'NOT_RECEIVED'].includes(choice.value),
+  );
```
nerjan: also here we end up with UPPERCASE I think.

nerjan (changes requested):
Not sure what is the mechanism exactly but these choices are not going to change frequently for sure not within 1 session so it would be good to cache them/save to memory. If thats already the case its ok.

src/frontend/src/hooks/useSexChoices.ts
```diff
@@ -0,0 +1,18 @@
+import { Choice } from '@restgenerated/models/Choice';
+import { IndividualChoices } from '@restgenerated/models/IndividualChoices';
+import { RestService } from '@restgenerated/services/RestService';
+import { useQuery } from '@tanstack/react-query';
+import { useBaseUrl } from '@hooks/useBaseUrl';
+
+export function useSexChoices(): Choice[] {
```
domdinicola: pls review this function.. anything weird? @johniak @mmaciekk

johniak: What do you mean? It looks good as react hooks.

domdinicola: why for gender choices we are checking business areas? @johniak

mmaciekk: because individualChoices are per business area

johniak: Yes, you're right — it shouldn't be required. It only is because we added it to the action in the `IndividualViewSet`. We could move this into `ChoicesViewSet`.

Btw. I didn't know we were doing quizzes during code reviews. ;)

mmaciekk:
> Not sure what is the mechanism exactly but these choices are not going to change frequently for sure not within 1 session so it would be good to cache them/save to memory. If thats already the case its ok.

I'm working on caching in a separate PR with react-query optimization
https://github.com/unicef/hope/pull/6260

src/frontend/src/components/payments/VerifyManual.tsx
mmaciekk: Handled — while the status choices query is loading, the field now renders `<LoadingComponent />` instead of an empty radio group, and failures surface via `showApiErrorMessages(statusChoicesError, showMessage)` in a `useEffect` on `isError`. The dialog can no longer show a comment box + Save with no way to mark the payment.

src/frontend/src/components/grievances/PaymentGrievance/VerifyPaymentGrievance/VerifyPaymentGrievance.tsx
mmaciekk: Same fix applied here — `<LoadingComponent />` while the choices query is in flight, and `showApiErrorMessages` on error via a `useEffect`, so a pending/failed request no longer renders an empty Status group with no explanation.

src/frontend/src/components/payments/CreateVerificationPlan.tsx
mmaciekk: Done — both the FULL LIST and RANDOM SAMPLING tabs now bind to the API-driven `channelChoices` (from `useVerificationChannelChoices`). The inline array is gone, so the two tabs stay in sync and a new channel shows up on both.

src/frontend/src/components/payments/EditVerificationPlan.tsx
mmaciekk: Done — both tab panels now use `channelChoices` from `useVerificationChannelChoices`; the inline `RAPIDPRO/XLSX/MANUAL` array on the first panel is removed, so the dialog no longer contradicts itself between tabs.

src/frontend/src/components/payments/CreateVerificationPlan.tsx
mmaciekk: Extracted into hooks as suggested: `useVerificationChannelChoices` (which also attaches the `radio-<value>` dataCy) and `useVerificationStatusChoices`. CreateVerificationPlan, EditVerificationPlan and VerificationRecordsFilters consume the channel hook; VerifyManual and VerifyPaymentGrievance consume the status hook. No more copy-pasted query/dataCy map.

src/hope/apps/core/api/views.py
mmaciekk: Renamed the action to `payment-verification-plan-channel` to match the other `PaymentVerificationPlan` siblings (`-status`, `-sampling`). Done before the generated client method / enum were in use, so no breaking rename later.

src/frontend/src/components/payments/EditVerificationPlan.tsx
mmaciekk: Done — the FULL LIST tab now uses `choices={channelChoices}`.

src/frontend/src/components/payments/CreateVerificationPlan.tsx
mmaciekk: Done — the FULL LIST tab now uses `choices={channelChoices}`.

src/frontend/src/components/payments/VerifyManual.tsx
mmaciekk: Good catch. The earlier frontend re-cast used `formatNormalCaseValue` (lodash `startCase`), which is a no-op on all-caps input, so it still rendered "RECEIVED". Fixed at the source in 49fc0431bb: re-cased `PaymentVerification.STATUS_CHOICES` labels so the `payment-verification-status` choices endpoint returns "Received" / "Not Received", and removed the frontend re-cast. The `['RECEIVED','NOT_RECEIVED']` **value** filter stays hardcoded on purpose — manual verification is a business-rule subset (only those two statuses may be set here), which isn't derivable from the enum itself.

src/frontend/src/components/grievances/PaymentGrievance/VerifyPaymentGrievance/VerifyPaymentGrievance.tsx
mmaciekk: Same fix — labels now come cased from the backend (49fc0431bb: re-cased `PaymentVerification.STATUS_CHOICES`), so no more UPPERCASE here either. The two-status subset is likewise an intentional business-rule filter for grievance verification, not a backend enum we can fetch.

src/frontend/src/hooks/useSexChoices.ts
mmaciekk: Addressed — added a global `sex` action to `ChoicesViewSet` and pointed `useSexChoices` at `restChoicesSexList()`, so gender choices no longer go through the business-area-scoped `individuals/choices` endpoint. Thanks for flagging it @domdinicola @johniak.

src/frontend/src/hooks/useVerificationChannelChoices.ts
```diff
@@ -0,0 +1,16 @@
...
+import { useQuery } from '@tanstack/react-query';
+
+export function useVerificationChannelChoices(): Choice[] {
+  const { data } = useQuery<Array<Choice>>({
+    queryKey: ['verificationChannelChoices'],
+    queryFn: () => RestService.restChoicesPaymentVerificationPlanChannelList(),
+    staleTime: Infinity, // choices don't change within a session
+  });
+
+  return (data ?? []).map((choice) => ({
```
johniak: `data ?? []` hides errors. If this request fails, the Create/Edit Verification Plan dialog shows the Verification Channel section with no radios and no error message — the user just can't submit and doesn't know why. Same problem in `useSexChoices`. This is exactly what you fixed in VerifyManual after the last review. Return the whole query result like `useVerificationStatusChoices` does, so components can show a loader and the error.

src/frontend/src/containers/tables/payments/VerificationRecordsTable/VerificationRecordsFilters.tsx
```diff
@@ -47,18 +46,10 @@ export function VerificationRecordsFilters({
...
+  const verificationChannelChoices = useVerificationChannelChoices();
 
-  const verificationChannelChoices = [
-    { name: 'Manual', value: 'MANUAL' },
-    { name: 'RapidPro', value: 'RAPIDPRO' },
-    { name: 'XLSX', value: 'XLSX' },
-  ];
-
-  if (!verificationStatusChoices) {
+  if (!verificationStatusChoices || !verificationChannelChoices.length) {
```
johniak: With `!verificationChannelChoices.length` added here, a failed channel request hides the whole filter bar forever, with no message — the hook returns `[]` on error. Before, the channel list was hardcoded so this couldn't happen. Once the hook exposes loading/error, show those instead of returning `null`.

johniak (changes requested):

domdinicola:
@mmaciekk pls fix conflicts and change requests

src/frontend/src/hooks/useVerificationChannelChoices.ts
mmaciekk: Fixed. `useVerificationChannelChoices` (and `useSexChoices`) now return the full `UseQueryResult<Array<Choice>>`, like `useVerificationStatusChoices` — no more `data ?? []` swallowing failures. The `radio-<value>` `data-cy` decoration moved into react-query's `select`, so the Selenium selectors are unchanged.

Consumers (Create/Edit Verification Plan, Create Communication, Create Survey) destructure `{ data = [], isLoading, isError, error }`, render `<LoadingComponent />` in place of the field while the query is in flight, and surface failures through the new `useApiErrorSnackbar(isError, error)` hook — which also replaces the duplicated `useEffect` + `showApiErrorMessages` block in `VerifyManual` / `VerifyPaymentGrievance`.

While merging `develop` I also switched all three choices hooks to `restQueryKey(RestService.restChoices*List)` to match the new query-key convention.

src/frontend/src/containers/tables/payments/VerificationRecordsTable/VerificationRecordsFilters.tsx
mmaciekk: Fixed. The `!verificationChannelChoices.length` guard is gone — the early `return null` was removed entirely. `VerificationRecordsFilters` now always renders the filter bar; both choice lists default to `[]` while loading, and `useApiErrorSnackbar` surfaces a failed status or channel lookup through the snackbar. A failed lookup can no longer make the whole filter bar disappear.

src/frontend/src/containers/tables/payments/VerificationRecordsTable/VerificationRecordsFilters.tsx
```diff
@@ -48,20 +47,20 @@ export function VerificationRecordsFilters({
...
+  const {
+    data: verificationStatusChoices = [],
+    isError: isStatusChoicesError,
+    error: statusChoicesError,
+  } = useVerificationStatusChoices();
+  const {
+    data: verificationChannelChoices = [],
+    isError: isChannelChoicesError,
+    error: channelChoicesError,
+  } = useVerificationChannelChoices();
```
nerjan, line 59: this now renders the uppercase channel names ("MANUAL"/"RAPIDPRO"/"XLSX") — the ol array had the as "Manual"/"RapidPro"/"XLSX".

nerjan (dismissed):

nerjan (dismissed):

domdinicola:
@johniak I created https://dev.azure.com/unicef/ICTD-HCT-MIS/_workitems/edit/333810 to not lose track of this

src/frontend/src/hooks/useApiErrorSnackbar.ts
```diff
@@ -0,0 +1,15 @@
...
+// Surfaces a failed query through the snackbar, so a component never renders an
+// empty field with no explanation of why the data is missing.
+export function useApiErrorSnackbar(isError: boolean, error: unknown): void {
+  const { showMessage } = useSnackbar();
+
+  useEffect(() => {
+    if (isError) {
+      showApiErrorMessages(error, showMessage);
+    }
+  }, [isError, error, showMessage]);
```
johniak: `showMessage` in the deps makes this effect re-fire forever. `SnackbarProvider` builds `{showMessage, message, showRestApiError}` as a fresh object literal every render and nothing is memoised, so every consumer re-renders when the snackbar opens or auto-hides after 5s. The effect then re-runs, `isError` is still true, and the toast comes back — every 5 seconds, for as long as the page is open, covering every other message. `CreateVerificationPlan`, `EditVerificationPlan` and `VerificationRecordsFilters` each call this twice, so two effects alternate two messages. Fire once per error identity instead:

```ts
const reported = useRef<unknown>(null);

useEffect(() => {
  if (isError && reported.current !== error) {
    reported.current = error;
    showApiErrorMessages(error, showMessage);
  }
  if (!isError) reported.current = null;
  // showMessage omitted on purpose: new closure on every provider render
}, [isError, error]);
```

johniak (changes requested):

nerjan (approved):

johniak (approved):

PR #6249, opened 2026-07-27
feat(currency): Handle two currencies sharing one ISO code but differing by vision code

kosior:
Currency Business Rules:
- at most one active row per ISO code
- unique vision_code

Other:
- VisionAPI.send_payment_plan - use vision_code
- get_unore_exchange_rate - use vision_code
- Payment.utils.get_exchange_rate - use vision_code

Most important changes:
- src/hope/models/currency.py
- src/hope/api/utils.py
- src/hope/admin/currency.py - deprecate_currency action - should be used when we want to deprecate the vision_code


src/hope/admin/currency.py
```diff
@@ -8,5 +11,35 @@
 class CurrencyAdmin(HOPEModelAdminBase):
     list_display = ("code", "name", "is_crypto", "vision_code", "active", "number_of_decimals")
     list_filter = ("is_crypto", "active")
-    search_fields = ("code", "name")
+    search_fields = ("code", "vision_code", "name")
     ordering = ("code", "vision_code")
+    actions = ("deprecate_currency",)
+
+    @admin.action(description="Deprecate: swap the active currency for this code")
```
domdinicola: not sure this is needed... what's the case?

src/hope/api/endpoints/rdi/lax.py
```diff
@@ -557,11 +558,11 @@ class HouseholdSerializer(serializers.ModelSerializer):
     consent_sharing = serializers.MultipleChoiceField(choices=DATA_SHARING_CHOICES, required=False)
     village = serializers.CharField(allow_blank=True, allow_null=True, required=False)
     consent_sign = serializers.CharField(allow_blank=True, required=False)
-    currency = serializers.SlugRelatedField(
+    currency = CurrencySlugRelatedField(
         slug_field="code",
         required=False,
         allow_null=True,
-        queryset=Currency.objects.all(),
+        queryset=Currency.objects.active(),
```
domdinicola, line 564: what happen to existing data using inactive currency?

kosior: Household, Payment and PaymentPlan all have FK to Currency model. The current PR is not changing that. 

The new objects will use active Currency.

|code|vision code|active|
|---|---|---|
|SYP|SYP|No|
|SYP|SYP01|Yes

.active().get('SYP') will return the one with the **SYP01** vision code. There can be at most one active currency per ISO code.

src/hope/models/currency.py
```diff
@@ -13,17 +25,23 @@ class Currency(models.Model):
...
 
+    objects = CurrencyManager()
+
     class Meta:
         app_label = "core"
         ordering = ["code", "vision_code"]
         verbose_name_plural = "currencies"
         constraints = [
-            models.UniqueConstraint(fields=["code", "vision_code"], name="unique_code_vision_code"),
+            models.UniqueConstraint(Lower("code"), condition=models.Q(active=True), name="unique_code_active"),
```
johniak, line 46: Dropping the unique `code` means two rows can share a code, and with `ordering = ["code", "vision_code"]` a `filter(code="SYP").first()` returns the old `SYP` row before `SYP01`. Two call sites weren't updated and will now pick the deprecated currency: `grievance/services/data_change/household_data_update_service.py:183` and `individual_data_update_service.py:302`. Both should use `Currency.objects.resolve_code(...)`.

tests/unit/models/test_currency_query_patterns.py
```diff
@@ -1,93 +1,91 @@
...
+def test_handle_currency_field_resolves_active_row_for_shared_code():
+    Currency.objects.create(code="SYP", name="Syrian pound Old", vision_code="SYP", active=False)
+    new = Currency.objects.create(code="SYP", name="Syrian pound", vision_code="SYP01", active=True)
+
+    assert handle_currency_field("SYP", "currency", None, None, None) == new
+
+
+@pytest.mark.django_db
+def test_handle_currency_field_returns_none_for_unknown_code():
+    assert handle_currency_field("MISSING", "currency", None, None, None) is None
```
johniak: This test fails — `handle_currency_field` now calls `resolve_code`, which raises `Currency.DoesNotExist` instead of returning `None` for an unknown code. Either assert `pytest.raises(Currency.DoesNotExist)` or keep the handler returning `None`; right now the test still encodes the old contract.

tests/unit/admin/test_currency_admin.py
```diff
@@ -0,0 +1,78 @@
...
+@pytest.mark.django_db
+def test_deprecate_requires_one_active_one_inactive(admin_instance, request_with_messages):
+    # Two inactive rows sharing a code is allowed by the constraint but invalid for deprecation.
+    Currency.objects.create(code="SYP", name="old", vision_code="SYP", active=False)
+    Currency.objects.create(code="SYP", name="new", vision_code="SYP01", active=False)
+
+    admin_instance.deprecate_currency(request_with_messages, Currency.objects.filter(code="SYP"))
+
+    messages = [str(m) for m in get_messages(request_with_messages)]
+    assert any("one selected currency must be active" in m for m in messages)
```
johniak, line 91: This test fails — with two inactive rows the action hits the first branch and emits "Select exactly two currencies: one active and one inactive.", which doesn't contain "one selected currency must be active". Looks like the assertion was written against an earlier version of the message.

src/hope/api/endpoints/rdi/upload.py
```diff
@@ -183,11 +183,11 @@ class HouseholdSerializer(serializers.ModelSerializer):
     size = serializers.IntegerField(required=False, allow_null=True)
     consent_sharing = serializers.MultipleChoiceField(choices=DATA_SHARING_CHOICES, required=False)
     village = serializers.CharField(allow_blank=True, allow_null=True, required=False)
-    currency = serializers.SlugRelatedField(
+    currency = CurrencySlugRelatedField(
         slug_field="code",
         required=False,
         allow_null=True,
-        queryset=Currency.objects.all(),
+        queryset=Currency.objects.active(),
```
johniak, line 189: The push endpoints now accept only active currencies, but the lookup they're discovered from — `CurrencyAPIView` in `api/endpoints/lookups/base.py:55` — is still `Currency.objects.all()` and its serializer exposes no `active` flag. Once a code is deprecated, integrators see two identical `code` entries and can get a validation error for a code this endpoint advertised. Worth switching that queryset to `active()`.

src/hope/apps/universal_update_script/universal_individual_update_service/validator_and_handlers.py
```diff
@@ -110,15 +110,15 @@ def handle_currency_field(
 ) -> Currency | None:
     if value is None or value == "":
         return None
-    return Currency.objects.filter(code=value).first()
+    return Currency.objects.resolve_code(value)
```
johniak: The signature still says `-> Currency | None`, but this can no longer return `None` for an unknown code — it raises. It only stays safe because `validate_currency` runs in an earlier pass; any direct caller, or a currency deactivated between the two passes, gets an uncaught `DoesNotExist`. Update the annotation to match whichever contract you settle on.

src/hope/apps/core/management/commands/demo_data/core.py
```diff
@@ -173,6 +173,173 @@
...
+    ("SGD", "Singapore dollar", False),
+    ("SHP", "Saint Helena pound", False),
+    ("SLE", "Sierra Leonean leone", False),
+    ("SOS", "Somali shilling", False),
+    ("SRD", "Surinamese dollar", False),
+    ("SSP", "South Sudanese pound", False),
+    ("STN", "São Tomé and Príncipe dobra", False),
+    ("SVC", "Salvadoran colón", False),
+    ("SYP", "Syrian pound Old", False),
+    ("SYP01", "Syrian pound", False),
```
johniak: These two entries still model `SYP01` as its own ISO code, which is the shape this PR is replacing. A fresh demo DB therefore never reproduces the shared-code case — one `code="SYP"` with `vision_code="SYP01"` plus a deprecated `SYP` row would actually exercise it.

src/hope/apps/core/management/commands/demo_data/core.py
```diff
@@ -173,6 +173,173 @@
 }
 
 
+CURRENCIES = [
```
johniak: This list is a verbatim copy of `CURRENCIES` in `core/migrations/0020_migration.py`. Dropping the `importlib` import of a migration module was the right call, but two copies of 160 entries will drift the first time someone adds a currency — worth importing the constant from one place.

src/hope/api/utils.py
```diff
@@ -1,8 +1,19 @@
...
+from django.utils.encoding import smart_str
+from rest_framework import serializers
 from rest_framework.authentication import SessionAuthentication
 
 
+class CurrencySlugRelatedField(serializers.SlugRelatedField):
+    def to_internal_value(self, data: Any) -> Any:
+        queryset = self.get_queryset()
+        try:
+            return queryset.resolve_code(data)
```
johniak, line 15: The field accepts `slug_field` (every call site passes `"code"`) but this always resolves by `code`. If anyone ever passes `slug_field="vision_code"`, reads use `vision_code` and writes silently use `code`. Either honour the kwarg or hardcode `slug_field = "code"` on the class and drop it from the call sites.

tests/unit/api/test_currency_slug_field.py
```diff
@@ -0,0 +1,60 @@
...
+from hope.models.currency import Currency
+
+
+class _CurrencyCarrierSerializer(serializers.Serializer):
+    currency = CurrencySlugRelatedField(slug_field="code", queryset=Currency.objects.all(), allow_null=True)
+
+
+@pytest.mark.django_db
+def test_field_resolves_active_currency():
+    currency = Currency.objects.create(code="TST", name="Test", active=True)
```
johniak: The new tests build data with `Currency.objects.create(...)` inline instead of local fixtures using `CurrencyFactory`, and none of them assert query counts — same in `tests/unit/admin/test_currency_admin.py` and `tests/unit/models/test_currency_query_patterns.py`. `test_get_quantity_in_usd_keys_fx_feed_by_vision_code` in this PR does it the project way; worth aligning the rest, especially since `resolve_code` runs per imported row.

tests/unit/api/test_currency_slug_field.py
```diff
@@ -0,0 +1,60 @@
...
+    new = Currency.objects.create(code="SYP", name="Syrian pound", vision_code="SYP01", active=True)
+
+    serializer = _CurrencyCarrierSerializer(data={"currency": "SYP"})
+
+    assert serializer.is_valid(), serializer.errors
+    assert serializer.validated_data["currency"] == new
+
+
+@pytest.mark.django_db
+def test_field_shared_code_does_not_raise_multiple_objects_returned():
```
johniak: Same setup as `test_field_resolves_active_row_for_shared_code` above, with a weaker assertion — that test already proves no `MultipleObjectsReturned`.

johniak (changes requested):
This PR does not run CI because is not merged to develop/staging/master or long-term/* this is risky

src/hope/admin/currency.py
kosior: It's not necessary. You can achieve the same end state manually by deactivating (SYP, SYP) first and then activating (SYP, SYP01) (the order matters because of the unique-active constraint). The action just wraps both updates in one atomic transaction, so there's never a moment where SYP has no active row.

src/hope/models/currency.py
kosior: fixed

tests/unit/models/test_currency_query_patterns.py
kosior: fixed

tests/unit/admin/test_currency_admin.py
kosior: fixed

src/hope/api/endpoints/rdi/upload.py
kosior: fixed

src/hope/apps/universal_update_script/universal_individual_update_service/validator_and_handlers.py
kosior: Fixed. I added `resolve_code_or_none`

src/hope/apps/core/management/commands/demo_data/core.py
kosior: CURRENCIES list updated

src/hope/api/utils.py
kosior: slug_field hardcoded. Good idea.

src/hope/admin/currency.py
johniak: Without `permissions=[...]` this action is offered to any staff user who can open the changelist — Django only filters actions that declare `allowed_permissions`, so a view-only Currency admin can swap which currency is active for a code. Make it `@admin.action(description=..., permissions=["change"])`.

tests/unit/apps/grievance/services/data_change/test_household_data_update_service.py
```diff
@@ -701,3 +701,23 @@ def test_close_resolves_currency_from_code(all_currencies) -> None:
 
     usd_currency = Currency.objects.get(code="USD")
     assert household.currency == usd_currency
+
+
+def test_close_resolves_active_currency_for_shared_code() -> None:
+    Currency.objects.create(code="SYP", name="Syrian pound Old", vision_code="SYP", active=False)
```
johniak: These currencies are built inline in the test body; the convention — and what the other files in this PR now do after the round-1 fixes (e.g. `test_currency_slug_field.py`) — is `CurrencyFactory` inside a local fixture. Same pattern in `test_currency_model_references.py:74` and the new test in `test_update_individual_data_service.py`.

src/hope/admin/currency.py
```diff
@@ -8,5 +11,35 @@
...
+
+        old, new = active[0], inactive[0]
+        if old.code != new.code:
+            self.message_user(request, "Selected currencies must share the same code.", level=messages.ERROR)
+            return
+
+        # Deactivate first, then activate
+        with transaction.atomic():
+            Currency.objects.filter(pk=old.pk).update(active=False)
+            Currency.objects.filter(pk=new.pk).update(active=True)
```
johniak, line 39: The swap writes no admin `LogEntry`, so there's no record of who deprecated a currency or when. Since this silently changes which FX rate payments get, a `self.log_change(request, ...)` on both rows is cheap insurance.

johniak (changes requested):

johniak (changes requested):
.

saxix (commented):
witing for PR stack to be merged

src/hope/admin/currency.py
kosior: Done 👍

src/hope/admin/currency.py
kosior: Good idea. Done.

PR #6253, opened 2026-07-27
Frontend TypeScript type-safety audit: remove type suppressions

mmaciekk:
## Summary

Audit and remediation of TypeScript type-suppression debt in the frontend app code. The frontend compiled cleanly before only because `tsconfig` has `strict: false` and ESLint has `no-explicit-any`/`ban-ts-comment` off — so `@ts-ignore`, `as any`, and `: any` were silently masking type holes rather than expressing real types. This replaces the tractable suppressions with real types and documents the rest.

**Scope:** app code only. Generated `restgenerated/**` is untouched. No `tsconfig`/CI/lint-rule changes. No behavior changes — this is a typing pass.

## Impact (app code, excluding `restgenerated/`)

| Suppression | Before | After |
|---|--:|--:|
| `: any` | 325 | 206 |
| `@ts-ignore` | 47 | 31 |
| `as any` | 34 | 22 |
| `as unknown as` | 7 | 7 (all now documented) |
| `@ts-expect-error` | 0 | 3 (self-documenting) |

~147 suppressions removed; the remainder are documented codegen-workarounds or genuinely-dynamic Formik form bags.

## Changes by phase

- **Phase 1 — shared plumbing:** new `shared/autocompletes/types.ts` + generic `BaseAutocompleteFilterRest<TOption>` (removed ~50 `: any` across 7 filters); typed `utils.ts` camelize + API-error helpers; typed `api.ts` errors (removed all 4 `@ts-ignore` + 2 `as any` via a typed error).
- **Phase 2 — managerialConsole:** new `managerialConsole/types.ts`; typed the 3 sections, page, and modals from `PaymentPlan`/`PaginatedPaymentPlanList` models; a `ManagerialStatus` union removed the page's `@ts-ignore` (34 → 4).
- **Phase 3 — grievances (clean wins + documented rest):** 15 `onError: (error: any)` handlers → exported `ApiErrorShape`; `AssignedToDropdown`, `EditIdentityRow`, `CreateGrievancePage`, `RequestedHouseholdDataChange`, `createGrievanceUtils` typed. Dynamic form bags + loose codegen fields-attributes data left as documented intentional `any`.
- **Phase 4 — long tail:** typed `renderDeduplicationStatus` (6 `as any` removed); legacy IE `document.selection` via a single `LegacyDocument` cast; `BlackLink`, `useInterval`; `TargetingCriteriaForm` `@ts-ignore` → `@ts-expect-error` with reasons; documented all 7 `as unknown as`.

## Deliberately surfaced, not forced

- **`adjustHeadCells` / `UniversalRestTable` (deferred):** typing `adjustHeadCells` properly exposes a real pre-existing generic mismatch in 6 tables (their `headCells` generic disagrees with `renderRow`). That's cross-file work for the tables area — reverted here to keep this PR green; tracked for follow-up.
- **Suspected latent bug (flagged, not fixed):** `ImportedPeopleTableRow`'s golden-record else-branch passes `*Results` arrays where a `*Status` string is expected (the batch branch uses `*Status`). Left behavior-identical with a `FIXME`.

## Verification

- `tsc --noEmit` — 0 errors
- `bun run lint` — clean
- `bun run test` — 37 passing


src/frontend/src/containers/tables/rdi/ImportedPeopleTable/ImportedPeopleTableRow.tsx
```diff
@@ -128,6 +128,9 @@ export function ImportedPeopleTableRow({
             individual={individual}
           />
         ) : (
+          // FIXME: this passes *Results arrays where a status string is expected
```
johniak: This one isn't just suspected — it reproduces. The else-branch only runs when both result arrays are empty, so `renderDeduplicationStatus([], [], dict)` looks up `dict[String([])]` i.e. `dict[""]`, which is undefined, and the Golden Record cell renders blank while the Batch cell beside it shows the right label. Same two lines as the batch branch fixes it and makes your new `string` signature honest:

```tsx
renderDeduplicationStatus(
  individual.deduplicationGoldenRecordStatus,
  individual.biometricDeduplicationGoldenRecordStatus,
  deduplicationGoldenDict,
)
```

If you'd rather not change behaviour here, a ticket beats a FIXME parked next to a signature you just tightened.

src/frontend/src/shared/autocompletes/types.ts
```diff
@@ -0,0 +1,27 @@
...
+  id: string;
+  name: string;
+}
+
+/**
+ * Props shared by the REST autocomplete filter components. Individual filters
+ * intersect this with their own extras (e.g. `label`, `additionalVariables`)
+ * and may relax required fields to optional where their callers omit them.
+ */
+export interface BaseAutocompleteFilterProps {
```
johniak: `BaseAutocompleteFilterProps` is exported but referenced nowhere — all five filters still spell their props out inline. The docblock even describes filters intersecting it with their own extras, which no file does. Either wire it into the filters (presumably the point) or drop it; `bun run prune-unused-exports` will flag it either way.

src/frontend/src/containers/forms/TargetingCriteriaForm.tsx
```diff
@@ -255,7 +255,8 @@ export const TargetingCriteriaForm = ({
   const filteredIndividualData = useMemo(
     () => ({
       allFieldsAttributes: data
-        //@ts-ignore
+        // @ts-expect-error codegen types this endpoint as PaginatedFieldAttributeList,
```
johniak: The comment is right about the symptom but the cause is fixable on the backend: `all_fields_attributes` (`src/hope/apps/core/api/views.py:111`) doesn't set `pagination_class=None`, so spectacular wraps the declared `FieldAttributeSerializer(many=True)` response anyway and injects `limit`/`offset`. The sibling `all_kobo_projects` three lines below does set it and generates `Array<KoboAssetObject>` correctly. One line removes all three directives:

```python
@action(detail=True, methods=["get"], url_path="all-fields-attributes", pagination_class=None)
```

src/frontend/src/containers/pages/grievances/CreateGrievancePage.tsx
```diff
@@ -83,6 +83,10 @@ function EmptyComponent(): ReactElement {
   return null;
 }
 
+// `rolesInHousehold` is generated as `Record<string, any>` but is an array of
+// role entries at runtime; this is the shape the delegate lookup relies on.
+type HouseholdRoleEntry = { role: string; individual?: unknown };
```
johniak: Same story as the fields-attributes casts: `rolesInHousehold` is `Record<string, any>` only because `get_roles_in_household` (`src/hope/apps/household/api/serializers/household.py:357`) is annotated `-> dict` while returning `many=True` data. `@extend_schema_field` is already used ~32× in this repo including the neighbouring `individual.py`, and fixing it removes the need for this shim. Also `individual` is never absent — the serializer lists it in `fields` — so it shouldn't be optional here.

johniak (changes requested):

src/frontend/src/containers/tables/rdi/ImportedPeopleTable/ImportedPeopleTableRow.tsx
mmaciekk: You're right — it reproduced. Fixed in 0ffebabc3a: the else-branch now mirrors the batch branch and passes the `*Status` fields, so the Golden Record cell renders the real label. Removed the `FIXME` and both `as any` casts, and the `string` signature is now honest.

```tsx
) : (
  renderDeduplicationStatus(
    individual.deduplicationGoldenRecordStatus,
    individual.biometricDeduplicationGoldenRecordStatus,
    deduplicationGoldenDict,
  )
)
```

src/frontend/src/shared/autocompletes/types.ts
mmaciekk: Dropped it in 0ffebabc3a (along with its now-unused `Filter` import). `AutocompleteOption` stays since it's used. Wiring a shared props interface into the five filters is a broader refactor better left out of this typing pass.

src/frontend/src/containers/forms/TargetingCriteriaForm.tsx
mmaciekk: Done the backend way in 0ffebabc3a — added `pagination_class=None` to the `all-fields-attributes` action:

```python
@action(detail=True, methods=["get"], url_path="all-fields-attributes", pagination_class=None)
```

The endpoint now generates `Array<FieldAttribute>` and all three `@ts-expect-error` directives are gone. Regenerating the types also surfaced two other consumers (`FlexFieldTab`, `AddFilterTargetingCriteriaDisplay`) that were reading `.results` off what is actually a bare array — i.e. `undefined` at runtime, hidden by the old mistyping — so those are fixed too. The runtime response is unchanged (the action returns `Response(...)` directly), which the contract tests confirm.

src/frontend/src/containers/pages/grievances/CreateGrievancePage.tsx
mmaciekk: Fixed at the source in 0ffebabc3a — annotated `get_roles_in_household` to match the ~32 existing uses:

```python
@extend_schema_field(IndividualRoleInHouseholdForHouseholdSerializer(many=True))
def get_roles_in_household(self, obj: Household) -> dict:
```

`rolesInHousehold` now generates as `Array<IndividualRoleInHouseholdForHousehold>`, so the `HouseholdRoleEntry` shim is gone — `CreateGrievancePage` uses the generated model, and since the serializer lists `individual` in `fields` the generated `individual` is required (no more `individual?`).

pkujawa (dismissed):

johniak (dismissed):

domdinicola:
@mmaciekk pls resovle conflicts

PR #6050, opened 2026-05-29
[WIP] 317775 & 316500: Deduplication initiation goes to Country Workspace

Jeeziorny:
[AB#317775](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/317775)
[AB#316500](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/316500)

## Status table

| ID | Done | Description | Decisions |
|---|---|---|---|
| **BE-01** | ✅ | Either Business Area can push RDI only from Country Workspace (and have automerge flow with biometric deduplication) or only from other sources, like KoBo or XLSX (and have no biometric deduplication). Once BA starts uploading only through Country Workspace, they cannot go back to other methods. BE-01 adds a per-BA `ingest_source` CharField that partitions ingest. The field is one-way at the model layer (`COUNTRY_WORKSPACE_ONLY → ALL_EXCEPT_COUNTRY_WORKSPACE` rejected) and admin-readonly once set to `COUNTRY_WORKSPACE_ONLY`. | 1 |
| **BE-02** | ✅ | Drops the dead serializer-only `individual_id` field from the lax `IndividualSerializer` and simplifies `_prepare_individual` + the `individual_id_mapping` response. | 2 |
| **BE-03** | ✅ | Makes `country_workspace_id` required on both `lax.IndividualSerializer` and `PushPeopleSerializer`| 3 |
| **BE-04** | ✅ | Deletes every HOPE→engine approve/reject ack call site, along with `report_ack_to_biometric_deduplication_engine` / `report_rdi_approved` / `report_individuals_status` and the BIOMETRIC_DEDUPLICATION_REPORT_INDIVIDUALS_STATUS. Going forward, the only way HOPE talks to dedup is `GET /findings/` — a CW push *is* the approval. |  |
| **BE-05** | ✅  | Removing legacy biometric stuff. Detailed list below | 4 |
| **BE-06** | ✅ | Reduces `DeduplicationEngineAPI` to a single method hitting `deduplication_sets/{deduplication_set_id}/findings/` with `deduplication_set_id = rdi.country_workspace_id` |  |
| **BE-07** | ✅ | Rework `CompleteRDIView`. When CW completes uploading RDI `process_country_workspace_rdis` is being called and Country Workspace RDI(s) are getting processed. |  |
[...]

PR #6256, opened 2026-07-28
[LONG TERM] 317775 & 316500: New biometric dedup flow

Jeeziorny:
[AB#317775](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/317775)
[AB#316500](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/316500)

## Status table

| ID | Done | Description | Decisions |
|---|---|---|---|
| **BE-01** | ✅ | Either Business Area can push RDI only from Country Workspace (and have automerge flow with biometric deduplication) or only from other sources, like KoBo or XLSX (and have no biometric deduplication). Once BA starts uploading only through Country Workspace, they cannot go back to other methods. BE-01 adds a per-BA `ingest_source` CharField that partitions ingest. The field is one-way at the model layer (`COUNTRY_WORKSPACE_ONLY → ALL_EXCEPT_COUNTRY_WORKSPACE` rejected) and admin-readonly once set to `COUNTRY_WORKSPACE_ONLY`. | 1 |
| **BE-02** | ✅ | Drops the dead serializer-only `individual_id` field from the lax `IndividualSerializer` and simplifies `_prepare_individual` + the `individual_id_mapping` response. | 2 |
| **BE-03** | ✅ | Makes `country_workspace_id` required on both `lax.IndividualSerializer` and `PushPeopleSerializer`| 3 |
| **BE-04** | ✅ | Deletes every HOPE→engine approve/reject ack call site, along with `report_ack_to_biometric_deduplication_engine` / `report_rdi_approved` / `report_individuals_status` and the BIOMETRIC_DEDUPLICATION_REPORT_INDIVIDUALS_STATUS. Going forward, the only way HOPE talks to dedup is `GET /findings/` — a CW push *is* the approval. |  |
| **BE-05** | ✅  | Removing legacy biometric stuff. Detailed list below | 4 |
| **BE-06** | ✅ | Reduces `DeduplicationEngineAPI` to a single method hitting `deduplication_sets/{deduplication_set_id}/findings/` with `deduplication_set_id = rdi.country_workspace_id` |  |
| **BE-07** | ✅ | Rework `CompleteRDIView`. When CW completes uploading RDI `process_country_workspace_rdis` is being called and Country Workspace RDI(s) are getting processed. |  |
| **BE-08** | ✅ | At every RDI creation entry point, rejects requests that mismatch the BA's `ingest_source`. All places listed below |  |
| **BE-10** | ✅ | Adds an ability to rerun *failed* RDI from admin panel. By failed I mean MERGE_SCHEDULED, MERGE_ERROR and IMPORT_ERROR. |  |
| **BE-11** | ✅ | CreateRDIView, PushToRDIView, PushLaxToRDIView and CompleteRDIView can be called by BusinessAreas that have INGEST_SOURCE set to COUNTRY_WORKSPACE_ONLY |  |

### BE-05:
- remove "Fetch biometric results" admin button,
- remove run deduplication endpoint (`registration_data/api/views.py`)
- remove webhook deduplication
- remove Permissions.RDI_RERUN_DEDUPE and Permissions.RDI_WEBHOOK_DEDUPLICATION
- remove `src/hope/admin/program.py`  `_handle_biometric_deduplication_set` and `clean` 
- remove RDI.deduplication_engine_status - removed together with choices.
- remove CountryWorkspaceIdConditionalMixin - it was bad approach from first iteration when adjusting to new deduplication flow
- remove part of `close_needs_adjudication_new_ticket` with reporting back to deduplication engine removed
- remove`BIOMETRIC_DEDUPLICATION_REPORT_INDIVIDUAL_STATUS` (individuals no longer are reported from HOPE to dedup)

### BE-08: (CW - Country Workspace)
| Location | Comment |
|---|---|
| `api/endpoints/rdi/base.py` | Endpoints used only by country workspace. Each has `BusinessAreaIngestCWOnlyMixin` allowing access ONLY IF the BA uses ONLY CW for uploading RDIs |
| `api/endpoints/rdi/upload.py:264` | BA must use ingest sources all except CW |
| `apps/registration_data/api/views.py:413, 494` | BA must use ingest sources all except CW |
| `contrib/aurora/services/base_flex_registration_service.py:55` | BA must use ingest sources all except CW |
| `apps/generic_import/api/views.py:79` | BA must use ingest sources all except CW |
| `apps/generic_import/views.py:87` | BA must use ingest sources all except CW |
| `apps/registration_data/api/serializers.py:182` | Dropped — guard is on viewset |
| `apps/program/utils.py:322` | Dropped — RDI created in MERGED status |
| `apps/program/utils.py:424` | Same as above |
| `admin/registration_data.py:87` | Buttons enabled/disabled using BA ingest source criteria |
| `admin/record.py:283` | Checks added at `BaseRDIForm` and `AmendRDIForm` |

## Decisions
1. Cannot express XOR condition with boolean value and make it readable. That's why I used char field with two states
2. CW team told me, that individual_id is the same as country_workspace_id. No need to keep both. country_workspace_id was chosen as the right one, because PushPeopleSerialiser contains also country_workspace_id.
3. From discussion with people in the team, lax.py and push_people.py contain logic used only by country workspace. No need to check that on request level.
4. `bulk_upload_individuals_photos` - I left it intact - it might be used by some other internals that I'm not aware of: **update** Marek told me that it can be removed too.
5. ProgramDetailSerializer - `can_import_rdi` -> can't manual import when given BA ingest source is CW only, can otherwise

# Pre-deployment checks
1. Assert, that BAs with `biometric_deduplication_enabled=True` have NO UNMERGED RDIs
2. PR Introduces unique constraint on `(BA, country_workspace_id)`. Make sure, that there are NO violation for this rule on given environment.


Jeeziorny:
@domdinicola requested renaming the branch. Github way of renaming branches is to create a new one with new PR, so all review comments are pruned. The same PR (but under different branch name and PR name) is here: https://github.com/unicef/hope/pull/6050

Jeeziorny:
Document was written in 25 August 2026. Treat this comment as a help for understanding the changes on this branch. PR description is more of an action / decision log.

## Summary

Country Workspace drives biometric deduplication flow. HOPE reads findings and merges RDI into population immediately as it comes from CW.

## Impact

New flow is enabled when two conditions are met.
1. BA has biometric deduplication enabled.
2. BA can push RDI using Country Workspace app.

Whether BA pushes RDIs from CW or not is controlled by `BusinessArea.ingest_source attribute.` (`src/hope/models/business_area.py`)

#### Guards

HOPE checks whether RDI being uploaded is a permitted action by using `BusinessAreaIngestCWOnlyPermission` and `BusinessAreaIngestAllExceptCWPermission` permissions (`src/hope/api/auth.py`). Guards examples:
1. As of 25 August 2026 All endpoints implemented in `src/hope/api/endpoints/rdi/base.py` should only be used by Country Workspace, that's why HOPE checks it.
2. Uploading RDI through generic import should not trigger new automerge flow.

Once BA starts using Country Workspace for uploading RDIs they can't go back to legacy flow (see BusinessArea.clean method). 

## What got deleted
1. `services/biometric_deduplication.py` - biometric deduplication is driven by Country Workspace. No more need to keep utils for HOPE to do that.
2. `api/deduplication_engine.py` - all methods were removed except the one for getting findings from DE.
3. `deduplication_engine_status` - no more need to track the DE statuses. Every RDI coming from Country Workspace IS ALREADY deduplicated and findings are waiting to be fetched

## The merge queue

Below diagram depicts how rdi `tasks/rdi_merge_dispatcher.py`. It triggers the merge RDI flow
```mermaid
  flowchart TD
      T1[CW completes RDI POST] --> A
      T2[Admin retry button] --> A
      A[RDI Dispatcher<br/>celery task] --> B{Oldest RDI<br/>for program exists?}
      B -->|Queue empty| D[FINISH]
      B -->|Oldest in MERGE_ERROR| P[FINISH — queue paused]
      B -->|Oldest in MERGING| W[FINISH — worker active]
      B -->|Oldest in MERGE_SCHEDULED| C[Acquire program lock<br/>select_for_update → set MERGING]
      C --> E[AUTOMERGE<br/>fetch_findings_and_merge_rdi]
      E --> F{Success?}
      F -->|NO| G[remove ES docs<br/>set MERGE_ERROR] --> P
      F -->|YES| A
```
If given RDI is left in MERGE_ERROR (e. g. findings could not be fetched from Deduplication Engine), admin user is able to retrigger the dispatcher:

```mermaid
  flowchart TD
      A[Admin opens Programme] --> B{MERGE_ERROR RDI<br/>exists for programme?}
      B -->|No| C[Button disabled]
      B -->|Yes| D[Button enabled]
      D --> E[Click: Retry merge queue]
      E --> F[Enqueue dispatcher]
```

## UNDO: the REST endpoint
```mermaid
sequenceDiagram
    autonumber
    participant CW as Country Workspace
    participant H as HOPE API
    participant W as HOPE Celery worker

    CW->>H: POST /api/rest/BA/rdi/create/
    H-->>CW: 201 Created - RDI A, country_workspace_id X

    CW->>H: POST /rdi/A/push/lax/individuals/
    H--xCW: connection lost - push aborts
    Note over H: RDI A stuck LOADING, 0 ind / 0 hh, RDI.country_workspace_id has unique constraint

    CW->>H: POST /rdi/A/reset/ with callback_url and signed_token
    H->>H: row lock, status DELETE_SCHEDULED, enqueue wipe, kick dispatcher
    H-->>CW: 202 Accepted

    H->>W: wipe job
    W->>W: null head_of_household, rdi.delete cascade, invalidate cache, drop ES docs
    Note over W: RDI A row gone - country_workspace_id X is free

    W->>CW: POST /api/rest/callbacks/hope/rdps/push-ready/ with signed_token

    CW->>H: POST /api/rest/BA/rdi/create/
    H-->>CW: 201 Created - RDI B, same country_workspace_id X
    CW->>H: POST /rdi/B/push/lax/individuals/
    H-->>CW: 201 Created
    CW->>H: POST /rdi/B/push/lax/households/
    H-->>CW: 201 Created
    CW->>H: POST /rdi/B/completed/
    H-->>CW: 200 OK

    H->>W: merge queue - automerge
    W->>W: fetch_findings_and_merge_rdi
    Note over W: RDI B - MERGED
```

src/hope/apps/registration_data/tasks/rdi_merge_dispatcher.py
```diff
@@ -0,0 +1,44 @@
...
+        RegistrationDataImport.MERGING,
+        RegistrationDataImport.MERGE_ERROR,
+        RegistrationDataImport.IMPORT_ERROR,
+    ]
+
+    def execute(self, program_id: str) -> None:
+        from hope.apps.registration_data.celery_tasks import fetch_findings_and_merge_rdi
+
+        head = (
+            RegistrationDataImport.objects.filter(program_id=program_id, status__in=self._QUEUE_STATUSES)
```
johniak: Is CW-only guaranteed to be enabled only for business areas with no prior XLSX/Kobo imports? If an old RDI without `country_workspace_id` can sit here in MERGE_SCHEDULED/MERGE_ERROR/IMPORT_ERROR, it becomes the head and blocks the queue forever — `fetch_findings_and_merge_rdi` refuses it and nothing re-dispatches. If that history can exist, add `country_workspace_id__isnull=False` here and in `ProgramAdmin.retry_cw_merge_queue`.

src/hope/api/endpoints/rdi/push_people.py
```diff
@@ -59,6 +60,31 @@ class DynamicAreaChoiceField(serializers.ChoiceField):
     pass
 
 
+class PushPeopleListSerializer(serializers.ListSerializer):
+    def validate(self, attrs: list[dict]) -> list[dict]:
+        cw_ids = [item["country_workspace_id"] for item in attrs if item.get("country_workspace_id")]
+        existing_cw_ids = sorted(
+            Individual.all_objects.filter(
+                is_removed=False,
+                withdrawn=False,
```
johniak: Skipping withdrawn (and removed) rows here lets CW reuse a cw id in the same program, but `BiometricDedupeSimilarityPair._resolve_id_to_hope_pk` maps cw id -> pk with an unordered dict over `Individual.all_objects`, so a finding can bind to the withdrawn twin instead of the new pending individual. Such pairs don't intersect the RDI's pending individuals, so stats and NA-ticket creation silently drop them and the automerge merges an engine-flagged duplicate with no ticket. Either exclude withdrawn/removed rows in `_resolve_id_to_hope_pk` (preferring the current RDI's pending row) or don't allow the reuse.

src/hope/apps/program/migrations/0021_disable_biometric_dedup_for_non_cw.py
```diff
@@ -0,0 +1,37 @@
...
+    For a business area that does not ingest from CW nothing writes
+    ``dedup_engine_batch_duplicates`` / ``dedup_engine_golden_record_duplicates`` any more, so the
+    counters stay at their default of 0 and the RDI details page renders them as "100% unique"
+    (``number_of_individuals - 0``) — a scan that never ran, shown as a scan that found nothing.
+    Turning the flag off makes the serializer omit the biometric column instead of computing it.
+    """
+    Program = apps.get_model("program", "Program")
+    updated = Program.objects.filter(
+        biometric_deduplication_enabled=True,
+        business_area__ingest_source=ALL_EXCEPT_COUNTRY_WORKSPACE,
```
johniak, line 22: At the moment this runs, every BA still has the fresh default ALL_EXCEPT_COUNTRY_WORKSPACE, so this disables `biometric_deduplication_enabled` on every program that has it on — including the pilot CW business area. After deploy someone must set `ingest_source` on the pilot BA and re-enable the flag per program (in that order, since the admin makes the flag readonly for non-CW BAs). Is that runbook written down, or should the migration skip BAs known to be on the CW pilot?

src/hope/api/endpoints/rdi/push_people.py
```diff
@@ -59,6 +60,31 @@ class DynamicAreaChoiceField(serializers.ChoiceField):
     pass
 
 
+class PushPeopleListSerializer(serializers.ListSerializer):
+    def validate(self, attrs: list[dict]) -> list[dict]:
```
johniak, line 64: This only checks against the DB — duplicates within one payload pass, create N pending individuals sharing a cw id, and both the `individual_id_mapping` response and later findings resolution collapse them silently (dict last-wins). The deleted skip-marked test even pinned the old behavior: it asserted "Duplicate country_workspace_id values in payload: ...". Add the in-payload uniqueness check back here (a `Counter` over `attrs`).

src/hope/api/endpoints/rdi/lax.py
```diff
@@ -153,10 +153,9 @@ class IndividualSerializer(serializers.ModelSerializer):
     accounts = AccountLaxSerializer(many=True, required=False)
     photo = serializers.CharField(allow_blank=True, required=False)
     disability_certificate_picture = serializers.CharField(allow_null=True, allow_blank=True, required=False)
-    individual_id = serializers.CharField(required=True)
     disability = DisabilityChoiceField(choices=DISABILITY_CHOICES, required=False, allow_blank=True)
     sex = serializers.ChoiceField(SEX_CHOICE, allow_blank=False, default=NOT_COLLECTED)
-    country_workspace_id = serializers.CharField(required=False, allow_blank=True, allow_null=True, max_length=150)
+    country_workspace_id = serializers.CharField(required=True, max_length=150)
```
johniak, line 166: The lax path (the main CW household flow) got no equivalent of `PushPeopleListSerializer`'s already-exists check — the same problem is solved in one sibling endpoint and ignored in the other. A lax re-push after a partial failure (without going through reset) creates BA-wide cw-id duplicates that the findings mapping then resolves arbitrarily. Either add the same list-level check here or state explicitly why lax is exempt.

src/hope/models/business_area.py
```diff
@@ -103,8 +122,21 @@ class BusinessArea(NaturalKeyModel, TimeStampedUUIDModel):
...
+            new_ingest_source == BusinessArea.IngestSource.ALL_EXCEPT_COUNTRY_WORKSPACE
+            and old_ingest_source == BusinessArea.IngestSource.COUNTRY_WORKSPACE_ONLY
+        ):
+            raise ValidationError(
+                {"ingest_source": _("Cannot go back to legacy ingest sources when CW source was chosen.")}
+            )
+
     def save(self, *args: Any, **kwargs: Any) -> None:
         unique_slugify(self, self.name, slug_field_name="slug")
+        self.full_clean()
```
johniak: `full_clean()` inside `save()` changes the contract of every programmatic BusinessArea save: `loadbusinessareas`, the split-office admin flow, `self.parent.save()` two lines below, and any task/script save now run full field+unique validation plus the extra SELECT from `clean()`. Any legacy row with a value that no longer passes a validator becomes unsavable, and the admin already validates via the form. The repo convention is explicit `full_clean()` at call sites — if the goal is just to enforce the one-way `ingest_source` outside the admin, check that single condition in `save()` instead.

src/hope/admin/registration_data.py
```diff
@@ -158,41 +178,28 @@ def rerun_merge_rdi(self, request: HttpRequest, pk: str) -> HttpResponse | None:
...
         try:
-            fetch_biometric_deduplication_results_and_process_async_task(str(rdi.program_id), str(rdi.id))
-        except Exception as e:  # noqa: BLE001
+            rdi = self.get_object(request, pk)
+            if rdi is None:
+                self.message_user(request, "Registration Data Import not found", messages.ERROR)
+                return None
+            from hope.apps.registration_data.celery_tasks import fetch_findings_and_merge_rdi
+
+            fetch_findings_and_merge_rdi(rdi)
```
johniak: This calls `fetch_findings_and_merge_rdi(rdi)` directly, bypassing the dispatcher's oldest-first invariant — and the button is enabled for MERGE_SCHEDULED too, so an admin can merge RDI #3 while #1 waits at head (golden-record stats depend on earlier RDIs merging first). The `ProgramAdmin` retry added in this same PR routes through `rdi_dispatcher_task`; this one should too (set the RDI back to MERGE_SCHEDULED if failed, then dispatch).

src/hope/apps/core/management/commands/dedup_teardown.py
```diff
@@ -0,0 +1,122 @@
...
+        # Bottom-up hard delete on BaseManager (all_objects) querysets.
+        # 1) break the Household <-> Individual PROTECT cycle via the nullable side.
+        Individual.all_objects.filter(business_area_id__in=ba_ids).update(household=None)
+        # 2) leaves -> roots.
+        Household.all_objects.filter(business_area_id__in=ba_ids).delete()
+        Individual.all_objects.filter(business_area_id__in=ba_ids).delete()
+        RegistrationDataImport.objects.filter(business_area_id__in=ba_ids).delete()
+        ProgramCycle.objects.filter(program__business_area_id__in=ba_ids).delete()
+        Program.all_objects.filter(business_area_id__in=ba_ids).delete()
+        BusinessArea.objects.filter(id__in=ba_ids).delete()
```
johniak, line 116: Teardown removes the BAs but never deletes the all-grants `APIToken` that `dedup_setup` minted for root (and printed to stdout). The `valid_for` M2M rows cascade away, but `HOPEPermission` only enforces `valid_for` on views with a `business_area_slug` kwarg — the leftover token still authenticates on non-BA-scoped token endpoints for its full 1-year validity. Stamp the harness tokens (e.g. with the run tag) and delete them here.

src/hope/api/auth.py
```diff
@@ -48,3 +49,19 @@ def has_permission(self, request: Request, view: Any) -> bool:
         if ba_slug and not request.auth.valid_for.filter(slug=ba_slug).exists():
             raise Http404
         return True
+
+
+class BusinessAreaIngestCWOnlyPermission(BasePermission):
+    message = CW_ONLY_INGEST_REJECT_MSG
+
+    def has_permission(self, request: Request, view: Any) -> bool:
+        return view.selected_business_area.is_rdi_ingest_source_country_workspace_only
```
johniak: The two new permission classes read different view attributes for the same concept: this one uses `view.selected_business_area`, `BusinessAreaIngestAllExceptCWPermission` uses `view.business_area`. Attach either to the other view family and it AttributeErrors at request time. A shared helper (or `getattr(view, "selected_business_area", None) or view.business_area`) removes the trap.

src/hope/admin/record.py
```diff
@@ -100,6 +101,11 @@ def clean_filters(self) -> tuple[dict[str, Any], dict[str, Any]]:
 
     def clean(self) -> None:
         super().clean()
+        registration = self.cleaned_data.get("registration")
+        if registration:
+            business_area = BusinessArea.objects.get(slug=registration.project.organization.slug)
```
johniak: A bare `.get()` here raises an uncaught `DoesNotExist` -> 500 in the admin form when the aurora organization slug doesn't match a BA. Use `.filter(...).first()` and skip the check (or raise a form error) when it's missing.

johniak (changes requested):

src/hope/api/endpoints/rdi/push_people.py
Jeeziorny: Resolved: https://github.com/unicef/hope/pull/6256/changes/67b636dc993523def2eea0e75ffbb5e7d9063209

src/hope/apps/program/migrations/0021_disable_biometric_dedup_for_non_cw.py
Jeeziorny: We agreed this feature to **not to be** backward compatible. I thought that it's ok to release this and say: from now on all of you DON'T have biometric deduplication. If you want it, you need to come to us and tell explicitly: "We'd like to have biometric deduplication (and thus use only Country Workspace for uploading RDIs). 

I'm not aware of any BAs that should be forced to have biometric deduplication and use Country Workspace as the system for RDIs right after this deployment, but it's probably more of a business decision that needs at least confirmation.

@ntrncic what do we do?

src/hope/apps/registration_data/tasks/rdi_merge_dispatcher.py
Jeeziorny: https://github.com/unicef/hope/pull/6256/changes/9112efcc9663d9f808be0a0a18263d9fabdd4b24

src/hope/api/endpoints/rdi/push_people.py
Jeeziorny: https://github.com/unicef/hope/pull/6256/changes/0435724b196d49904ce87d56aa68e78c1f0eb972

src/hope/apps/grievance/services/needs_adjudication_ticket_services.py
Jeeziorny, line 377: 1. Let's assume we have 10 biometric findings.
2. On this line, every ticket is assigned with 9 tickets (`ALL_TICKETS - ITSELF`). To save it, we're adding 10*9 rows in M2M table.
3. linked_tickets has `symmetrical=True`, so for EACH row we need to double it: now we need to add 10*9*2 rows.
4. Before postgres save the row, it will go and see if those linked tickets exist. Having pair (TICKET A, TICKET B) it needs to check whether TICKET A exists and if TICKET B exists. That's again *2, now we have 10 * 9 * 2 * 2 queries to be made for 10 biometric findings.

The problem wasn't really a problem before. Now I'm using `select_for_update` in `fetch_findings_and_merge` which can lock an RDI for some time.. Am I dramatising or should I do something about this part? @MarekBiczysko

johniak (dismissed):

src/hope/models/business_area.py
Jeeziorny: https://github.com/unicef/hope/pull/6256/changes/41aedb01999b24a1ad5ff029a302d5cb9527dc7d

src/hope/api/endpoints/rdi/lax.py
Jeeziorny: https://github.com/unicef/hope/pull/6256/changes/c747bf734bc0d35b84abce870998815aa8334fa1

src/hope/admin/registration_data.py
Jeeziorny: Good catch. It was leftover from first implementation, the only way to retry is to rerun the whole queue per program. https://github.com/unicef/hope/pull/6256/changes/5abab72a8d800f1d7542cc2f2fb4e69c2e5a86d2

src/hope/admin/record.py
Jeeziorny: https://github.com/unicef/hope/pull/6256/changes/5abab72a8d800f1d7542cc2f2fb4e69c2e5a86d2

src/hope/api/auth.py
Jeeziorny: https://github.com/unicef/hope/pull/6256/changes/eb0fccb183950bfa92ffaa18ebab0abc0b5583a3

src/hope/apps/grievance/services/needs_adjudication_ticket_services.py
MarekBiczysko: • For 10 tickets with the current code:

  - 10 Python .set() calls
  - Approximately 46 SQL statements
      - 10 queries reading existing links
      - 18 queries checking missing links
      - 18 bulk INSERT statements

  - 90 rows inserted into the M2M table
  - 180 internal FK checks by PostgreSQL, but these are not separate database calls
  
  I guess we can leave as is, you can create separate optimisation ticket to handle this later if needed

MarekBiczysko (dismissed):

src/hope/api/endpoints/rdi/upload.py
```diff
@@ -160,6 +162,7 @@ class Meta:
             "vector_column",
             "unicef_id",
             "program",
+            "country_workspace_id",
```
johniak, line 167: Excluding `country_workspace_id` here also strips it from `HouseholdSerializer.members`, which `PushToRDIView` and `PushLaxToRDIView` (base.py:106/140) consume — both are CW-only endpoints now. DRF silently ignores unknown keys, so CW household pushes create members with NULL cw id, findings can't bind to them, and automerge proceeds without NA tickets. Either keep the field writable for the members serializer or make those endpoints reject it explicitly.

johniak (changes requested):

src/hope/api/endpoints/rdi/upload.py
Jeeziorny: https://github.com/unicef/hope/pull/6256/changes/16ce79b70463e58d927eb4c42ab4aca1a9bb6478 (added unique constraint in model too)

MarekBiczysko (dismissed):

johniak (dismissed):

johniak (dismissed):

src/hope/apps/program/migrations/0021_disable_biometric_dedup_for_non_cw.py
ntrncic: Yes all BAs will have to use country workspace if they want biometric deduplication. But the programmes and did this should remain.

MarekBiczysko (dismissed):

src/hope/models/individual.py
```diff
@@ -597,6 +597,11 @@ class Meta:
                 condition=Q(is_removed=False) & Q(originating_id__isnull=False),
                 name="originating_id_ind_unique_constraint",
             ),
+            UniqueConstraint(
+                fields=["business_area", "country_workspace_id"],
```
johniak, line 653: We have features that copy population from one program into another inside the same business area: RDI "import from program population" and whole-program copy (both via `CopyProgramPopulation.copy_individual`), plus admin "enroll households to program" (`program.utils.copy_individual`). All of them keep `country_workspace_id` on the copy, so with this constraint they will necessarily blow up: two live individuals end up with the same `(business_area, country_workspace_id)` pair (reproduced on both helpers). Should we clear `country_workspace_id` on the copy, or rather add `program` to the constraint?

src/hope/api/endpoints/rdi/upload.py
```diff
@@ -242,6 +246,31 @@ def validate(self, attrs: dict) -> dict:
...
+
+class CountryWorkspaceHouseholdSerializer(HouseholdSerializer):
+    members = CountryWorkspaceIndividualSerializer(many=True, required=True)
+
+    def validate_members(self, value: list[dict]) -> list[dict]:
+        existing = self.context.get("existing_cw_ids", set())
+        duplicated = self.context.get("duplicated_cw_ids", set())
+        errors = [cw_id_error(member.get("country_workspace_id"), existing, duplicated) for member in value]
+        if any(errors):
+            raise ValidationError(errors)
```
johniak: `errors` holds `None` for members whose id is fine, and DRF renders `None` as the string `"None"`, so the 400 body reads `{"members": {"Member #1": ["None"], "Member #2": [{"country_workspace_id": [...]}]}}`. Use `{}` for the clean slots (`[e or {} for e in errors]`); DRF and `humanize_errors` both drop empty dicts.

src/hope/models/business_area.py
```diff
@@ -103,8 +122,21 @@ class BusinessArea(NaturalKeyModel, TimeStampedUUIDModel):
...
 
+    def clean(self) -> None:
+        super().clean()
+        new_ingest_source = self.ingest_source
+        old_ingest_source = BusinessArea.objects.filter(pk=self.pk).values_list("ingest_source", flat=True).first()
+        if (
+            new_ingest_source == BusinessArea.IngestSource.ALL_EXCEPT_COUNTRY_WORKSPACE
+            and old_ingest_source == BusinessArea.IngestSource.COUNTRY_WORKSPACE_ONLY
+        ):
+            raise ValidationError(
```
johniak, line 140: This only blocks going back to legacy. Switching a BA to `COUNTRY_WORKSPACE_ONLY` while a legacy RDI is still in IN_REVIEW / IMPORTING / DEDUPLICATION strands it: the SPA hides Merge/Refuse behind `isManualIngest`, `BusinessAreaIngestAllExceptCWPermission` rejects the POST, the admin retry buttons are disabled, and the dispatcher only picks rows with a cw id. Should the switch be rejected while such RDIs exist, or how are they meant to be finished?

src/hope/api/endpoints/rdi/cw_ids.py
```diff
@@ -0,0 +1,84 @@
...
+        queryset = queryset.exclude(originating_id__in=exclude_originating_ids)
+    return set(queryset.values_list("country_workspace_id", flat=True))
+
+
+def collect_cw_ids(request_data: Iterable[object]) -> list[str]:
+    """Country workspace ids of a flat individual payload."""
+    return [
+        row["country_workspace_id"]
+        for row in request_data
+        if isinstance(row, dict) and isinstance(row.get("country_workspace_id"), str) and row["country_workspace_id"]
```
johniak: Only `str` values are collected (same in `collect_member_cw_ids`). An integer cw id skips both the existing and in-payload checks, DRF's `CharField` coerces it to a string, and the new DB constraint fires instead: unhandled `IntegrityError` → 500 on `rdi-push`, and `"Failed to create lax individuals"` on the lax endpoint. `str()` non-empty scalars here, or reject non-string ids up front.

src/hope/apps/household/migrations/0048_migration.py
```diff
@@ -0,0 +1,27 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0033_migration"),
+        ("household", "0047_migration"),
+        ("program", "0021_disable_biometric_dedup_for_non_cw"),
+        ("registration_data", "0017_migration"),
+    ]
+
+    operations = [
+        migrations.AddConstraint(
```
johniak: Let's make sure before deploy that there's nothing on prod that breaks this constraint and blows up the migration and the whole deployment. cw id has been there for months with no uniqueness check, so two live individuals with the same id in one BA is totally possible. This is really important, please check it.

src/hope/apps/registration_data/celery_tasks.py
```diff
@@ -490,90 +485,163 @@ def deduplicate_documents_for_rdi(rdi_id: str) -> bool:
...
+    try:
+        with locked_cache(key=f"fetch_findings_and_merge_rdi-program-{program_id}") as locked:
+            if not locked:
+                logger.info(
+                    f"RDI:{registration_data_import_id} fetch_findings_and_merge_rdi skipped (program lock held)"
+                )
+                return True
+            merged = FetchFindingsAndMergeRdi().execute(registration_data_import_id)
+    except Exception as exc:  # noqa
+        if merge_still_owns_rdi(registration_data_import_id):
```
johniak: `merge_still_owns_rdi` is check-then-act: between its `exists()` and `handle_rdi_exception`'s `get()` + `save()` a CW reset can still flip the row to DELETE_SCHEDULED, and MERGE_ERROR overwrites it — the case the docstring wants to prevent. A filtered `.update(status=MERGE_ERROR)` excluding those statuses (as `RdiRemovalTask.mark_failed` does) closes the window.

src/frontend/src/components/core/Drawer/DrawerItems.tsx
```diff
@@ -169,8 +169,10 @@ export const DrawerItems = ({
   );
 
   const { isAccountabilityApplicable } = businessAreaData;
+  const isManualIngest = businessAreaData.isManualIngest;
   const flags = {
     isAccountabilityApplicable,
+    isManualIngest,
```
johniak: `isManualIngest` is added to `flags`, but no entry in `menuItems.tsx` declares `flag: 'isManualIngest'`, so it's dead. Drop it or wire the RDI item to it.

src/frontend/src/containers/pages/rdi/RegistrationDataImportPage.tsx
```diff
@@ -70,66 +56,27 @@ function RegistrationDataImportPage(): ReactElement {
...
-  if (permissions === null || loading) return null;
+  if (permissions === null) return null;
 
   if (!hasPermissions(PERMISSIONS.RDI_VIEW_LIST, permissions))
     return <PermissionDenied permission={PERMISSIONS.RDI_VIEW_LIST} />;
 
   const toolbar = (
     <PageHeader title={t('Registration Data Import')}>
-      <Box
+            <Box
```
johniak: Run prettier on this one, mate.

johniak (changes requested):

src/hope/models/individual.py
Jeeziorny: As agreed on 3.09.2026 daily with @ntrncic when we copy programme P (and have P'), the result of this copy (P') has all CW IDs set to NONE. This needs separate PR which I'll open in a moment.

Jeeziorny: https://github.com/unicef/hope/pull/6374

Leaving constraint (BA, CWID)

src/hope/models/business_area.py
Jeeziorny: Waiting for @ntrncic if we're able to ask BAs with biometric DD enabled to merge all pending RDIs and not create new ones before release

johniak (dismissed):

johniak (dismissed):

tests/unit/apps/registration_data/test_fetch_findings_and_merge_rdi.py
```diff
@@ -0,0 +1,1271 @@
...
+    cw_rdi.refresh_from_db()
+    assert cw_rdi.status == RegistrationDataImport.MERGE_ERROR
+    assert cw_rdi.error_message == "merge boom"
+
+
+@pytest.mark.skip(
+    reason="N1 — E2E/ES: needs real Elasticsearch (make services); the CW happy path with full "
+    "merge side effects is exercised by tests/unit/api/test_cw_dedup_flow_e2e.py"
+)
+def test_fetch_and_merge_happy_path_full_side_effects() -> None:
```
johniak: Three empty `@pytest.mark.skip` placeholders here and at L991/L1187. Their reasons cite a "(not-yet-built) Problem 1 dispatcher" (it exists: `rdi_merge_dispatcher.py`) and a `PROBLEM0_TRANSACTION_TESTS.md` that isn't in the repo. Delete them; the skip reason itself says the happy path lives in `test_cw_dedup_flow_e2e.py`.

tests/unit/apps/registration_data/test_fetch_findings_and_merge_rdi.py
```diff
@@ -0,0 +1,1271 @@
...
+    queue_and_run_retry_task(fetch_findings_and_merge_rdi, registration_data_import=cw_rdi)
+
+    cw_rdi.refresh_from_db()
+    assert cw_rdi.status == RegistrationDataImport.MERGED
+    mock_rdi_merge.return_value.execute.assert_not_called()
+    assert BiometricDedupeSimilarityPair.objects.filter(program=cw_rdi.program).count() == 0
+
+
+# ---------------------------------------------------------------------------
+# Problem 0 — Part 2: transaction rework.
```
johniak: This "Problem 0 — Part 2: transaction rework" banner and the `N1`…`N8` labels in the comments below describe the history of the diff, not what the tests assert. Drop the banner and the labels.

src/hope/admin/program.py
```diff
@@ -266,6 +246,53 @@ def get_queryset(self, request: HttpRequest) -> QuerySet[Program]:
     def frontend_url(self, obj: Program) -> str | None:
         return f"/{obj.business_area.slug}/programs/{obj.code}/details/{obj.code}"
 
+    def get_readonly_fields(self, request: HttpRequest, obj: Program | None = None) -> tuple[str, ...]:
+        readonly_fields = tuple(super().get_readonly_fields(request, obj))
+        if obj and obj.business_area.is_rdi_ingest_source_all_except_country_workspace:
```
johniak, line 251: Read-only only covers the change view. The add view still accepts `biometric_deduplication_enabled=True` for a programme in an `ALL_EXCEPT_COUNTRY_WORKSPACE` BA, and `Program.clean()` has no rule for it, so an admin can recreate exactly the state migration 0021 cleans up. A check in `ProgramAdminForm.clean` (or `Program.clean`) covers both views.

johniak (changes requested):

johniak (dismissed):

johniak (dismissed):

domdinicola (dismissed):

src/hope/apps/core/management/commands/dedup_setup.py
```diff
@@ -0,0 +1,361 @@
+"""dedup_setup — build the dedup-regression test substrate.
```
domdinicola, line 1: maybe we add this to a separate up for dev, or in extra

src/hope/apps/core/management/commands/dedup_teardown.py
```diff
@@ -0,0 +1,130 @@
+"""dedup_teardown — remove everything dedup_setup created.
+
```
domdinicola, line 2: same as above

src/hope/models/business_area.py
```diff
@@ -145,6 +177,14 @@ def __str__(self) -> str:
     def natural_key(self) -> tuple[str]:
         return (self.code,)
 
+    @property
+    def is_rdi_ingest_source_country_workspace_only(self) -> bool:
+        return self.ingest_source == self.IngestSource.COUNTRY_WORKSPACE_ONLY
+
+    @property
```
domdinicola, line 184: same here

src/hope/models/business_area.py
```diff
@@ -145,6 +177,14 @@ def __str__(self) -> str:
     def natural_key(self) -> tuple[str]:
         return (self.code,)
 
+    @property
+    def is_rdi_ingest_source_country_workspace_only(self) -> bool:
```
domdinicola, line 181: this necessary?

src/hope/models/registration_data_import.py
```diff
@@ -270,24 +245,9 @@ def refresh_population_statistics(self) -> None:
...
-            return False
-
-        is_still_processing = RegistrationDataImport.objects.filter(
-            program=self.program,
-            deduplication_engine_status__in=[
-                RegistrationDataImport.DEDUP_ENGINE_IN_PROGRESS,
-            ],
-        ).exists()
-        return not is_still_processing
+        return self.program.is_active()
```
domdinicola, line 250: needed?

src/hope/apps/core/management/commands/dedup_setup.py
Jeeziorny: We don't have separate app for dev. initdemo is in core as well. I like your idea about splitting dev only and prod commands, but I feel like it's out of scope of this PR

src/hope/apps/core/management/commands/dedup_teardown.py
Jeeziorny: SAA

src/hope/models/business_area.py
Jeeziorny: Used in ~10 places, it earned its own property IMO

src/hope/models/registration_data_import.py
Jeeziorny: used by FE

MarekBiczysko (dismissed):

PR #6258, opened 2026-07-28
330287_It_should_be_possible_to_create_PaymentPlan_flex_fields_even_b…

MarekBiczysko:
…efore_reconcilitaion

[AB#330287](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/330287)

## Summary

Adds Payment Plan-level XLSX import/export for FSP-required Payment Record fields.

## Key changes

- Available while the Payment Plan is LOCKED_FSP.
- Template contains payment_id and only existing FSP extra-field columns.
- Imported non-empty cells create or overwrite values; empty cells do not delete existing values.
- Stores fields in the existing Payment.extras JSON:

{
"extra_fields": {}, # reconciliation import extra fields
"fsp_extra_fields": {}
}

- Includes FSP extra fields in manual FSP XLSX exports and Payment Gateway payloads.
- System-generated Payment Gateway values take precedence on name collisions.
- Displays separate Extra fields and FSP extra fields sections in Payment Details.
- Import runs asynchronously with background-action validation and retry support.

## Deployment note

After deploying migration 0073, run:

from hope.one_time_scripts.migrate_payment_extras import migrate_payment_extras
migrate_payment_extras()

This moves existing flat production Payment.extras values under 'extra_fields' key.


src/hope/apps/payment/xlsx/xlsx_payment_plan_fsp_extra_fields_import_service.py
```diff
@@ -0,0 +1,199 @@
...
+                continue
+            seen_payment_ids.add(payment_id)
+            updates = self._fsp_extra_fields_for_row(row)
+            if not updates or not any(payment.fsp_extra_fields.get(key) != value for key, value in updates.items()):
+                continue
+            old_payment = cast("Payment", copy_model_object(payment))
+            payment.extras[Payment.FSP_EXTRA_FIELDS_KEY] = {
+                **payment.fsp_extra_fields,
+                **updates,
+            }
```
pkujawa: -> ` payment.extras = {**payment.extras, Payment.FSP_EXTRA_FIELDS_KEY: {**payment.fsp_extra_fields, **updates}} `              
                                                                                                                                                                                                  
after `old_payment = cast("Payment", copy_model_object(payment))`, the `payment` and `old_payment` share the same dict for `extras`, so this line would modify both. And `bulk_log_payment_changes(log_pairs, user)` wouldn't log any change for `fsp_extra_fields`

src/hope/apps/payment/xlsx/xlsx_payment_plan_fsp_extra_fields_import_service.py
```diff
@@ -0,0 +1,199 @@
...
+            log_pairs.append((old_payment, payment))
+
+        if payments_to_update:
+            Payment.objects.bulk_update(
+                payments_to_update,
+                fields=("extras", "signature_hash"),
+                batch_size=self.BATCH_SIZE,
+            )
+            user = User.objects.filter(pk=user_id).first() if user_id else None
+            bulk_log_payment_changes(log_pairs, user)
```
pkujawa, line 178: Actually could you add test checking the LogEntry being created and the diff?

src/frontend/src/components/paymentmodule/PaymentDetails/PaymentDetails.tsx
```diff
@@ -285,25 +285,36 @@ function PaymentDetails({
...
-                  value={safeStringify(value)}
-                />
-              </Grid>
-            ))}
-          </Grid>
-        </Overview>
-      )}
+      <Overview>
+        <Title>
+          <Typography variant="h6">{t('Extra fields')}</Typography>
```
MarekBiczysko:
```suggestion
          <Typography variant="h6">{t('Reconciliation Information: Extra Info')}</Typography>
```

src/hope/one_time_scripts/migrate_payment_extras.py
```diff
@@ -0,0 +1,134 @@
...
+from psycopg2 import sql
+
+from hope.models import Payment
+
+ADVISORY_LOCK_NAME = "hope.migrate_payment_extras"
+DEFAULT_BLOCKS_PER_BATCH = 10_000
+DEFAULT_LOCK_TIMEOUT_MS = 5_000
+LEGACY_EXTRAS_PREDICATE = """
+extras <> '{}'::jsonb
+AND NOT (extras ? 'extra_fields' OR extras ? 'fsp_extra_fields')
```
johniak: This predicate skips rows that already contain one of the namespaced keys, but both new import paths create exactly that mixed shape on legacy rows: `set_extra_fields()` and the FSP import only set their own key, keeping the flat legacy keys as siblings. Any import run between deploy and this script turns a flat row into a mixed one that gets skipped forever — and `_get_legacy_extras_count` uses the same predicate, so it isn't even counted as remaining. Detect legacy rows by "has any top-level key other than the two namespaces" and make the UPDATE merge the flat keys into `extra_fields` instead of skipping.

src/hope/models/payment.py
```diff
@@ -285,8 +289,23 @@ class Meta:
         "delivered_quantity_usd",
         "delivery_date",
         "transaction_reference_id",
+        "fsp_extra_fields",
```
johniak, line 292: `update_signature_hash` hashes this via `str(value)`, which is key-order sensitive. The import computes the hash on an in-memory dict (`{**payment.fsp_extra_fields, **updates}`, insertion order), but after the JSONB round-trip Postgres returns keys in its own sorted order — so the next `recalculate_signatures_in_batch()` (which runs after every reconciliation import) rewrites a different hash for identical data. Serialize dict values deterministically, e.g. `json.dumps(value, sort_keys=True, default=str)` in `_normalize`.

src/frontend/src/components/paymentmodule/PaymentDetails/PaymentDetails.tsx
```diff
@@ -285,25 +285,36 @@ function PaymentDetails({
...
-                <LabelizedField
-                  label={formatNormalCaseValue(key)}
-                  value={safeStringify(value)}
-                />
-              </Grid>
-            ))}
-          </Grid>
-        </Overview>
-      )}
+      <Overview>
```
johniak: Dropping the emptiness guard means every payment detail page now shows two boxes that are usually empty — just a heading with nothing under it, which reads like a rendering bug. If always showing both boxes is intentional per the ticket, consider at least an empty-state placeholder; otherwise keep the sections conditional.

src/hope/apps/payment/api/views.py
```diff
@@ -1174,6 +1186,81 @@ def entitlement_import_xlsx(self, request: Request, *args: Any, **kwargs: Any) -
...
+        responses={200: PaymentPlanDetailSerializer, 400: XlsxErrorSerializer},
+    )
+    @action(
+        detail=True,
+        methods=["post"],
+        url_path="fsp-extra-fields-import-xlsx",
+        parser_classes=[DictDrfNestedParser],
+    )
+    @transaction.atomic
+    def fsp_extra_fields_import_xlsx(self, request: Request, *args: Any, **kwargs: Any) -> Response:
```
johniak, line 1258: Unlike `entitlement_import_xlsx`, which writes a `log_create` for the plan when queuing the import, this action logs nothing at dispatch time and relies on the task's `log_payment_plan_change` afterwards. Probably fine, but it's a silent deviation from the direct sibling — either add the dispatch log or leave a note why it's omitted.

src/hope/apps/payment/xlsx/xlsx_payment_plan_fsp_extra_fields_import_service.py
```diff
@@ -0,0 +1,199 @@
...
+            payment_id = str(payment_id_value) if payment_id_value not in (None, "") else ""
+            payment = self.payments.get(payment_id)
+            if payment is None or payment_id in seen_payment_ids:
+                continue
+            seen_payment_ids.add(payment_id)
+            updates = self._fsp_extra_fields_for_row(row)
+            if not updates or not any(payment.fsp_extra_fields.get(key) != value for key, value in updates.items()):
+                continue
+            old_payment = cast("Payment", copy_model_object(payment))
+            payment.extras[Payment.FSP_EXTRA_FIELDS_KEY] = {
```
johniak: The model grew `set_extra_fields()` for the reconciliation namespace, but this path pokes `extras[FSP_EXTRA_FIELDS_KEY]` directly. When you apply the rebinding fix from the earlier thread, consider adding a symmetric `set_fsp_extra_fields()` so both namespaces are written the same way.

johniak (changes requested):

johniak (changes requested):
According to last discussion about e2e test. There was a request to test everything which touches frontend with e2e (modify add step to existing test or add test if necessary)

johniak (dismissed):

src/hope/apps/payment/xlsx/xlsx_payment_plan_fsp_extra_fields_import_service.py
```diff
@@ -0,0 +1,201 @@
...
+            if value in (None, ""):
+                continue
+            values[header] = self._normalize_value(value)
+        return values
+
+    @staticmethod
+    def _normalize_value(value: Any) -> object:
+        if isinstance(value, Decimal):
+            return float(value)
+        if isinstance(value, datetime.datetime | datetime.date | datetime.time):
```
ntrncic, line 197: should update this to isinstance(value, (datetime.datetime, datetime.date, datetime.time))
isinstance(value, (int, float, bool, str)). are | supported. was this tested?

ntrncic (dismissed):

src/hope/apps/payment/xlsx/xlsx_payment_plan_fsp_extra_fields_import_service.py
MarekBiczysko: it's a valid syntax, tested in  test_fsp_extra_fields_import_normalizes_json_values

src/hope/apps/payment/xlsx/xlsx_payment_plan_delivery_export_service.py
```diff
@@ -75,17 +76,30 @@ def __init__(
...
+        self.add_accounts_fields = False
         self.header_list = []
         self.template_columns = []
         self.core_fields = []
         self.flex_fields = []
         self.document_fields = []
+        self.fsp_extra_fields_headers = (
+            fsp_extra_fields_headers
+            if fsp_extra_fields_headers is not None
+            else self.get_fsp_extra_fields_headers(self.payment_plan.eligible_payments.only("extras"))
```
pkujawa, line 89: When the caller doesn't pass headers we take the columns from `self.payment_plan` only, and that breaks the follow-up export.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
The follow-up delivery export builds this service from the representative plan (`order_by("created_at").first()`) but writes rows from every child plan. So an extra field that only exists on a later child plan gets no column and its values never make it into the file. 
                                                                                                                                                                                                                                            
Same fix as the group exporter: pass `fsp_extra_fields_headers` computed over `Payment.objects.filter(parent__follow_up_instruction=self.instruction).eligible()`. 
                                                                                                                                                                                 
test_follow_up_instruction_exports.py has no extra-fields test at all, that's why this is green. We need a test with two child plans with a key only on the second.

src/hope/apps/payment/xlsx/xlsx_payment_plan_fsp_extra_fields_import_service.py
```diff
@@ -0,0 +1,201 @@
...
+        self.errors: list[XlsxError] = []
+        self.headers: list[Any] = []
+        self.header_to_index: dict[str, int] = {}
+        self.custom_headers: list[str] = []
+        self.payments = {
+            str(payment.unicef_id): payment
+            for payment in payment_plan.eligible_payments.select_related(
+                "currency",
+                "delivery_type",
+                "household_snapshot",
```
pkujawa, line 45: `household_snapshot` is only needed by `update_signature_hash()` in `import_payment_list`, but this dict is also built in the request when the view calls `validate()`, and validation only reads `unicef_id` and `fsp_extra_fields`. 
Could we skip `household_snapshot` here, as it's heavy, and load it where the signatures are recalculated?

pkujawa (changes requested):

src/hope/apps/payment/xlsx/xlsx_payment_plan_delivery_export_service.py
```diff
@@ -261,9 +293,14 @@ def prepare_headers(self, fsp_xlsx_template: "FinancialServiceProviderXlsxTempla
         column_list.extend(self.document_fields)
 
         # add headers for Account from FSP in PaymentPlan
-        if add_accounts_fields:
+        if self.add_accounts_fields:
             column_list.extend(self.account_fields_headers)
 
+        self.fsp_extra_fields_headers = [
```
johniak, line 299: Filtering the extras tail per plan here undoes the group-wide header list the group export computes on purpose. The group sheet has one header row (taken from the first plan) but rows from every plan, and `column_list` is not plan-invariant: `account_fields_headers` is empty for a plan whose snapshots have no `account_data`. Reproduced with template `["payment_id", "account_data"]`, plan A snapshots holding `{iban, number}` and plan B holding none plus `fsp_extra_fields = {"number": "EXTRA-NUMBER-B"}`:

```
('payment_id', 'iban', 'number', 'financial_institution_pk', 'financial_institution_name')
('RCPT-…001', 'IBAN-A',         'NUM-A', '',   '')
('RCPT-…002', 'EXTRA-NUMBER-B', None,    None, None)
```

Plan B's `number` lands under the `iban` header, so the FSP gets a wrong value where the cell used to be blank. Only reduce the list when it was not supplied by the caller:

```suggestion
        if fsp_extra_fields_headers_from_caller is None:
            self.fsp_extra_fields_headers = [
                header for header in self.fsp_extra_fields_headers if header not in column_list
            ]
```

A group-export test with two plans that differ in account headers would catch this.

src/hope/apps/payment/services/payment_gateway.py
```diff
@@ -217,7 +217,7 @@ def get_payload(self, obj: Payment) -> dict:
         if not payload.is_valid():
             raise PaymentGatewayAPI.PaymentGatewayAPIError(payload.errors)
 
-        return payload.data
+        return {**obj.fsp_extra_fields, **payload.data}
```
johniak, line 220: "System values win on collision" only holds for keys that are actually in `payload.data`. `origination_currency` is `required=False` and never set in `payload_data`, and `account` is a `DictField(required=False)` set only when the snapshot has account data — DRF omits unset optional fields from `.data`, so both slots stay open. An extra field named `account` then sends a plain string where the gateway parses an object. `get_system_controlled_headers` reserves FSP template columns and account keys but not payload key names, so nothing stops it at import time. Either drop payload keys from the merge or add them to the reserved set.

src/frontend/src/components/paymentmodule/PaymentPlanDetails/FspExtraFields/FspExtraFields.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  const canImport = hasPermissions(
+    PERMISSIONS.PM_IMPORT_XLSX_WITH_RECONCILIATION,
+    permissions,
+  );
+  const actionsDisabled = !isActiveProgram;
+
+  const handleImport = async (): Promise<void> => {
+    if (!file) return;
+    await importFile({
```
johniak, line 87: `mutateAsync` rejects, so this `await` without a `try/catch` throws an unhandled promise rejection every time the backend answers 400 — and that is the normal path here, since a bad header or an unknown payment id returns the XLSX error list as 400. `ImportXlsxPaymentPlanPaymentList.tsx:80-84` handles the same case explicitly.

```suggestion
    try {
      await importFile({
        // @ts-ignore File is required for multipart upload despite the generated string type.
        file,
      });
    } catch {
      // handled by onError
    }
```

johniak (changes requested):

pkujawa (dismissed):

johniak (approved):

PR #6260, opened 2026-07-28
Optimize React Query caching and refetch behavior

mmaciekk:
## Summary

Adds caching defaults to TanStack Query and fixes the stale-data bug that surfaced once caching was enabled.

Previously the `QueryClient` had no `defaultOptions`, so almost every query ran with `staleTime: 0` and `refetchOnWindowFocus: true`, refetching on every remount and window focus. Adding caching defaults fixed that, but exposed a pre-existing problem: query keys in this app are hand-written string arrays, not generated, and many had drifted out of sync with the keys queries actually read. Several mutations invalidated keys nothing reads, or invalidated nothing at all. This never mattered at `staleTime: 0` because everything refetched on remount. With caching on, those no-op invalidations left stale data on screen until the cache expired (e.g. edit a Programme name, save, and the details page still shows the old name).

This PR includes both the caching change and the invalidation fixes needed to make it safe.

## Part 1 — Caching and refetch changes

- `providers.tsx`: global `QueryClient` defaults — `staleTime` 60s, `gcTime` 10min, `refetchOnWindowFocus: false`, `retry: 1`. Hooks that need always-fresh data still set their own `staleTime: 0`.
- `PaymentPlanDetailsPage.tsx`, `TargetPopulationDetailsPage.tsx`: removed `refetchIntervalInBackground: true`, so the 3s status polls pause when the tab is hidden.
- Added `placeholderData: keepPreviousData` to 34 paginated tables so rows stay visible while the next page/filter loads.
- `VerifyManual.tsx`: removed a `setQueryData` that was immediately overwritten by `invalidateQueries` on the same key — dead code, no behavior change.

## Part 2 — Fix stale data after mutations

**2a. Global safety net.** Added a `MutationCache.onSuccess` handler in `providers.tsx` that invalidates active queries after any successful mutation, regardless of whether that mutation defines its own `onSuccess`. A new `isStaticReferenceQuery` predicate (`utils/queryCacheUtils.ts`, unit-tested) exempts static reference data — choices, geo areas, permissions/profile, business-area metadata — so those keep their long cache time. This is the fix for the reported bug: it only refetches mounted, non-static queries, and only after a mutation, so it's strictly less traffic than the old `staleTime: 0` behavior.

**2b. Query-key corrections.** Fixed the drifted keys so each flow is also correct without relying on the safety net:

- **Programme** — Create/Duplicate/Delete invalidated wrong keys; Edit never invalidated the details page.
- **Grievance** — ticket edit had no invalidation; payment-grievance verification used a wrong key; bulk assign/priority/urgency/note modals passed a key that never matched when a specific programme was selected.
- **Target Population** — Create only refreshed the autocomplete; Delete and Duplicate had no invalidation.
- **Payment Plan** — two invalidations had elements in the wrong order (no-op); status changes and Create/Edit/Delete never refreshed the list.
- **Feedback / Surveys / Communication** — adding a feedback message used a key matching no reader; create/edit didn't refresh lists.
- **Periodic Data Updates** — a malformed key matched nothing on detail pages; new online edits didn't appear in the list; some cross-list invalidations relied on sort order instead of the correct key.

## Out of scope

Per-endpoint volatility tiers, `select` adoption, a centralized query-key factory, removing redundant per-hook cache config, optimistic updates.

## Verification

- `bun run lint` — clean
- `tsc --noEmit` — no new errors (5 pre-existing errors on unrelated files also exist on `develop`)
- `bun run test` — new `queryCacheUtils` unit tests pass

### DIFF
--- a/src/frontend/src/components/paymentmodule/PaymentPlanDetails/PaymentPlanDetailsHeader/ApprovePaymentPlan.tsx
+++ b/src/frontend/src/components/paymentmodule/PaymentPlanDetails/PaymentPlanDetailsHeader/ApprovePaymentPlan.tsx
@@ -69,9 +69,6 @@ export function ApprovePaymentPlan({
       queryClient.invalidateQueries({
         queryKey: restQueryKey(RestService.restBusinessAreasProgramsPaymentPlansList),
       });
-      queryClient.invalidateQueries({
-        queryKey: ['businessAreasPaymentPlans'],
-      });
     },
   });
   const initialValues = {
--- a/src/frontend/src/components/paymentmodule/PaymentPlanDetails/PaymentPlanDetailsHeader/ApprovePaymentPlan.tsx
+++ b/src/frontend/src/components/paymentmodule/PaymentPlanDetails/PaymentPlanDetailsHeader/ApprovePaymentPlan.tsx
@@ -19,6 +19,7 @@ import { PaymentPlanDetail } from '@restgenerated/models/PaymentPlanDetail';
 import { RestService } from '@restgenerated/services/RestService';
 import { FormikTextField } from '@shared/Formik/FormikTextField/FormikTextField';
 import { useMutation, useQueryClient } from '@tanstack/react-query';
+import { restQueryKey } from '@utils/queryKeys';
 import { Field, Form, Formik } from 'formik';
 import { ReactElement, useState } from 'react';
 import { useTranslation } from 'react-i18next';
@@ -63,10 +64,10 @@ export function ApprovePaymentPlan({
       showMessage(t('Payment Plan has been approved.'));
       setApproveDialogOpen(false);
       queryClient.invalidateQueries({
-        queryKey: ['paymentPlan', businessArea, paymentPlan.id, programId],
+        queryKey: restQueryKey(RestService.restBusinessAreasProgramsPaymentPlansRetrieve),
       });
       queryClient.invalidateQueries({
-        queryKey: ['businessAreasProgramsPaymentPlansList'],
+        queryKey: restQueryKey(RestService.restBusinessAreasProgramsPaymentPlansList),
       });
       queryClient.invalidateQueries({
         queryKey: ['businessAreasPaymentPlans'],
--- a/src/frontend/src/components/paymentmodule/PaymentPlanDetails/PaymentPlanDetailsHeader/ApprovePaymentPlan.tsx
+++ b/src/frontend/src/components/paymentmodule/PaymentPlanDetails/PaymentPlanDetailsHeader/ApprovePaymentPlan.tsx
@@ -65,6 +65,12 @@ export function ApprovePaymentPlan({
       queryClient.invalidateQueries({
         queryKey: ['paymentPlan', businessArea, paymentPlan.id, programId],
       });
+      queryClient.invalidateQueries({
+        queryKey: ['businessAreasProgramsPaymentPlansList'],
+      });
+      queryClient.invalidateQueries({
+        queryKey: ['businessAreasPaymentPlans'],
+      });
     },
   });
   const initialValues = {
--- a/src/frontend/src/containers/tables/targeting/TargetPopulationTable/TargetPopulationTable.tsx
+++ b/src/frontend/src/containers/tables/targeting/TargetPopulationTable/TargetPopulationTable.tsx
@@ -5,6 +5,7 @@ import { PaginatedTargetPopulationListList } from '@restgenerated/models/Paginat
 import { TargetPopulationList } from '@restgenerated/models/TargetPopulationList';
 import { RestService } from '@restgenerated/services/RestService';
 import { keepPreviousData, useQuery } from '@tanstack/react-query';
+import { restQueryKey } from '@utils/queryKeys';
 import { adjustHeadCells } from '@utils/utils';
 import { ReactElement, useEffect, useMemo, useState } from 'react';
 import { usePersistedCount } from '@hooks/usePersistedCount';
@@ -74,22 +75,21 @@ export function TargetPopulationTable({
   }, [initialQueryVariables]);
 
   // Count query (enabled only on page 0)
+  const targetPopulationsCountParams = createApiParams(
+    {
+      businessAreaSlug: businessArea,
+      programCode: programId,
+    },
+    queryVariables,
+  );
   const { data: countData } = useQuery({
-    queryKey: [
-      'businessAreasProgramsTargetPopulationsCount',
-      businessArea,
-      programId,
-      queryVariables,
-    ],
+    queryKey: restQueryKey(
+      RestService.restBusinessAreasProgramsTargetPopulationsCountRetrieve,
+      targetPopulationsCountParams,
+    ),
     queryFn: () =>
       RestService.restBusinessAreasProgramsTargetPopulationsCountRetrieve(
-        createApiParams(
-          {
-            businessAreaSlug: businessArea,
-            programCode: programId,
-          },
-          queryVariables,
-        ),
+        targetPopulationsCountParams,
       ),
     enabled: page === 0,
   });
@@ -97,29 +97,27 @@ export function TargetPopulationTable({
   const persistedCount = usePersistedCount(page, countData);
 
   // Main data query
+  const targetPopulationsListParams = createApiParams(
+    {
+      businessAreaSlug: businessArea,
+      programCode: programId,
+    },
+    queryVariables,
+    { withPagination: true },
+  );
   const {
     data: targetPopulationsData,
     isLoading,
     isFetching,
     error,
   } = useQuery<PaginatedTargetPopulationListList>({
-    queryKey: [
-      'businessAreasProgramsTargetPopulationsList',
-      businessArea,
-      programId,
-      queryVariables,
-      page,
-    ],
+    queryKey: restQueryKey(
+      RestService.restBusinessAreasProgramsTargetPopulationsList,
+      targetPopulationsListParams,
+    ),
     queryFn: () => {
       return RestService.restBusinessAreasProgramsTargetPopulationsList(
-        createApiParams(
-          {
-            businessAreaSlug: businessArea,
-            programCode: programId,
-          },
-          queryVariables,
-          { withPagination: true },
-        ),
+        targetPopulationsListParams,
       );
     },
     placeholderData: keepPreviousData,
--- a/src/frontend/src/containers/tables/targeting/TargetPopulationTable/TargetPopulationTable.tsx
+++ b/src/frontend/src/containers/tables/targeting/TargetPopulationTable/TargetPopulationTable.tsx
@@ -4,7 +4,7 @@ import { useBaseUrl } from '@hooks/useBaseUrl';
 import { PaginatedTargetPopulationListList } from '@restgenerated/models/PaginatedTargetPopulationListList';
 import { TargetPopulationList } from '@restgenerated/models/TargetPopulationList';
 import { RestService } from '@restgenerated/services/RestService';
-import { useQuery } from '@tanstack/react-query';
+import { keepPreviousData, useQuery } from '@tanstack/react-query';
 import { adjustHeadCells } from '@utils/utils';
 import { ReactElement, useEffect, useMemo, useState } from 'react';
 import { usePersistedCount } from '@hooks/usePersistedCount';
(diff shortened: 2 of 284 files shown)


src/frontend/src/containers/tables/targeting/TargetPopulationTable/TargetPopulationTable.tsx
```diff
@@ -121,6 +121,7 @@ export function TargetPopulationTable({
         ),
       );
     },
+    placeholderData: keepPreviousData,
```
johniak, line 123: `UniversalRestTable` only accepts `isLoading` (prop at `:64`, passed to `loading` at `:150`), and with placeholder data present `isLoading` is false on every fetch after the first. So changing page, sort or filter now renders the previous result set with no spinner and no cue, while the pagination control already shows the new page. Same for all 38 tables here. Minimum fix is to pull `isPlaceholderData` off the query and pass `isLoading={isLoading || isPlaceholderData}`; better, thread `isFetching` into `UniversalRestTable` once and render a progress bar over the existing rows.

src/frontend/src/providers.tsx
```diff
@@ -15,7 +20,29 @@ interface ProvidersProps {
...
+// After ANY successful mutation, invalidate active queries so the UI reflects the write
+// immediately (create/edit/status-change etc.). Static reference data (choices, geo areas,
+// permissions) is exempt so it keeps its long staleTime. Individual mutations may still add
+// their own targeted invalidation; this is the safety net for the app's drifted query keys.
+// The onSuccess closure references `queryClient` lazily — it only runs on a mutation success,
+// long after the binding is initialised — so the self-reference is safe.
+const queryClient: QueryClient = new QueryClient({
+  mutationCache: new MutationCache({
+    onSuccess: () => {
+      queryClient.invalidateQueries({
```
johniak, line 29: No `queryKey` here means every mutation refetches every mounted non-static query — on Payment Plan details one button click refires the plan detail, payments table, payments count, program cycle and exchange rate. It also refetches things that no longer exist: `MutationCache.config.onSuccess` runs before the mutation's own `onSuccess` (`mutation.js:116` vs `:123`), so `DeletePaymentPlan` fires this while the plan detail query is still mounted, producing a GET on a deleted plan plus a retry before `navigate()`. Narrow the predicate to the mutated resource, or use `refetchType: 'none'`.

src/frontend/src/components/paymentmodule/PaymentPlanDetails/PaymentPlanDetailsHeader/ApprovePaymentPlan.tsx
```diff
@@ -65,6 +65,12 @@ export function ApprovePaymentPlan({
       queryClient.invalidateQueries({
         queryKey: ['paymentPlan', businessArea, paymentPlan.id, programId],
       });
+      queryClient.invalidateQueries({
+        queryKey: ['businessAreasProgramsPaymentPlansList'],
```
johniak: These two list invalidations went into nine dialogs but six sibling status transitions in the same folder were skipped, all still invalidating only `['paymentPlan', …]`: `HeaderButtons/AbortedPaymentPlanHeaderButtons.tsx:42`, `FinishedPaymentPlanHeaderButtons.tsx:45`, `LockedPaymentPlanHeaderButtons.tsx:44`, `LockedFspPaymentPlanHeaderButtons.tsx:42` and `:64`, `ReadyForClosurePaymentPlanHeaderButtons.tsx:42`, plus `SplitIntoPaymentLists.tsx`. So the description's "every status change" is really just the seven named ones.

johniak (changes requested):

src/frontend/src/utils/queryCacheUtils.ts
```diff
@@ -0,0 +1,24 @@
...
+// invalidates every query after any successful mutation to keep the UI fresh, but static
+// reference data (choices, geo areas, permissions) is never changed by those mutations and
+// carries a deliberately long staleTime. This predicate exempts that data from the blanket
+// invalidation so it is not needlessly refetched after every write.
+const STATIC_QUERY_KEY_ROOTS = new Set<string>([
+  'profile', // permissions / current-user profile (usePermissions, AppBar)
+  'businessArea', // business-area details
+  'allAreasTree', // geo areas
+  'beneficiaryGroups',
+  'restChoicesCountriesList',
```
Jeeziorny: caught by `return root.toLowerCase().includes('choices');`

src/frontend/src/providers.tsx
```diff
@@ -15,7 +20,29 @@ interface ProvidersProps {
   children: ReactNode[];
 }
 
-const queryClient = new QueryClient();
+// After ANY successful mutation, invalidate active queries so the UI reflects the write
```
Jeeziorny: I'm not FE guy and probably won't dig into it. Those comments are just for you and while I'm fine with whatever you put in the FE comment, you might reconsider having such long comments like this or in `src/frontend/src/utils/queryCacheUtils.ts§`, because they're using a lot of context :)

src/frontend/src/containers/tables/targeting/TargetPopulationTable/TargetPopulationTable.tsx
mmaciekk: Good catch. Rather than reverting to the full skeleton on every fetch (which would negate `keepPreviousData`), I threaded an optional `isFetching` prop through `UniversalRestTable` → `TableRestComponent` and render a `LinearProgress` (`data-cy="table-fetching-progress"`) over the still-visible rows whenever `isFetching && !loading`. `position: relative` is applied on the container only while the bar is shown, so idle tables render identically and existing snapshots stay stable. `isFetching` is now destructured from the query and passed at all 34 tables (aliases and the multi-query/multi-table files handled). Two files — `LinkedGrievancesModal` and `CreateImportFromProgramPopulation` — do not render `UniversalRestTable`, so adding `isFetching` there would be an unused variable; left as-is.

src/frontend/src/providers.tsx
mmaciekk: Fixed by adding `refetchType: 'none'` to the safety-net invalidation. It now marks non-static queries stale **without** an active refetch, so: (1) a delete no longer re-fires a GET (+`retry`) on the just-removed resource before the mutation's own `onSuccess` navigates away, and (2) a single mutation on e.g. Payment Plan details no longer refires every mounted query. Immediate on-screen refresh continues to come from the per-flow targeted invalidations (Part 2b). Narrowing the predicate to the mutated resource isn't feasible today — mutations carry no `meta`, and `mutationKey` is set on only 2 of ~28 — so `refetchType: 'none'` is the pragmatic route you suggested.

src/frontend/src/components/paymentmodule/PaymentPlanDetails/PaymentPlanDetailsHeader/ApprovePaymentPlan.tsx
mmaciekk: Added the two list invalidations to the six skipped transitions to match this reference: `AbortedPaymentPlanHeaderButtons` (reactivate), `FinishedPaymentPlanHeaderButtons` (`markReadyForClosure` **and** the previously-uninvalidated `sendToPaymentGateway`), `LockedPaymentPlanHeaderButtons` (unlock), `LockedFspPaymentPlanHeaderButtons` (unlock + `sendForApproval`), `ReadyForClosurePaymentPlanHeaderButtons` (`sendBack`), and `SplitIntoPaymentLists` (which also needed `useQueryClient` wired in, as it had no query client at all). Existing `exact: false` flags preserved.

src/frontend/src/utils/queryCacheUtils.ts
```diff
@@ -0,0 +1,24 @@
+// Query keys in this app are hand-written string-array literals (not derived from
```
Jeeziorny: Why do we prefer to fix the drift (from description: "query keys in this app are hand-written string-array literals, not derived from restgenerated, so they have drifted") instead of prevent it to happen in the future again? Can't we derive the queryKey from something so that we don't need to hand-write it?

Jeeziorny (commented):
From description:

```
This is the guarantee that fixes the reported bug
```

Is there reported bug somewhere that needs to e linked to that PR?

src/frontend/src/utils/queryCacheUtils.ts
mmaciekk: Great idea — that's exactly what we should do, so I did it in this PR: refactoring the keys to derive them and prevent drift from ever creeping back.

Added `utils/queryKeys.ts` with `restQueryKey(RestService.method[, params])`, which derives the key root from the fetcher's own name (normalized to our existing convention: drop the `rest` prefix, lower-case the first char — so `restBusinessAreasProgramsList` → `businessAreasProgramsList`). A reader and its invalidators now derive from the same `RestService` method, so they can't drift. `restQueryKey(m)` gives the bare list/invalidation prefix; `restQueryKey(m, params)` gives the scoped reader/detail key.

Since the normalized root equals the keys we already hand-write, migrating is behavior-preserving. I migrated readers + invalidators end-to-end for Programme, Payment Plan, Targeting, Feedback/Survey/Communication, Grievance, and Periodic Data Updates — and normalized `useHopeDetailsQuery` so every hook-based detail reader shares the derived root. Duplicate keys that hit one fetcher were folded into a single derived key (e.g. `program`/`businessAreaProgram` → `restBusinessAreasProgramsRetrieve`).

Added a `warn`-level ESLint rule (`no-restricted-syntax` on hand-written `queryKey` array literals) so new code goes through the factory and the remaining short-hand aliases get migrated over time; it flips to `error` once the sweep is complete.

src/frontend/src/utils/queryCacheUtils.ts
mmaciekk: Correct — `restChoicesCountriesList` is already matched by the `includes('choices')` heuristic, so I removed it from the set.

src/frontend/src/providers.tsx
mmaciekk: Fair — trimmed both block comments (`providers.tsx` and `queryCacheUtils.ts`) down to the non-obvious bits.

mmaciekk:
> @Jeeziorny: "Is there reported bug somewhere that needs to be linked to that PR?"

No formal issue — these were no-op invalidations I found while investigating this PR, so I fixed them here rather than filing tickets. Each silently matched nothing and left stale data after a mutation:

- **PDU pending-approval / pending-merge**: the invalidator keyed on the live `queryVariables` filter object, so once the user changed a filter it no longer matched the cached list. Now invalidates the bare `OnlineEditsList` prefix, matching every status-filtered variant.
- **`ProgramCyclesTablePaymentModule`**: a second invalidator keyed on `['programCycles', queryVariables, …]` never matched (the sibling `exact: false` call already covered it). Removed the dead call.
- **`PaymentVerificationTable`**: both reader keys pointed at unrelated fetchers (`…HouseholdsCount` / `…PaymentPlansList`) instead of the `PaymentVerifications` methods they actually call, so nothing could invalidate them precisely. Now derived from the real fetcher.
- **Grievance lists**: bulk-action invalidators used short-form aliases that didn't match `LookUpLinkedTicketsTable`'s full `…List` keys, so that table went stale after bulk actions. Deriving both sides from the same fetcher closes the gap.

src/hope/apps/core/api/views.py
```diff
@@ -108,7 +108,7 @@ def all_collector_fields_attributes(self, request: Request, *args: Any, **kwargs
             200: FieldAttributeSerializer(many=True),
         },
     )
-    @action(detail=True, methods=["get"], url_path="all-fields-attributes")
+    @action(detail=True, methods=["get"], url_path="all-fields-attributes", pagination_class=None)
```
Jeeziorny, line 111: Why this change?

Jeeziorny (approved):

src/hope/apps/core/api/views.py
mmaciekk: Doesn'\''t change behaviour — only the schema.

The endpoint returns a plain array, and never uses the paginator. But it picked up the viewset'\''s pagination class, so drf-spectacular documented it as a paginated object with `limit`/`offset` params. The generated TS type was then wrong, which is why `TargetingCriteriaForm` needed three `@ts-ignore`s.

`pagination_class=None` makes the schema say "array", which is what the endpoint actually sends. `all_kobo_projects` in the same viewset already does this.

johniak: Because it returns a list. Without pagination_class=None, the OpenAPI schema will show it as paginated, but it is not.
```Response(FieldAttributeSerializer(result_list, many=True).data, status=200)```

johniak (approved):

PR #6264, opened 2026-07-29
[HOTFIX] USDC handle photo flex field

pkujawa:
The USDC registration is the first Aurora import to include image-type flex fields (wallet_num_image_i_f, id_wallet_image_i_f). The existing code treated them as ordinary flex fields and copied the raw base64 submission value into flex_fields.                                  
                                                                                                                                                                                                                                                                                                                             
The import itself succeeded — the household and individual were created correctly — but the API serializer resolves IMAGE-type flex fields via default_storage.url(value), which expects a storage path. Given a base64 blob instead, the individual detail page returned a 500.                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                             
  What changed                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                             
  - Split the wallet image fields into a dedicated INDIVIDUAL_IMAGE_FLEX_FIELDS list so they're handled as images rather than copied verbatim.                                                                                                                                                                               
  - For each image field, decode the base64 and save it through default_storage, storing the returned path — the same mechanism documents already use (_prepare_picture_from_base64).   
  
  
## Also fixed: phone number validity not set on Aurora imports                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                             
  Imported individuals had `phone_no_valid` and `phone_no_alternative_valid` left unset (`None`).                                                                                 
                                                                                                                                                                                                                                                                                                                             
  This was a regression from the datahub removal. Pending import records used to live in a separate `ImportedIndividual` model, and `recalculate_phone_numbers_validity` took a `model` argument so each caller could pass the model that applied. Once the datahub was removed and `ImportedIndividual` was folded into     
  `Individual` — pending versus merged is now just `rdi_merge_status` — the helper was left looking up the existing row through `Individual.objects`, which is the MERGED-only manager (`SoftDeletableMergedManager`). Import records are PENDING, so that lookup found nothing on the update path, the branch was skipped,  
  and validity was never recomputed. The fix switches the lookup to `all_merge_status_objects` (MERGED + PENDING) and drops the now-unused `model` parameter, since `Individual.save` is the only caller.                                                                                                                    
                                                                                                                                                                                                                                                                                                                             
  Sri Lanka had a second, unrelated cause: it creates individuals with `bulk_create`, which bypasses the model's `save()` hook entirely, so validity was never calculated there at all. The fix adds an explicit `calculate_phone_numbers_validity` pass before the bulk insert.                                             
                                                                                                                                                                                                                                                                                                                             
  Added valid and invalid phone-number tests across the affected Aurora services (USDC, Ukraine, Czech Republic, and Sri Lanka).                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                      
## Also fixed: Aurora image handling

A malformed image no longer aborts the whole batch. - `_prepare_picture_from_base64` called `base64.b64decode` unguarded, so a truncated or malformed blob raised `binascii.Error` — not a `ValidationError` — which escaped the per-record handler. It now raises `ValidationError`, so a bad image fails only its own record. The guard lives in the shared base helper, so it covers every Aurora service (base flex, Czech, generic, Nigeria, Sri Lanka, Ukraine/USDC).

### DIFF
--- a/src/hope/contrib/aurora/services/ukraine_flex_registration_service.py
+++ b/src/hope/contrib/aurora/services/ukraine_flex_registration_service.py
@@ -3,6 +3,7 @@ from typing import Any
 import uuid
 
 from django.core.exceptions import ValidationError
+from django.core.files.storage import default_storage
 from django.forms import modelform_factory
 
 from hope.apps.core.utils import (
@@ -353,7 +354,8 @@ class UkraineUSDCRegistrationService(UkraineBaseRegistrationService):
         ),
     }
 
-    INDIVIDUAL_FLEX_FIELDS: list[str] = ["wallet_num_image_i_f", "id_wallet_image_i_f"]
+    INDIVIDUAL_FLEX_FIELDS: list[str] = []
+    INDIVIDUAL_IMAGE_FLEX_FIELDS: list[str] = ["wallet_num_image_i_f", "id_wallet_image_i_f"]
 
     def create_household_for_rdi_household(self, record: Any, registration_data_import: RegistrationDataImport) -> None:
         record_data_dict = record.get_data()
@@ -411,7 +413,12 @@ class UkraineUSDCRegistrationService(UkraineBaseRegistrationService):
     ) -> dict:
         individual_data = super()._prepare_individual_data(individual_dict, household, registration_data_import)
         individual_data["estimated_birth_date"] = False
-        if flex_fields := build_flex_arg_dict_from_list_if_exists(individual_dict, self.INDIVIDUAL_FLEX_FIELDS):
+        flex_fields = build_flex_arg_dict_from_list_if_exists(individual_dict, self.INDIVIDUAL_FLEX_FIELDS)
+        for field_name in self.INDIVIDUAL_IMAGE_FLEX_FIELDS:
+            if image_base64 := individual_dict.get(field_name):
+                image = self._prepare_picture_from_base64(image_base64, f"{field_name}_{uuid.uuid4()}")
+                flex_fields[field_name] = default_storage.save(image.name, image)
+        if flex_fields:
             individual_data["flex_fields"] = flex_fields
         return individual_data
 
--- a/src/hope/contrib/aurora/services/ukraine_flex_registration_service.py
+++ b/src/hope/contrib/aurora/services/ukraine_flex_registration_service.py
@@ -395,6 +395,8 @@ class UkraineUSDCRegistrationService(UkraineBaseRegistrationService):
             )
             individual.phone_no = phone_no
             individual.detail_id = record.source_id
+            if image_flex_fields := self._prepare_wallet_image_flex_fields(individual_dict):
+                individual.flex_fields = image_flex_fields
             individual.save()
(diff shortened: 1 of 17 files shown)


MarekBiczysko (dismissed):

kosior (dismissed):

src/hope/contrib/aurora/services/ukraine_flex_registration_service.py
```diff
@@ -411,7 +412,12 @@ def _prepare_individual_data(
     ) -> dict:
         individual_data = super()._prepare_individual_data(individual_dict, household, registration_data_import)
         individual_data["estimated_birth_date"] = False
-        if flex_fields := build_flex_arg_dict_from_list_if_exists(individual_dict, self.INDIVIDUAL_FLEX_FIELDS):
+        flex_fields = {}
+        for field_name in self.INDIVIDUAL_IMAGE_FLEX_FIELDS:
+            if image_base64 := individual_dict.get(field_name):
+                image = self._prepare_picture_from_base64(image_base64, str(uuid.uuid4()))
```
johniak: `b64decode` raises `binascii.Error` on a truncated/malformed blob (or `TypeError` on a non-string), and `process_records` only catches `ValidationError` per record — anything else hits the outer `except Exception`, marks the whole RDI as IMPORT_ERROR and re-raises, leaving already-imported records without `STATUS_IMPORTED`. Before this change the value was copied verbatim and could never raise. Wrap the decode and raise `ValidationError` so one bad record doesn't kill the batch:
```python
try:
    image = self._prepare_picture_from_base64(image_base64, str(uuid.uuid4()))
except (binascii.Error, TypeError, ValueError) as e:
    raise ValidationError(f"Invalid image data for {field_name}") from e
```

src/hope/contrib/aurora/services/ukraine_flex_registration_service.py
```diff
@@ -411,7 +412,12 @@ def _prepare_individual_data(
...
+                flex_fields[field_name] = default_storage.save(image.name, image)
```
johniak: The file is written to storage before the individual passes `IndividualForm` validation. If `_create_object_and_validate` raises afterwards, the record's `atomic()` block rolls the DB back but the blobs stay in storage as unreferenced orphans — and every Aurora retry of a failing record writes fresh copies (uuid-seeded names, never reused). Consider saving the images only after the individual validates, or cleaning them up when the record fails.

johniak (changes requested):

src/hope/contrib/aurora/services/ukraine_flex_registration_service.py
pkujawa: Actually every existing aurora import has the same treatment of photo/document, that the malformed blob would kill the whole batch.
I applied change to the helper then, rather than at the 8 call sites (across 6 services).
Added to PR description.

johniak (approved):

kosior (approved):

PR #6269, opened 2026-07-30
[FEATURE] 332302 enhance storage file paths

nerjan:
[AB#332302](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/332302)

Uploaded files all landed in one flat namespace, and media storage is configured to overwrite on a name collision. Two unrelated uploads that shared a file name replaced each other. The worst situation was with the files written under fixed names: every universal update job saved its template as template.xlsx and its backup as snapshot.json.

- **Every upload has its own folder** — `upload_to=UniqueUploadPath("<prefix>")` files
  each upload at `prefix/YYYY/MM/<random-uuid>/<name>`, so no two uploads can land on the same
  path even when the names match.
- **Widen the file name columns to 255** - because of much longer paths because of prefix.
- **Display file names as the last part only** — `upload_basename()` trims
  the folder.
- **Name Kobo attachments ourselves** — the importer reused the enumerator's file name,
  so two submissions with `signature-14_59_24.jpg` overwrote each other. It now generates
  a unique name.
- **Flex field images** — they are kept as a name inside a
  JSONField with no FileField to carry an `upload_to`, so `save_flex_field_image()`
  applies the same path and the same name validation a FileFIeld would.
- **Save grievance photos under the same path as the rest of their model** — a photo attached through a grievance ticket like document photos now land under the document folder like every other document photo.

- **NOTE!!! Leave `Individual.photo` as it was** — its name is sent to the deduplication engine,
  which resolves it against storage, so it has to stay exactly was it was.


domdinicola:
@nerjan let's discuss this and we may rethink it a little

PR #6273, opened 2026-07-31
BaseAPI passing creds in constructor

Jeeziorny:
## Problem

BaseAPI requires API URL and API Token to be set in settings. If we get callback URL as a full URL we don't need to / we can't set it in settings, but BaseAPI children requires them to have it defined. This PR implements a workaround that enables API creds to be passed in a constructor and use settings values as a fallback.

## Context

This PR is required for adjusting to new deduplication flow. CW will be able to DELETE RDI using HOPE endpoint and HOPE needs to tell CW: "We finished deleting a RDI" using callback provided by CW.

### DIFF
--- a/src/hope/apps/core/api/mixins.py
+++ b/src/hope/apps/core/api/mixins.py
@@ -76,7 +76,7 @@ class BaseAPI:
             self._client.headers.update({"Authorization": f"Token {self.api_key}"})
 
     def get_url(self, endpoint: str) -> str:
-        return f"{self.api_url}{endpoint}"
+        return f"{self.api_url.rstrip('/')}/{endpoint.lstrip('/')}"
 
     def validate_response(self, response: Response) -> Response:
         if not response.ok:
--- a/src/hope/apps/core/api/mixins.py
+++ b/src/hope/apps/core/api/mixins.py
@@ -56,9 +56,9 @@ class BaseAPI:
                 f"Missing {self.__class__.__name__} settings: {', '.join(missing_setting_names)}"
             )
 
-    def __init__(self) -> None:
-        self.api_url = self.get_api_url()
-        self.api_key = self.get_api_key()
+    def __init__(self, api_url: str | None = None, api_key: str | None = None) -> None:
+        self.api_url = api_url if api_url is not None else self.get_api_url()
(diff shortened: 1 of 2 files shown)


MarekBiczysko (dismissed):

src/hope/apps/core/api/mixins.py
kosior, line 79: I wonder if we need to change the way how url is created. If we pass:

api_url: `https://cw.example.com/api/cb/`
endpoint: `/rest/rdi/delete-callback/`

We'd end up with: `https://cw.example.com/api/cb//rest/rdi/delete-callback/`

Side note: Should we have some sort of api_url validation? To check for example if the domain is correct and allowed. Maybe validation could be against the api url value that we have in settings 🤔 or does the validation exist in some other place?

Jeeziorny: After offline discussion we came to following conclusions:
1. `get_url` should use [urllib](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin)
2. We need to place validation in init, so that case mentioned by kosior could be detected before the call being made.
3. Ticket: https://unicef.visualstudio.com/ICTD-HCT-MIS/_boards/board/t/HOPE%20Team/Stories?System.AreaPath=HOPE&workitem=332966

kosior (changes requested):
After some consideration I think we should include https://unicef.visualstudio.com/ICTD-HCT-MIS/_boards/board/t/HOPE%20Team/Stories?System.AreaPath=HOPE&workitem=332966 in this PR or a least it should be implemented before the new deduplication flow.

src/hope/apps/core/api/mixins.py
kosior: We'll also need https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/334130

I think we should ditch the urllib - it drops the last path segment.

Jeeziorny:
As we agreed offline with @kosior 
1. For DD Host validation will be performed during CW POST.
2. https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/334130 -> separate scope, not in this PR.

kosior (approved):

MarekBiczysko (approved):

PR #6274, opened 2026-07-31
331730 slow queries n1

MarekBiczysko:
[AB#331730](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/331730)

  ## Changes

- Optimized payment verification XLSX export by eagerly loading related payment, household, administrative area, and head-of-household data.

- Optimized Payment Gateway synchronization:
  - bulk creates and updates FSPs;
  - skips writes for unchanged FSPs;
  - indexes gateway records by remote ID;
  - bulk updates payments while preserving signatures.

- Optimized payment-plan creation and rebuilding:
  - loads collectors in bulk;
  - validates collector accounts in batches;
  - reduces repeated account, delivery configuration, mapping, and financial institution queries.

- Optimized payment household snapshot generation:
  - prefetches individuals, roles, documents, and related objects;
  - resolves delivery data in batches;
  - calculates needs-adjudication ticket counts in bulk;
  - correctly handles collectors belonging to another household and households without a primary collector.

  ## Sentry issues

  - Sentry #2587 (https://monitoring.hope.unicef.org/organizations/hope/issues/2587/?environment=production&project=4&query=is%3Aunresolved&referrer=issue-stream)
  - Sentry #2589 (https://monitoring.hope.unicef.org/organizations/hope/issues/2589/?environment=production&project=4&query=is%3Aunresolved&referrer=issue-stream)
  - Sentry #2559 (https://monitoring.hope.unicef.org/organizations/hope/issues/2559/?environment=production&project=4&query=is%3Aunresolved&referrer=issue-stream)
  - Sentry #2660 (https://monitoring.hope.unicef.org/organizations/hope/issues/2660/?environment=production&project=4&query=is%3Aunresolved&referrer=issue-stream)


pkujawa (dismissed):

src/hope/apps/payment/services/payment_gateway.py
```diff
@@ -549,39 +578,55 @@ def sync_fsps(self) -> None:
...
-                fsp.save()
+            fsp.payment_gateway_id = payment_gateway_id
+            fsp.vision_vendor_number = fsp_data.vendor_number
+            fsp.name = fsp_data.name
+            fsp.data_transfer_configuration = data_transfer_configuration
+
+            if created:
+                fsps_to_create.append(fsp)
+            elif changed:
+                fsp.updated_at = sync_time
```
kosior, line 611: Should we also set `updated_at` on created fsps? Otherwise I think it will be null. I'm not sure if this field is used anywhere that matters, but previously it was set on .save.

src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -485,16 +486,33 @@ def create_payments(payment_plan: PaymentPlan) -> None:
...
 
             has_valid_wallet = True
-            if payment_plan.delivery_mechanism and payment_plan.financial_service_provider:
-                has_valid_wallet = PaymentDataCollector.validate_account(
-                    payment_plan.financial_service_provider,
-                    payment_plan.delivery_mechanism,
-                    collector,
-                )
+            if delivery_mechanism and financial_service_provider:
+                has_valid_wallet = wallet_validity_by_collector_id.get(collector.id, True)
```
kosior: Just double checking, the `has_valid_wallet` will be True if collector.id is not in the dict, is that default ok?

src/hope/apps/payment/services/payment_household_snapshot_service.py
```diff
@@ -150,18 +266,25 @@ def get_individual_snapshot(individual: Individual, payment: Payment | None = No
...
-    ).exists()
+    is_hh_collector = (
+        individual.id in collector_ids
+        if collector_ids is not None
+        else IndividualRoleInHousehold.objects.filter(
+            role__in=[ROLE_PRIMARY, ROLE_ALTERNATE],
+            household=individual.household,
+            individual=individual,
+        ).exists()
+    )
```
kosior: Could `is_hh_collector` be only computed if payment is not None? 

If yes, that could optimise:
`create_backup_snapshot.py -> get_household_snapshot(household)` where we don't pass a payment.

src/hope/apps/payment/services/payment_gateway.py
```diff
@@ -634,12 +679,11 @@ def update_payment(  # noqa: PLR0913
         payment_plan: PaymentPlan,
         exchange_rate: Decimal | float | None,
         log_pairs: "list[tuple[Payment, Payment]]",
-    ) -> None:
-        try:
-            matching_pg_payment = next(p for p in pg_payment_records if p.remote_id == str(payment.id))
-        except StopIteration:
+    ) -> tuple[str, ...] | None:
+        matching_pg_payment = next((p for p in pg_payment_records if p.remote_id == str(payment.id)), None)
```
kosior: This scan is done once per payment. I think we could improve this by adding a mapping elsewhere `{p.remote_id: p}` and passing it instead of `pg_payment_records`. If you feel like it's not a part of the ticket's scope maybe we could create a separate ticket.

src/hope/models/payment_data_collector.py
```diff
@@ -84,13 +113,120 @@ def resolve_required_field(
...
+        delivery_mechanism: "DeliveryMechanism",
+        collectors: Iterable[Individual],
+    ) -> AccountValidityByCollector:
+        if not delivery_mechanism.account_type_id:
+            return {collector.id: True for collector in collectors}
+        _, validity_by_collector = cls._process_collectors(fsp, delivery_mechanism, collectors)
+        return validity_by_collector
+
     @classmethod
     def delivery_data(
```
kosior, line 259: If I understand correctly the `_process_collectors` re-implements delivery_data and validate_account. Just thinking if there a good way to reuse the logic somehow and get rid of the duplication?

Like could we do something like: `_process_collectors(fsp, delivery_mechanism, [collector])`

src/hope/models/payment_data_collector.py
```diff
@@ -84,13 +113,120 @@ def resolve_required_field(
...
+            if account:
+                delivery_data.setdefault("number", account.number)
+                delivery_data.setdefault(
+                    "financial_institution_name", getattr(account.financial_institution, "name", "")
+                )
+                delivery_data.setdefault(
+                    "financial_institution_pk", str(getattr(account.financial_institution, "pk", ""))
+                )
+            delivery_data_by_collector[collector.id] = delivery_data
+        return delivery_data_by_collector, validity_by_collector
```
kosior: When the `_process_collectors` method is used one value is always discarded. 

For example if I want to only validate_accounts is there any logic that is computed only for `delivery_data_by_collector`? 

It's difficult to tell by reading the `_process_collectors` where the account validation happens and which code is only for delivery_data. I understand it overlaps, but maybe it could be worth splitting that functionality further into a class or something.

src/hope/apps/payment/services/payment_household_snapshot_service.py
```diff
@@ -62,28 +65,102 @@ def bulk_create_payment_snapshot_data(payments_ids: list[str]) -> None:
...
+    needs_adjudication_counts: dict[UUID, int] | None = None,
+) -> PaymentHouseholdSnapshot:
     household = payment.household
-    household_data = get_household_snapshot(household, payment)
+    household_data = get_household_snapshot(
+        household,
+        payment,
+        delivery_data_by_individual_id,
+        collector_ids,
+        needs_adjudication_counts,
```
kosior: Those three arguments travel through create_payment_snapshot_data -> get_household_snapshot -> get_individual_snapshot. Maybe we could wrap them as a snapshot context (dataclass?) - less clutter, types won't be repeated.

src/hope/apps/payment/services/payment_household_snapshot_service.py
```diff
@@ -62,28 +65,102 @@ def bulk_create_payment_snapshot_data(payments_ids: list[str]) -> None:
...
+        ]
         PaymentHouseholdSnapshot.objects.bulk_create(to_create)
 
 
-def create_payment_snapshot_data(payment: Payment) -> PaymentHouseholdSnapshot:
+def create_payment_snapshot_data(
+    payment: Payment,
+    delivery_data_by_individual_id: DeliveryDataByCollector | None = None,
+    collector_ids: set[UUID] | None = None,
+    needs_adjudication_counts: dict[UUID, int] | None = None,
```
kosior: Here we might not need the default values if the `test_create_payment_snapshot_data_uses_single_collector_fallback` is updated.

src/hope/apps/payment/services/payment_gateway.py
MarekBiczysko: bulk_create() populates both timestamps in pre_save(). I'll add unit test assertion for this case

src/hope/apps/payment/services/payment_plan_services.py
MarekBiczysko: I'll make it more explicit

pkujawa (approved):

kosior (approved):

ntrncic:
@MarekBiczysko description of the PR please.

PR #6282, opened 2026-08-03
Set a per-process application_name on database connections

kosior:
Tags every Postgres connection with an `application_name` identifying the kind of process that opened it - https://www.postgresql.org/docs/15/runtime-config-logging.html#GUC-APPLICATION-NAME

Beyond diagnostics, having the process role available as settings.PROCESS_ROLE lets us tune database settings per process.

Part of https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/331051/


MarekBiczysko (approved):

src/hope/config/settings.py
```diff
@@ -91,6 +92,8 @@
 
 DB_ZOMBIE_TIMEOUT_MS = 4 * 60 * 60 * 1000
 
+PROCESS_ROLE = get_process_role()
+
 # prefix all non-production emails
 if ENV != "prod":
```
saxix, line 98: this should be removed

saxix (changes requested):
I like thie PR, but I would take this opportunity to remove bad practices as 'if ENV != ...'

johniak (approved):

src/hope/config/settings.py
ntrncic: created a new ticket and will be handled in seperate PR. @saxix

PR #6291, opened 2026-08-05
[333832] remove environment based setting logic

kosior:
Closes: [AB#333832](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/333832)

---
**Breaking**: non-production emails no longer carry an [env] prefix in the subject. The environment marker moves to the sender display name - set per deployment via DEFAULT_EMAIL_DISPLAY (hope-deployment 0173ddc, e.g. [DEV] HOPE), so subjects are now identical across all environments. Any mail-client rule or filter matching on a subject prefix must be re-pointed at the sender name instead.

**Breaking**: the ENV == "prod" guard is gone, so superusers (and, for payment plan actions, staff users) are now excluded from notification recipients in every environment, not just prod. Testing notifications on dev/stg/trn therefore needs a non-superuser, non-staff account with the relevant role - or enable the new NOTIFY_INTERNAL_USERS Constance flag (default off) on that environment to have internal accounts included again.

### DIFF
--- a/src/hope/config/settings.py
+++ b/src/hope/config/settings.py
@@ -85,18 +85,12 @@ EMAIL_PORT = env("EMAIL_PORT")
 EMAIL_HOST_USER = env("EMAIL_HOST_USER")
 EMAIL_HOST_PASSWORD = env("EMAIL_HOST_PASSWORD")
 EMAIL_USE_TLS = env.bool("EMAIL_USE_TLS")
+EMAIL_SUBJECT_PREFIX = ""
 
 # Get the ENV setting. Needs to be set in .bashrc or similar.
 ENV = env("ENV")
 
 DB_ZOMBIE_TIMEOUT_MS = 4 * 60 * 60 * 1000
-
-# prefix all non-production emails
-if ENV != "prod":
-    EMAIL_SUBJECT_PREFIX = f"{ENV}"
-else:
-    EMAIL_SUBJECT_PREFIX = ""
-
 
 RO_CONN = env.db("REP_DATABASE_URL")
 RO_CONN.update(
--- a/src/hope/apps/payment/notifications.py
+++ b/src/hope/apps/payment/notifications.py
@@ -97,17 +97,14 @@ class PaymentNotification:
             .exclude(expiry_date__lt=timezone.now())
             .distinct()
         )
-        users = (
+        return (
             User.objects.filter(
                 Q(role_assignments__in=role_assignments) | Q(partner__role_assignments__in=role_assignments)
             )
             .exclude(id=self.action_user.id)
+            .exclude(Q(is_superuser=True) | Q(is_staff=True))
             .distinct()
         )
-
-        if settings.ENV == "prod":
-            users = users.exclude(Q(is_superuser=True) | Q(is_staff=True))
-        return users
 
     def _prepare_email(self) -> MailjetClient:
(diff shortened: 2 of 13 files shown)


src/hope/config/settings.py
```diff
@@ -85,19 +85,13 @@
...
+EMAIL_SUBJECT_PREFIX = ""
 
 # Get the ENV setting. Needs to be set in .bashrc or similar.
 ENV = env("ENV")
 
 DB_ZOMBIE_TIMEOUT_MS = 4 * 60 * 60 * 1000
 
-# prefix all non-production emails
-if ENV != "prod":
-    EMAIL_SUBJECT_PREFIX = f"{ENV}"
```
pkujawa, line 99: Are we sure that we want to remove the prefixes as well?
I believe they were specifically added because of confusion (https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/198143)

src/hope/config/settings.py
```diff
@@ -85,19 +85,13 @@
 EMAIL_HOST_USER = env("EMAIL_HOST_USER")
 EMAIL_HOST_PASSWORD = env("EMAIL_HOST_PASSWORD")
 EMAIL_USE_TLS = env.bool("EMAIL_USE_TLS")
+EMAIL_SUBJECT_PREFIX = ""
```
pkujawa, line 89: I think after the changes, `EMAIL_SUBJECT_PREFIX` is not used anywhere?

pkujawa (commented):
I believe the `EMAIL_SUBJECT_PREFIX` logic was fine, the check was inside the config.

 For the superuser exclusion - how about we add a flag `EXCLUDE_INTERNAL_USERS_FROM_NOTIFICATIONS` (default `True`)? We will get exactly the same behaviour as before, just without the `ENV == "prod"` check in the query. And we would be able to test emails like before.
Also, we already keep constances for emails like `ENABLE_MAILJET`, `SEND_PAYMENT_PLANS_NOTIFICATION` etc.

I would just double check if we actually want to change the result and remove the impact of these settings.

pkujawa:
BTW
To link the azure ticket, just use AB#<ticket_number> in the PR description (AB#333832)
This way, it will be linked also from azure side.

src/hope/config/settings.py
kosior: The EMAIL_SUBJECT_PREFIX was "replaced" with DEFAULT_EMAIL_DISPLAY setting which can be configured in the hope-deployment repo, per environment. We will get emails like this - from: `[DEV] HOPE <hope@example.com>`

src/hope/config/settings.py
kosior: Django default for EMAIL_SUBJECT_PREFIX is `[Django] `. ADMINS and ALERTS_EMAIL are set, but wasn't sure if code error notifications are sent, so I decided to leave it as it was.

pkujawa (approved):

domdinicola:
@kosior do we have a related pr in hope-deployment?

kosior:
> @kosior do we have a related pr in hope-deployment?

@domdinicola https://github.com/unicef/hope-deployment/pull/215 here it is. It's already merged.

domdinicola (approved):

PR #6292, opened 2026-08-06
strip trailing slash from the RapidPro host

nerjan:
rapid_pro_host is a field on business_area that has no validation about trailing slashes. 
Record entered as "https://url/" produced
"https://url//api/v2/runs.json" with double slashes "//" and RapidPro answered with 404 on
every periodic verification sync. 
This PR normalizes the resolved host to be without slash at the end.


johniak (approved):

kosior:
@ntrncic @johniak @nerjan It's fine as a quick fix - but the same issue likely exists in a few other places. I raised a follow-up ticket to introduce a shared URL-join helper and normalize the base addresses system-wide (https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/334130)

PR #6294, opened 2026-08-06
328305_ezCash_vision_integration_flow

MarekBiczysko:
[AB#328305](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/328305)

## Summary

Implements the automatic Vision integration flow for Payment Plans.

When both Vision flags are enabled, authorized Payment Plans are sent to Vision, remain blocked in IN_REVIEW until an FC is assigned, and are then released automatically. Payment Gateway plans continue automatically to PG delivery; non-PG plans use the existing XLSX export flow.

  ## Main changes

  - Added global and Business Area Vision enablement checks.
  - Added automatic Vision sending after authorization.
  - Added Vision workflow states, logging, retry handling, and UI status display.
  - Processed Vision callbacks with duplicate and out-of-state callback protection.
  - Matched fc_num to an FC group and atomically assigned all its items.
  - Returned 400 / KO / FC not found when automatic FC assignment fails.
  - Blocked manual FC assignment, release, and PG sending during active Vision workflows.
  - Added Django admin recovery:
      - Retry failed Vision sends.
      - Assign one or more same-group FC items after callback/FC failures.
      - Automatically release and start PG delivery after successful recovery.

  - Recorded the Payment Plan creator as the automatic release actor without a release comment.
  - Sent non-blocking ABORTED and REJECTED status notifications to Vision.
  - Preserved the normal workflow when either Vision flag is disabled.
  - Included regular, Top-Up, and Top-Up Amendment plans.
  - Excluded instruction-managed Follow-Up plans.
  - Preserved legacy item-level FC assignments and the normal manual item-selection flow.

  ## Payment Gateway behavior

  - Successful Vision release automatically starts PG delivery.
  - Vision PG failures can be retried through Django admin.
  - The React PG action remains hidden for Vision-managed plans.
  - Non-Vision plans retain their existing admin and React send/retry actions.
  - Payment Plan Group bulk PG sending skips active Vision workflows; Vision plans start PG delivery automatically after FC assignment.

  ## Migrations

  - Added BusinessArea.vision_integration_active, defaulting to False.
  - Renamed the Payment Plan admin permission to pm_manage_vision_workflow while preserving existing permission assignments.


domdinicola:
@MarekBiczysko pls fix conflicts

src/hope/contrib/vision/services.py
```diff
@@ -0,0 +1,252 @@
...
+                VisionStatus.CALLBACK_FAILED,
+                VisionErrorCode.FC_CONFLICT,
+            )
+        cls.assign_selected_funds_commitment_items(payment_plan, funds_commitment_items)
+        cls.complete_funds_commitment_assignment(payment_plan)
+
+    @classmethod
+    def can_recover_with_funds_commitment_items(cls, payment_plan: PaymentPlan) -> bool:
+        vision_status = payment_plan.vision_status
+        vision_was_sent_or_send_failed = payment_plan.sent_to_vision or vision_status == VisionStatus.SEND_FAILED.value
```
johniak: This clause makes the WAITING_FOR_CALLBACK-but-never-sent state unrecoverable. The task commits WAITING before the HTTP call (AsyncJob actions run in autocommit), so a worker killed mid-request leaves that state with no `sent` flag: no callback will come, `can_send_to_vision` is false, and this check hides the recovery button too — the plan is wedged in IN_REVIEW with only a raw DB edit as the way out. Every other status in MANUAL_RECOVERY_STATUSES already implies sent-or-send-failed, so this clause only excludes that dead end; drop it or add WAITING_FOR_CALLBACK to it.

src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -343,6 +402,9 @@ def unlock_fsp(self) -> PaymentPlan | None:
         return self.payment_plan
 
     def acceptance_process(self) -> PaymentPlan | None:
+        if self.action == PaymentPlan.Action.REVIEW.value and self.payment_plan.vision_managed:
+            raise ValidationError("Vision-managed Payment Plans are released automatically after FC assignment")
```
johniak, line 406: This guard breaks the managerial console bulk release. `PaymentPlanManagerialViewSet.bulk_action` loops all selected plans inside one `transaction.atomic`, the Pending for Release section lists every IN_REVIEW plan, and the list serializer exposes no vision fields — so one Vision-managed plan in the selection makes the whole batch 400 and roll back, with no way to tell which plan caused it. Skip Vision-managed plans in the bulk REVIEW path (as `sendable_to_payment_gateway_plans` does for group PG sends) or expose `vision_managed` on the managerial list so the FE can filter.

src/hope/contrib/vision/tasks.py
```diff
@@ -0,0 +1,80 @@
...
+
+def send_payment_plan_to_vision_async_task_action(job: AsyncJob) -> None:
+    payment_plan = PaymentPlan.objects.select_related("business_area").get(pk=job.config["payment_plan_id"])
+    if not payment_plan.can_send_to_vision:
+        return
+
+    # Persist the in-flight state before the network call so a fast callback can be correlated with this send.
+    # Abort/reject invalidates it, and the locked status checks below prevent a late response from restoring it.
+    VisionService.set_status(payment_plan, VisionStatus.WAITING_FOR_CALLBACK)
+    payment_plan.save(update_fields=["internal_data"])
```
johniak: This first write runs with no lock and no transaction, while reject/abort invalidate the attempt under `select_for_update`. If a reject commits between the `can_send_to_vision` check above and this save, the whole `internal_data` field is overwritten and WAITING_FOR_CALLBACK is restored on a plan that just left IN_REVIEW — after re-authorization the send task then silently no-ops (`can_send_to_vision` is false), leaving the exact dead state from the recovery-gate finding. Wrap this in the same select_for_update + re-check IN_REVIEW pattern as the error branch below.

src/hope/contrib/vision/services.py
```diff
@@ -0,0 +1,252 @@
...
+        if (
+            released
+            or not payment_plan.vision_integration_enabled
+            or payment_plan.status != PaymentPlan.Status.IN_REVIEW
+        ):
+            # Disabled flags and abort/reject both return the plan to the manual flow. The callback is logged by the
+            # view but must not assign FC data or release the plan.
+            return False
+        if payment_plan.vision_status != VisionStatus.WAITING_FOR_CALLBACK.value:
+            return cls.has_fc_assignment_failure(payment_plan)
```
johniak: This makes every non-WAITING status terminal for callbacks: after FC_MISSING or FC_NOT_FOUND, a follow-up callback that does carry a valid fc_num is ignored and answered 400 KO "FC not found". The ticket (Dom's answer 1E) says multiple callbacks are expected — first without an FC, a later one with it — yet the only path forward here is the admin recovery form. Consider re-running the assignment when a callback with an fc_num arrives in an FC-failure state, or get explicit business sign-off that later callbacks stay ignored.

src/hope/models/payment_plan.py
```diff
@@ -1223,11 +1224,41 @@ def can_send_to_payment_gateway(self) -> bool:
...
+        return str(vision_data.get("status") or VisionStatus.NOT_SENT.value)
+
+    @property
+    def vision_data(self) -> dict[str, Any]:
+        vision_data = (self.internal_data or {}).get("vision", {})
+        return vision_data if isinstance(vision_data, dict) else {}
+
+    @property
+    def vision_managed(self) -> bool:
+        return self.vision_integration_enabled and (
```
johniak, line 1265: Any plan already sitting in IN_REVIEW becomes vision_managed the moment both flags turn on, but the automatic send only fires on the AUTHORIZE transition. Plans authorized before the flag flip are stranded: manual release is blocked and nothing ever sends them to Vision — the only path is the per-plan admin "send to Vision" button. With a real IN_REVIEW backlog, enabling the BA flag is an incident. At minimum document "drain IN_REVIEW before enabling"; better, pick up NOT_SENT IN_REVIEW plans on enablement and queue the send.

src/hope/models/payment_plan.py
```diff
@@ -1223,11 +1224,41 @@ def can_send_to_payment_gateway(self) -> bool:
         ).exists()
         return status_accepted and has_payment_gateway_fsp and has_not_sent_to_payment_gateway_splits
 
+    @property
+    def vision_integration_enabled(self) -> bool:
+        # Instruction-managed Follow-Up plans use instruction-level delivery and do not require Vision.
+        if self.is_instruction_managed:
```
johniak: Only instruction-managed follow-ups are excluded, but plain follow-up plans are still creatable via `create_follow_up` and will enter the Vision flow. The ticket's rationale — follow-ups are failed payments whose money is already reserved — applies to all follow-ups, so a legacy follow-up now blocks in IN_REVIEW waiting for an FC that duplicates the source plan's commitment. If including them is intentional, note it in the docs; otherwise exclude `is_follow_up` plans too.

src/frontend/src/containers/pages/paymentmodule/ProgramCycle/PaymentPlanDetails/PaymentPlanDetailsPage.tsx
```diff
@@ -47,13 +55,17 @@ const PaymentPlanDetailsPage = (): ReactElement => {
       }),
     refetchInterval: (query) => {
       const data = query.state.data;
+      const visionProcessing =
```
johniak, line 58: This keeps the 3-second poll running for the whole NOT_SENT/WAITING_FOR_CALLBACK window. The 3s interval was designed for background actions that finish in minutes; a Vision callback from SAP can take hours or days, so an open tab hammers the heavy detail endpoint the entire time. Use a longer interval (30-60s) for the vision-wait case — same applies to FollowUpPaymentPlanDetailsPage.

johniak (changes requested):

src/hope/contrib/vision/services.py
MarekBiczysko: Agreed. Callback handling now reprocesses callbacks received in SEND_FAILED, CALLBACK_FAILED, FC_MISSING, and FC_NOT_FOUND. A later successful callback with a valid FC retries the assignment and continues automatic release and delivery. Terminal or inactive workflows remain log-only.

src/hope/models/payment_plan.py
MarekBiczysko: We intentionally do not enqueue the existing IN_REVIEW backlog when Vision flags are enabled. Eligible NOT_SENT plans can be sent individually through the existing Django admin “Send to Vision” action. This avoids mass external sending as a feature-flag side effect, and the behavior is documented.

src/hope/models/payment_plan.py
MarekBiczysko: Agreed. Vision is now disabled for all FOLLOW_UP Payment Plans, not only instruction-managed Follow-Ups, because their funds were already reserved for the source Payment Plan. Regular, Top-Up, and Top-Up Amendment plans remain eligible.

src/hope/models/payment_plan_group.py
```diff
@@ -134,3 +137,10 @@ def sendable_to_payment_gateway_plans(self) -> "QuerySet[PaymentPlan]":
             & Q(has_unsent_splits=True)
             & ~Q(background_action_status=PaymentPlan.BackgroundActionStatus.SEND_TO_PAYMENT_GATEWAY)
         )
+        if flag_state("VISION_INTEGRATION_ACTIVE"):
+            active_vision_statuses = [status.value for status in VisionStatus if status != VisionStatus.NOT_SENT]
+            payment_plans = payment_plans.exclude(
+                business_area__vision_integration_active=True,
+                internal_data__vision__status__in=active_vision_statuses,
+            )
```
pkujawa: I think this drops plans it should keep. `vision_status` falls back to `NOT_SENT` when the key is
missing but the SQL doesn't, so for a plan with no `vision` key the lookup comes back `NULL` - and
`NOT (TRUE AND NULL)` is `NULL`, so the row gets filtered out. 

The plans this hits are exactly the ones where `vision_managed` is `False`, i.e. the ones Vision
isn't driving and that should stay in the list. Follow-ups are the clearest case: they never get
a vision key now, so in a flag-on BA they'd drop out of the group send. Same for anything already ACCEPTED when a BA first enables the flag. So it wouldn't be possible to send them to PG.

Coalescing the missing key to `NOT_SENT` should cover both:

```
    vision_workflow_status=Coalesce(
        KeyTextTransform("status", KeyTextTransform("vision", "internal_data")),
        Value(VisionStatus.NOT_SENT.value),
        output_field=TextField(),
    )
```

Could we also get a test with `internal_data = {}`?

pkujawa (changes requested):

johniak (dismissed):

pkujawa (approved):

johniak (approved):

domdinicola (approved):

src/hope/models/payment_plan.py
```diff
@@ -1240,11 +1241,41 @@ def can_send_to_payment_gateway(self) -> bool:
...
 
+    @property
+    def vision_integration_enabled(self) -> bool:
+        # Follow-Up plans reuse funds reserved for the source plan and do not require a new Vision commitment.
+        if self.plan_type == PaymentPlan.PlanType.FOLLOW_UP:
+            return False
+        return bool(flag_state("VISION_INTEGRATION_ACTIVE")) and self.business_area.vision_integration_active
+
+    @property
+    def vision_status(self) -> str:
```
domdinicola, line 1252: good!

PR #6303, opened 2026-08-10
per-role Postgres connection options

kosior:
[AB#331051](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/331051)

1. Extract the inline `OPTIONS["options"]` strings out of `settings.py`
    - `hope.config.db_options.postgres_options(role, read_only=...)`

2. Per-role timeouts: 
    - web/runserver get 230s `statement_timeout` + `idle_in_transaction_session_timeout` 
    - every other role keeps the 4h (previously 4h everywhere) 
    - **expected side effect**: requests that today time out silently at the gateway as 504s will instead fail as Django 500 error and start appearing in Sentry 

3. New `client_connection_check_interval=5s` for web, runserver and celery so Postgres drops the backend when the worker process dies; `0` for migrate and all interactive roles

4. New `DB_CLIENT_CONNECTION_CHECK_INTERVAL_MS` env var (default `5000`); set it to `0` to disable the check

5. Read-only replica keeps `default_transaction_read_only=on` and the `-ro application_name` suffix


johniak (approved):

saxix (commented):
The issue in the linked ticket is about issues during the deployment. This changes are useful but will not help to mitigate the problem because this is not linked on existing queries.
Migrations should be designed to be cross-version compatible. There is no easy way to prevent all these types of issues. If we want to have zero-downtime upgrades, we need to handle this in the code using ETC pattern

kosior:
We split the problem into a few steps and are working through them cheapest-first:

1. This PR - limit the damage and make it visible. Web connections now time out after 230s instead of 4h, so the longest a query can block a migration drops from 4 hours to ~4 minutes.
2. Next PR - fix the slow grievance queries themselves. Already in progress. https://github.com/unicef/hope/pull/6321

PR #6306, opened 2026-08-12
Blue-green Elasticsearch reindexing for per-program indexes

johniak:
Zero-downtime blue-green reindexing for the per-program Elasticsearch indexes. Rehearsed end-to-end on eph-1: bootstrap of 292 programs + two full fleet reindexes, including the 2.1M-individual program. Combines the two stages merged to `ephemeral-1` (#6255, #6286), the rehearsal fixes (#6293) and developer docs.

## The model

- The suffix-less name every doc class addresses (`individuals_<ba>_<code>`) becomes an **alias** onto a physical `_vN` index.
- A mapping change builds a dark `_vN+1` next to the live index, then moves the program's alias pair in one atomic `_aliases` call.
- The app never notices. The old version stays unaliased as an instant rollback target.

## Commands

| Command | What it does |
|---|---|
| `es_bootstrap_aliases` | One-time: converts every bare index into `_v1` + alias via `_clone` (hard links, seconds per index). Crash-safe, resumable. |
| `es_reindex` | Build dark pair → populate → catch-up delta → count verify → atomic swap → two post-swap deltas. A verify failure aborts BEFORE the swap. |
| `es_populate_delta --target-suffix` | Catch-up pass into a dark version. |
| `es_drop_old_index_versions` | Deletes unaliased old versions after the sanity window. `--confirm` gated. |

Crash recovery in `es_reindex`:

- every index carries a `_meta` hash of the code mapping it was built from,
- a re-run of `--all` skips programs already built from the current mapping and resumes matching dark leftovers → only the remainder runs,
- `--force` rebuilds anyway; `--sweep-wrecks` deletes half-built leftovers first.

## Safety gates

| Guard | Protects against |
|---|---|
| Admin *Rebuild Index*: `confirm_action` + runs as a background `AsyncJob` (queued via `requeue`) | HTTP timeout on big programs; two concurrent rebuilds of one program |
| Administration ES panel: destructive actions need an extra checkbox | accidental live-index deletion |
| Signals and bulk consoles rerouted to `ensure_program_indexes` | automatic paths destroying a live index |
| `es_reindex`, `es_bootstrap_aliases`, `es_populate_delta`, `rebuild_program_indexes` refuse when `IS_ELASTICSEARCH_ENABLED` is off | `populate_index` no-ops on the flag → empty indexes reported as success |
| `es_drop_old_index_versions`: strict `_vN` name match + refuses while `es_reindex` holds its lock | deleting a lookalike index; deleting a mid-build dark version (no alias yet) |

## Performance (measured on eph-1)

| Change | Problem it fixes | Result |
|---|---|---|
| populate wrapped in `transaction.atomic()` | outside a txn the server-side cursor is `WITH HOLD` → Postgres materializes the WHOLE result on `DECLARE` | no more silent multi-minute hang before the first row |
| `SET LOCAL cursor_tuple_fraction = 1.0` | cursors are planned for partial reads → one random heap page read per row | bitmap heap scan: 3.3s vs 25s per 5000-row chunk |
| querysets select/prefetch everything `prepare()` touches | ~6 queries per individual dominated reindex time | prefetches filter exactly like the related managers, so index content is unchanged |
| progress tick every 1000 docs | populate looked hung for minutes | visible progress on long runs |

## Tests

Mocked tests assert the call sequences. What depends on the real API contract runs end-to-end against a live cluster:

- delta upsert/delete (caught a real bug: a `using` kwarg `Elasticsearch.bulk()` rejects),
- bootstrap: clone under a write block, atomic `remove_index` + `add`,
- reindex: the swap, and the `_meta` stamp round-trip the fleet-resume skip relies on.

## Docs

`docs/guide-dev/elasticsearch.md`: the alias model, golden rules for feature code, the sync paths, a step-by-step mapping change, what never to run, testing fixtures.

## Production notes

- check shard headroom first (blue-green doubles the index count; eph hit `cluster.max_shards_per_node=1000` mid-fleet),
- analyzer/settings-only changes are invisible to the mapping stamp → reindex with `--force` / `--sweep-wrecks`,
- ~650 docs/s at `--chunk-size 5000` → the whole production fleet fits in one night.


src/hope/apps/household/management/commands/es_populate_delta.py
```diff
@@ -0,0 +1,386 @@
...
+    def _apply_delta(delta: dict, ind_doc: type, hh_doc: type, using: str, opts: dict) -> None:
+        from hope.apps.utils.elasticsearch_utils import remove_elasticsearch_documents_by_matching_ids
+
+        chunk = opts["chunk_size"]
+        parallel = opts["parallel"]
+        # get_queryset() rather than raw model managers: same scope (program + merge status)
+        # plus the select/prefetch that keeps prepare() from doing per-row queries
+        if delta["ind_present"]:
+            qs = ind_doc().get_queryset().filter(id__in=delta["ind_present"]).iterator(chunk_size=chunk)
+            ind_doc().update(qs, action="index", parallel=parallel, using=using)
```
nerjan: `using` is passed there into the kwargs and then deeper into `Elasticsearch.bulk()` which does not recognize `using` as a keyword so it would throw an error. I think `using` can be dropped here at all, same for `hh_doc`

src/hope/admin/program.py
```diff
@@ -386,13 +387,27 @@ def check_index(self, request: HttpRequest, pk: int) -> HttpResponseRedirect:
...
+    def reindex_program(self, request: HttpRequest, pk: int) -> HttpResponse:
         program = Program.objects.get(pk=pk)
-        ok, msg = rebuild_program_indexes(str(program.id))
-        level = messages.SUCCESS if ok else messages.ERROR
-        message = "Rebuild indexes for program successful." if ok else f"Failed to rebuild indexes: {msg}"
-        messages.add_message(request, level, message)
-        return HttpResponseRedirect(reverse("admin:program_program_change", args=[pk]))
+
+        def _rebuild(request: HttpRequest) -> HttpResponseRedirect:
+            ok, msg = rebuild_program_indexes(str(program.id))
```
MarekBiczysko: this should be synchronous? no possible http timeout?

src/hope/apps/utils/elasticsearch_utils.py
```diff
@@ -16,19 +18,52 @@
...
+    parallel: bool = False,
+    chunk_size: int = 2000,
+    progress_cb: Callable[[int], None] | None = None,
+) -> None:
+    if not config.IS_ELASTICSEARCH_ENABLED:
+        return
+    # atomic() so iterator() opens a plain (lazy) server-side cursor: outside a transaction
+    # Django declares it WITH HOLD, which materializes the ENTIRE result set on DECLARE
+    # before the first row arrives - minutes-to-hours on big programs under IO pressure
+    with transaction.atomic():
```
nerjan, line 36: Doesn't this put the whole populate of one index in a single transaction? For the 2.1M-individual program that's ~an hour on one open transaction.

MarekBiczysko (dismissed):

src/hope/apps/household/management/commands/es_drop_old_index_versions.py
```diff
@@ -0,0 +1,95 @@
...
+        verb = "Dropped" if opts["confirm"] else "Would drop"
+        tail = "" if opts["confirm"] else " (re-run with --confirm to delete)"
+        self.stdout.write(self.style.SUCCESS(f"{verb} {dropped} index(es).{tail}"))
+
+    def _sweep_index(self, es: Elasticsearch, code: str, name: str, confirm: bool) -> int:
+        if not es.indices.exists_alias(name=name):
+            self.stdout.write(self.style.WARNING(f"{code}  {name}: not an alias - skipped (pre-bootstrap?)"))
+            return 0
+        current = set(es.indices.get_alias(name=name))
+        candidates = es.indices.get(index=f"{name}_v*", ignore_unavailable=True)
```
nerjan, line 79: this takes into account more than vN names, use guard from _sweep_wrecks where we check for number after _v "version_match = re.match(rf"^{re.escape(name)}_v(\d+)$", index)"

nerjan (changes requested):

src/hope/apps/household/management/commands/es_bootstrap_aliases.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def _bootstrap_index(self, es: Elasticsearch, doc_class: type) -> str:
+        name = doc_class._index._name
+        state = self._index_state(es, name)
+        if state == "alias":
+            return self._heal_aliased_index(es, name)
+        if state == "missing":
+            create_versioned_index(es, doc_class)
+            populate_index(doc_class().get_queryset(), doc_class)
+            return "created versioned index + alias, full-populated"
```
pkujawa, line 238: This command doesn't check `IS_ELASTICSEARCH_ENABLED`. Its sibling in this PR does — `es_reindex` bails out at the top with "populate would silently no-op" (es_reindex.py:142). Same thing should be applied here.                                                           
                                                                                                                                                                                                                                                                                                                             
The flag defaults to False. populate_index() returns immediately when it's off, but create_versioned_index() doesn't check anything and attaches the alias at creation time. So with the flag off this branch creates `<name>_v1`, attaches the alias, indexes nothing, and prints "created versioned index +  alias, full-populated".                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                             
`es_populate_delta` has the same gap -  _apply_delta's `update()` isn't gated, but `remove_elasticsearch_documents_by_matching_ids` is. So with the flag off it would apply every upsert and quietly skip every delete. 
(That one is hidden now because the `using` kwarg makes `update()` raise before it gets there - what @nerjan commented)                                                                         
                                                                                                                                                                                                                                                                                                                             
Could we copy the flag check into both?

src/hope/apps/household/management/commands/es_drop_old_index_versions.py
pkujawa, line 79: it can delete an index that a running `es_reindex` is still building.                                                                                                                                                                                                                 
While a reindex is in progress the new `_vN+1 has no alias on it. So it fails both safety checks here: it isn't the current alias target (that's still the old version), and it has no alias attached. It looks identical to a real leftover and gets deleted on `--confirm`.    
                                                                                                                                                                                                                                                                                                                             
Low risk in practice since the 24-72h sanity window means cleanup normally runs well after the reindex, so they shouldn't overlap. But `es_reindex` already writes `LOCK_INDEX` into ES while it runs, so checking for that here and refusing would be a couple of lines.

pkujawa:
All six new test files patch `elasticsearch.dsl.connections.get_connection`, so nothing in this PR ever talks to a real cluster.                                                                        
                                                                                                                                                                                                                                                                                                                             
That's also why the `using` bug wasn't caught. In `test_apply_delta_upserts_present_and_removes_soft_deleted` the doc classes are `MagicMock()`, and a MagicMock accepts any argument you throw at it, so `update(..., using=...)` succeeds there while it raises `TypeError` against a real ES client. The test then asserts `update.call_args.kwargs["using"] == "default"`- which needs to be dropped.                                                                   
                                                                                                                                                                                                                                                                                                                             
What needs coverage: cloning under a write block, `update_aliases` with `remove_index` + `must_exist`, the `_meta` mapping stamp surviving `indices.create` and coming back through `get_mapping`.

pkujawa (changes requested):

src/hope/admin/program.py
johniak: As you can see in diff, this issue is preexisting, but you are right this is a real issue so I will fix it in my PR

src/hope/apps/utils/elasticsearch_utils.py
johniak: Yes, deliberately. Outside a transaction Django declares the server-side cursor WITH HOLD, and Postgres then materializes the entire result set on DECLARE, that was the 20+ min hang we hit on the 2.1M program. Inside a transaction the cursor is lazy and rows stream immediately.

It's read-only, so no row locks, the only cost is holding a snapshot, same as any long SELECT.

MarekBiczysko (approved):

pkujawa (approved):

nerjan (approved):

PR #6307, opened 2026-08-12
tests: pin grievance list visibility behaviour

kosior:
For: [AB#331051](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/331051)

These pin the behaviour of `BusinessAreaVisibilityMixin.get_queryset` (`src/hope/apps/core/api/mixins.py`) before it is rewritten to use `Exists()` over the ticket -> program through model.


tests/unit/apps/grievance/test_grievance_list_global_visibility.py
```diff
@@ -0,0 +1,349 @@
...
+    with CaptureQueriesContext(connection) as three_rows:
+        client.get(list_url(afghanistan), {"limit": 3})
+    with CaptureQueriesContext(connection) as six_rows:
+        client.get(list_url(afghanistan), {"limit": 6})
+
+    assert len(six_rows.captured_queries) == len(three_rows.captured_queries)
+
+
+@pytest.mark.parametrize("permissions", [[], [Permissions.PROGRAMME_ACTIVATE]])
+def test_list_and_count_are_forbidden_without_grievance_permissions(
```
pkujawa: this test already exists - `test_grievance_ticket_global_list_without_permissions`

tests/unit/apps/grievance/test_grievance_list_global_visibility.py
```diff
@@ -0,0 +1,349 @@
...
+) -> None:
+    create_user_role_with_permissions(user, LIST_PERMISSIONS, afghanistan, program=program_one)
+
+    response = api_client(user).get(list_url(afghanistan))
+
+    assert response.status_code == status.HTTP_200_OK
+    assert [result["id"] for result in response.data["results"]] == [str(ticket_without_programs.id)]
+
+
+def test_area_limited_partner_sees_only_tickets_in_permitted_admin2(
```
pkujawa, line 209: just a note, this is partially overlapping with test_grievance_ticket_global_list_area_limits, but these 3 new area tests cover also additional cases, so I think all are beneficial. Although, we have tests with same checks in different files now

tests/unit/apps/grievance/test_grievance_list_global_visibility.py
```diff
@@ -0,0 +1,349 @@
...
+    create_user_role_with_permissions(user, LIST_PERMISSIONS, afghanistan, whole_business_area_access=True)
+    set_admin_area_limits_in_program(partner, program_one, [area_one])
+
+    response = api_client(user).get(list_url(afghanistan))
+
+    assert response.status_code == status.HTTP_200_OK
+    assert [result["id"] for result in response.data["results"]] == [str(ticket_in_program_two_area_two.id)]
+
+
+def test_user_without_program_access_gets_empty_list(
```
pkujawa: this does not test the described case.
user has permission for `whole_business_area_access=True`, so ALL programs access. Program fixture is just not loaded here, that's why the list is empty.
(Valid checks would be: permission in different program or permission not granted at all, both cases covered)

johniak (dismissed):

tests/unit/apps/grievance/test_grievance_list_global_visibility.py
kosior: Removed.

tests/unit/apps/grievance/test_grievance_list_global_visibility.py
kosior: I've added a comment at the top of the file.

tests/unit/apps/grievance/test_grievance_list_global_visibility.py
kosior: Good catch. Test renamed.

pkujawa (approved):

johniak (approved):

PR #6309, opened 2026-08-13
Scope object lookups to the business area and program from the URL

johniak:
> [!IMPORTANT]
> **Split up on request.** Eight of the fixes now live in their own PRs, one per app. This PR keeps only the shared mechanism and the serializers that use it — 9 files under `src/` instead of 26.
>
> | PR | App | Depends on |
> |---|---|---|
> | #6338 | core — field attributes, per-action permissions | — (**merge first**) |
> | #6339 | periodic_data_update — xlsx upload | — |
> | #6340 | account — partner choices | — |
> | #6341 | registration_data — import source | — |
> | #6342 | grievance — documents, bulk actions | — |
> | #6343 | token API — delegate people | — |
> | #6344 | accountability — feedback lookups | #6338 |
> | #6345 | payment — plans, verifications, documents | #6338 |
>
> What stays here: `ScopedRelatedField` itself, the four serializers that use it (grievance, accountability, beneficiary ticket, rdi upload/lax), and the contract test that fails the build when a new writable relation field is left unscoped. The contract test only passes once every app above is in, which is why it could not be split out.

## Problem

Two sources of truth that nobody compared:

| | Taken from |
|---|---|
| Permission check | business area + program in the **URL path** |
| Object | **id supplied by the caller** |

A user with a permission in business area A puts A in the path and an id from business area B in the request. Both checks pass. Advisory GHSA-2xf8-jjc2-9pmv (CWE-639).

The id reaches the object through three doors. All three were open:

| Door | Example |
|---|---|
| path | `.../payment-plans/<foreign id>/` |
| body | `{"household": "<foreign id>"}` |
| uploaded file | a template id written in a cell of the XLSX |

## What was reachable

**Ids from the path** — every action of these viewsets, on any business area:

| Viewset | |
|---|---|
| `PaymentPlanSupportingDocumentViewSet` | download / delete / upload (the advisory) |
| `PaymentPlanViewSet`, `TargetPopulationViewSet` | retrieve, patch, destroy, lock, rebuild |
| `PaymentVerificationViewSet`, `PaymentVerificationRecordViewSet` | retrieve, list, verification plan actions |
| `PaymentViewSet` | retrieve, mark as failed |
| `FeedbackViewSet` | retrieve, patch, message |

**Ids from the body**, each proven with a test:

| Request | Was |
|---|---|
| survey / message with foreign households | 201, the message was already on its way to RapidPro |
| program population import from a foreign program | 201, copying of foreign people scheduled |
| target population `copy` into a foreign cycle | 201 |
| feedback moved to a foreign program | 200 |
| grievance ticket with a foreign individual | 201 |
| grievance ticket linked to a foreign feedback | 201, **and the foreign feedback row was written to** |
| bulk update of priority on a foreign ticket | 202, **priority changed** |
| activation of a verification plan of a foreign payment plan | 200, **plan activated** |
| payment plan group created in a foreign cycle | 201 |
| target population created in a foreign cycle | not exploitable, but a 500 |

The bulk actions were the worst: they filtered tickets by id only and used the business area of the url just to clear a cache.

**Found by a second pass** over every REST endpoint, one by one:

| Request | Was |
|---|---|
| RDI upload / lax / delegate people (token API) | 201 / `accepted: 1` / `updated: 1` into a foreign business area |
| registration data import actions | 201, import data of another business area used |
| feedback with a foreign household | 201, and the feedback was pulled into the victim's program |
| update of a grievance document | 200, **document of another ticket overwritten** |
| household, individual and program reads in `account` / `core` | 200 on foreign objects |
| engine rule, financial service provider, payment details | foreign objects accepted |

**Ids from an uploaded file** — periodic data update:

- the upload read the template id out of the workbook and looked it up globally, so a file could bind the whole import to another business area's program
- the rows name people by raw uuid, also looked up globally, so one file wrote to individuals outside that program
- both proven with a test: the upload returned 202 and the victim's individual was modified

## Fix

**The mechanism**

- `ScopedRelatedField` narrows a related field to the scope of the url path, so an id outside it fails like an id that does not exist — 35 fields in grievance, 8 in accountability, plus the RDI and beneficiary ticket endpoints
- `ScopedPaymentPlanMixin` holds the payment plan lookup once, for six viewsets and for the ids in the body
- child objects are bound to their parent (`parent=self.payment_plan`), not filtered by the program on their own — otherwise an object of another plan in the same program still passes

**Lookups that took a bare id now take the scope too**

- feedback households and individuals
- grievance documents, scoped to their ticket
- registration import data
- engine rules and financial service providers, by `allowed_business_areas`
- periodic data update templates and individuals, by the program of the path
- `PaymentPlanService.create` takes the program of the path and scopes the cycle to it
- `BulkActionService` filters tickets by the business area of the path

**Two cleanups**

- `PaymentVerificationLogEntryFilter` deleted — dead code that looked up a payment plan by bare id
- `BusinessAreaProgramsAccessMixin` uses the permissions of the action instead of the class level `PERMISSIONS`, so it stops hiding objects the user is allowed to update

## Guardrail

`tests/unit/test_access_scope_contract.py` walks every REST route, takes the write actions from the router (138 endpoints), builds their serializers and fails on any writable relation field over business area or program data that is neither scoped nor on an allow list. Every entry on the list needs a reason next to it. Today there is one.

Two things worth knowing about it:

- it had a bug of its own. It matched scope names against the text of the query, and the table name `program_program` contains "program", so every field over `Program` looked scoped. Fixing that surfaced 3 more holes
- it cannot see an id typed as a plain `UUIDField` or read from an uploaded file and resolved in a service. Those need review, not a test — the periodic data update hole above is exactly that shape

## What changes for the API

- an out of scope id in the **path** returns 404, in the **body** 400, with the same message as an id that does not exist
- the scope is per program, not only per business area, because the path carries the program
- two feedback cases returned 403 before, they return 404 now

## Tests

- 59 new tests, each one checked by reverting its own fix
- 1 control test that a user without a role still gets 403
- 5 old tests updated — they built data that cannot exist in production, for example a payment plan whose program is in another business area than the plan
- full unit suite: 7677 passed, 49 skipped. `ruff` and `mypy` clean
- `test_observed_disability` fails, on `develop` as well — it came with the merge and is unrelated to this branch


pkujawa (dismissed):

Jeeziorny (dismissed):

tests/unit/test_access_scope_contract.py
```diff
@@ -0,0 +1,125 @@
+"""Guardrail for the cross business area IDOR class of bug (GHSA-2xf8-jjc2-9pmv).
+
+Permissions are checked against the business area and the program of the url path, while an object
```
Jeeziorny: I think this part is a redundant

Jeeziorny (commented):
Those tests should pass if I understand it correctly:
```
"""Exploit PoC: feedback create attaches a household from another business area.

FeedbackCreateSerializer.household_lookup is a plain UUIDField, and
FeedbackCrudServices._apply_fields resolves it with a global
get_object_or_404(Household, id=...) - no business_area / program filter.
The scoping the rest of the PR added lives in ScopedRelatedField, which this
field never uses, and the contract test only walks relation fields, so it is
blind to this service-layer lookup.

This test asserts the SECURE behaviour. It FAILS on this branch: the response
is 201 and the attacker's feedback ends up referencing the victim household and
sitting in the victim program.
"""

from typing import Any

from django.urls import reverse
import pytest
from rest_framework import status
from rest_framework.test import APIClient

from extras.test_utils.factories import (
    BusinessAreaFactory,
    HouseholdFactory,
    IndividualFactory,
    ProgramFactory,
    UserFactory,
)
from hope.apps.account.permissions import Permissions
from hope.models import BusinessArea, Feedback, Household, Program, User

pytestmark = pytest.mark.django_db


@pytest.fixture
def attacker_business_area() -> BusinessArea:
    return BusinessAreaFactory(slug="afghanistan")


@pytest.fixture
def attacker_program(attacker_business_area: BusinessArea) -> Program:
    return ProgramFactory(business_area=attacker_business_area)


@pytest.fixture
def victim_household() -> Household:
    victim_ba = BusinessAreaFactory(slug="ukraine")
    victim_program = ProgramFactory(business_area=victim_ba)
    head = IndividualFactory(
        household=None,
        program=victim_program,
        business_area=victim_ba,
    )
    return HouseholdFactory(
        program=victim_program,
        business_area=victim_ba,
        head_of_household=head,
        registration_data_import=head.registration_data_import,
    )


@pytest.fixture
def attacker(
    attacker_business_area: BusinessArea,
    attacker_program: Program,
    create_user_role_with_permissions: Any,
) -> User:
    user = UserFactory()
    create_user_role_with_permissions(
        user,
        [Permissions.GRIEVANCES_FEEDBACK_VIEW_CREATE],
        attacker_business_area,
        program=attacker_program,
    )
    return user


@pytest.fixture
def api_client(attacker: User) -> APIClient:
    client = APIClient()
    client.force_authenticate(user=attacker)
    return client


def test_create_feedback_with_foreign_household_lookup_is_denied(
    api_client: APIClient,
    attacker_business_area: BusinessArea,
    victim_household: Household,
) -> None:
    url = reverse(
        "api:accountability:feedbacks-list",
        kwargs={"business_area_slug": attacker_business_area.slug},
    )

    response = api_client.post(
        url,
        {
            "description": "cross-ba probe",
            "issue_type": "NEGATIVE_FEEDBACK",
            "consent": True,
            "household_lookup": str(victim_household.pk),
        },
        format="json",
    )

    # An id outside the path's scope must fail like an id that does not exist.
    assert response.status_code in (
        status.HTTP_400_BAD_REQUEST,
        status.HTTP_404_NOT_FOUND,
    ), response.status_code

    # And nothing may have been written that points at the foreign household.
    assert not Feedback.objects.filter(household_lookup=victim_household).exists()


def test_exploit_writes_foreign_household_and_program(
    api_client: APIClient,
    attacker_business_area: BusinessArea,
    victim_household: Household,
) -> None:
    """Demonstrator: passes on this branch, showing the concrete cross-BA write."""
    url = reverse(
        "api:accountability:feedbacks-list",
        kwargs={"business_area_slug": attacker_business_area.slug},
    )

    response = api_client.post(
        url,
        {
            "description": "cross-ba probe",
            "issue_type": "NEGATIVE_FEEDBACK",
            "consent": True,
            "household_lookup": str(victim_household.pk),
        },
        format="json",
    )

    assert response.status_code == status.HTTP_201_CREATED
    feedback = Feedback.objects.get(id=response.json()["id"])

    # Attacker's feedback lives in their own BA...
    assert feedback.business_area == attacker_business_area
    # ...yet points at the victim's household from another BA...
    assert feedback.household_lookup_id == victim_household.pk
    # ...and was dragged into the victim's program, across the BA boundary.
    assert feedback.program_id == victim_household.program_id
    assert feedback.program.business_area != attacker_business_area
```

Which suggests to me that exploit still exists

Jeeziorny (commented):
as above

src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -602,7 +602,9 @@ def create_targeting_criteria(self, targeting_criteria_input: dict, program: Pro
     @staticmethod
     def create(input_data: dict, user: "User", business_area_slug: str) -> PaymentPlan:
         business_area = BusinessArea.objects.get(slug=business_area_slug)
-        program_cycle = get_object_or_404(ProgramCycle, pk=input_data["program_cycle_id"])
+        program_cycle = get_object_or_404(
+            ProgramCycle, pk=input_data["program_cycle_id"], program__business_area=business_area
+        )
```
pkujawa: This is the one lookup in the PR still scoped to the business area rather than the program of the path.
We have a program so we can easily use it - `TargetPopulationCreateSerializer.create()` calls `self.get_program()` and then passes business_area.slug instead. Can we pass the program down?

```
    @staticmethod
    def create(input_data: dict, user: "User", program: Program) -> PaymentPlan:
        business_area = program.business_area
        program_cycle = get_object_or_404(ProgramCycle, pk=input_data["program_cycle_id"], program=program)
```
and in the serializer:

`        payment_plan = PaymentPlanService.create(input_data=data, user=request.user, program=program)`

tests/unit/test_access_scope_contract.py
```diff
@@ -0,0 +1,125 @@
...
+
+from hope.apps.core.api.fields import ScopedRelatedFieldMixin
+
+# Fields of these models never belong to a single business area.
+GLOBAL_MODELS = {"BusinessArea", "Partner", "Area", "AreaType", "Country", "User", "Role"}
+
+# Model fields that place an object inside a business area or a program.
+SCOPE_FIELDS = {"business_area", "program", "program_cycle", "payment_plan"}
+
+WRITE_ACTIONS = ("create", "update", "copy", "import", "assign", "message", "sample", "approve", "reassign", "split")
```
pkujawa: Plenty of actions that change data don't have words like "create" or "update" in the name.

34 actions with a serializer that run on POST, PATCH or DELETE get skipped. Including activate_payment_verification_plan, the action you fix. Also status_change, close_as_unique, exclude_beneficiaries, save_data and most of the other verification plan actions.

None of them is a problem today, but this file is here to catch the next one and it wouldn't.

The router puts an .actions dict on every viewset route it builds, saying which HTTP method runs which action, and it's on the same callback you already read .cls from in `_viewsets_of_rest_api`:                                                                         

```
    for method, action in getattr(pattern.callback, "actions", {}).items():
        if method in {"post", "put", "patch", "delete"}:
            found.setdefault(viewset.__name__, set()).add(action)
```

Collect that per viewset, check against it, drop WRITE_ACTIONS.

Small thing: a serializer passed to `@action(serializer_class=...)` never gets checked, since `_serializers_of` only looks at `serializer_classes_by_action` and the class-level serializer_class. Only `PaymentPlanManagerialViewSet.bulk_action` today, and its serializer has no related fields, but wirth covering.

Jeeziorny (dismissed):

Jeeziorny (dismissed):

pkujawa (dismissed):

src/hope/api/endpoints/rdi/delegate_people.py
```diff
@@ -22,13 +24,17 @@ class DelegatePeopleSerializer(serializers.Serializer):
...
             delegate_id = delegate["delegate_id"]
             delegated_for = delegate["delegated_for"]
+            if not PendingIndividual.objects.filter(id=delegate_id, registration_data_import=rdi).exists():
+                raise serializers.ValidationError(f"Delegate {delegate_id} does not belong to this import.")
             for delegated_for_id in delegated_for:
                 updated += PendingIndividualRoleInHousehold.objects.filter(
+                    household__registration_data_import=rdi,
                     household__individuals__in=[delegated_for_id],
                     individual_id=delegated_for_id,
                     role=ROLE_PRIMARY,
```
MarekBiczysko, line 51: This should be atomic, now it can leave some roles updates before raising exception.
We can also prevalidate all delegates before updating roles.

`
      def validate_delegates(self, delegates: list[dict]) -> list[dict]:
          rdi = self.context["registration_data_import"]
          delegate_ids = {delegate["delegate_id"] for delegate in delegates}
          valid_delegate_ids = set(
              PendingIndividual.objects.filter(
                  id__in=delegate_ids,
                  registration_data_import=rdi,
              ).values_list("id", flat=True)
          )

          for delegate in delegates:
              delegate_id = delegate["delegate_id"]
              if delegate_id not in valid_delegate_ids:
                  raise serializers.ValidationError(
                      f"Delegate {delegate_id} does not belong to this import."
                  )

          return delegates

      @transaction.atomic
      def create(self, validated_data: dict) -> dict:
          rdi = self.context["registration_data_import"]
          delegates = validated_data.pop("delegates")
          updated = 0

          for delegate in delegates:
              for delegated_for_id in delegate["delegated_for"]:
                  updated += PendingIndividualRoleInHousehold.objects.filter(
                      household__registration_data_import=rdi,
                      household__individuals__in=[delegated_for_id],
                      individual_id=delegated_for_id,
                      role=ROLE_PRIMARY,
                  ).update(individual_id=delegate["delegate_id"])

          return {"updated": updated}

  This also replaces one .exists() query per delegate with one query for the whole batch.
`

src/hope/apps/payment/api/serializers.py
MarekBiczysko, line 1790: please make this atomic

  Current failure order is:

  1. Existing targeting rules are deleted.
  2. New rules are created and the payment plan is saved.
  3. _update_fsp_and_delivery_mechanism() looks up the FSP.
  4. The foreign FSP causes Http404.
  5. Earlier writes remain committed.

MarekBiczysko (changes requested):

Jeeziorny (dismissed):

ntrncic:
Can we split this PR by different sections. RDI, Grievance etc

johniak:
Split it up — one PR per app, #6338 … #6345, listed in the description.

Two things came out of doing it that are worth flagging:

**#6338 has to merge first.** `BusinessAreaProgramsAccessMixin` narrowed the queryset by `self.PERMISSIONS`, which is the *list* permission, so detail and write actions were narrowed against the wrong permission set. I had this filed as a cleanup; it is not. Without it the feedback update endpoint answers 404 for a legitimate request, so #6344 and #6345 are stacked on it. GitHub retargets their base to `develop` once it merges.

**The contract test could not come out.** `tests/unit/test_access_scope_contract.py` walks every write endpoint and fails on any writable relation field that is not narrowed to the url path. Measured against a clean `develop` with only the mechanism added, that is **63 findings across 8 viewsets**. Shipping it early would mean merging a 62-entry allowlist into a public repo — a machine-readable map of exactly which endpoints are still exploitable, sitting there for the length of the sequence. So it stays here with the mechanism and lands when the last app is in.

Every branch was tested on its own base, not just as part of the whole. That is how the mixin dependency above surfaced.

Jeeziorny:
@domdinicola do you want me to go through every PR from new split and approve it explicitly? How you'd like to do it?

johniak:
Superseded by the stacked series: #6353 (core, ScopedRelatedField) ← #6354 (grievance) ← #6355 (token API) ← #6356 (accountability) ← #6357 (payment + contract test), plus three independent PRs #6350 (periodic data update), #6351 (account), #6352 (registration data). Content is byte-identical to this branch (verified blob-by-blob, 51/51 files); the bulk action test removed in 9ed19bd137 is restored in #6354. This PR's branch stays untouched as reference.

PR #6312, opened 2026-08-14
fix(account): filter role assignment inline business areas by partner allowed areas

domdinicola:
## Summary
The RoleAssignmentInline on the Partner admin listed **all** business areas, even ones the partner is not allowed in, and submitting a non-allowed area caused a **500**.

## Root causes
1. `AutocompleteForeignKeyMixin` auto-switched the inline's `business_area` field to an autocomplete widget. The autocomplete endpoint ignores the form field's queryset (limited to `partner.allowed_business_areas`), so the dropdown showed every business area.
2. `RoleAssignment.clean()` accessed `self.business_area.id`. When a non-allowed area is submitted, the field fails validation, leaving `business_area` unset; since the FK is non-nullable, accessing it raised `RelatedObjectDoesNotExist` (unhandled → 500) instead of a form error.

## Changes
- `src/hope/admin/user_role.py`: exclude `business_area` from `get_autocomplete_fields` in `RoleAssignmentInline` so the dropdown renders a plain select limited to allowed areas.
- `src/hope/models/role_assignment.py`: guard `RoleAssignment.clean()` with `self.business_area_id` so invalid submissions produce a validation error instead of crashing.
- `tests/unit/apps/account/test_admin.py`: added tests for the non-autocomplete inline field, and for formset POSTs with allowed/disallowed business areas.

## Verification
- `tox -e lint` (pre-commit --all-files): passed
- `tox -e mypy` (mypy src): success, 758 files
- `tests/unit/apps/account/test_admin.py` + `test_models.py`: all pass

### DIFF
--- a/src/hope/admin/user_role.py
+++ b/src/hope/admin/user_role.py
@@ -23,6 +23,12 @@ class RoleAssignmentInline(AutocompleteForeignKeyMixin, admin.TabularInline):
     formset = RoleAssignmentInlineFormSet
     ordering = ["business_area__name"]
 
+    def get_autocomplete_fields(self, request: HttpRequest) -> list[str]:
+        # business_area is intentionally excluded: its choices are restricted to the
+        # partner's allowed_business_areas, and the autocomplete widget would bypass
+        # that restriction by listing every business area.
+        return [field for field in super().get_autocomplete_fields(request) if field != "business_area"]
+
     def formfield_for_foreignkey(self, db_field: Any, request: Any = None, **kwargs: Any) -> Any:
         partner_id = request.resolver_match.kwargs.get("object_id")
 
--- a/src/hope/admin/user_role.py
+++ b/src/hope/admin/user_role.py
@@ -27,7 +27,9 @@ class RoleAssignmentInline(AutocompleteForeignKeyMixin, admin.TabularInline):
         # business_area is intentionally excluded: its choices are restricted to the
         # partner's allowed_business_areas, and the autocomplete widget would bypass
         # that restriction by listing every business area.
-        return [field for field in super().get_autocomplete_fields(request) if field != "business_area"]
+        # role is excluded because for regular partners the choices are restricted to
+        # roles with is_available_for_partner=True, which autocomplete would bypass.
+        return [field for field in super().get_autocomplete_fields(request) if field not in ("business_area", "role")]
 
     def formfield_for_foreignkey(self, db_field: Any, request: Any = None, **kwargs: Any) -> Any:
         partner_id = request.resolver_match.kwargs.get("object_id")
@@ -59,6 +61,11 @@ class RoleAssignmentInline(AutocompleteForeignKeyMixin, admin.TabularInline):
 
 class BaseRoleAssignmentAdmin(HOPEModelAdminBase):
     form = RoleAssignmentAdminForm
+
+    def get_autocomplete_fields(self, request: HttpRequest) -> list[str]:
+        # business_area is excluded because formfield_for_foreignkey restricts it to
+        # is_split=False, and the autocomplete widget would bypass that restriction.
+        return [field for field in super().get_autocomplete_fields(request) if field != "business_area"]
 
     def get_queryset(self, request: HttpRequest) -> QuerySet:
         return (
@@ -124,6 +131,12 @@ class PartnerRoleAssignmentAdmin(BaseRoleAssignmentAdmin):
         ("program", AutoCompleteFilter),
         ("role", AutoCompleteFilter),
     )
+
+    def get_autocomplete_fields(self, request: HttpRequest) -> list[str]:
+        # role is excluded because formfield_for_foreignkey restricts it to
+        # is_available_for_partner=True for regular partners, and the autocomplete
+        # widget would bypass that restriction.
+        return [field for field in super().get_autocomplete_fields(request) if field != "role"]
 
     def get_queryset(self, request: HttpRequest) -> QuerySet:
         qs = super().get_queryset(request)
--- a/tests/unit/apps/account/test_admin.py
+++ b/tests/unit/apps/account/test_admin.py
@@ -306,6 +306,21 @@ def test_role_assignment_inline_business_area_not_autocomplete(
     assert list(field.queryset) == [business_area_afg]
 
 
+def test_role_assignment_inline_role_not_autocomplete(
+    request_factory: RequestFactory,
+    partner: Partner,
+    role_available_for_partner: Role,
+    role_not_available_for_partner: Role,
+):
+    from django.contrib import admin
+
+    request = get_mock_request(request_factory, object_id=partner.id)
+
+    model_admin = RoleAssignmentInline(parent_model=Partner, admin_site=admin.site)
+    fields = model_admin.get_autocomplete_fields(request)
+    assert "role" not in fields
+
+
 def test_role_assignment_inline_formfield_for_foreignkey_role(
     request_factory: RequestFactory,
     admin_site: AdminSite,
@@ -340,6 +355,44 @@ def test_role_assignment_inline_formfield_for_foreignkey_role_regular_partner(
     field = admin.formfield_for_foreignkey(RoleAssignment._meta.get_field("role"), request)
     assert role_available_for_partner in field.queryset
     assert role_not_available_for_partner not in field.queryset
+
+
+def test_user_role_assignment_admin_business_area_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    staff_user: User,
+    business_area_afg: BusinessArea,
+    role_1: Role,
+):
+    admin = UserRoleAssignmentAdmin(model=RoleAssignment, admin_site=admin_site)
+
+    request = get_mock_request(request_factory, user=staff_user)
+    fields = admin.get_autocomplete_fields(request)
+    assert "business_area" not in fields
+
+
+def test_partner_role_assignment_admin_role_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    staff_user: User,
+):
+    admin = PartnerRoleAssignmentAdmin(model=RoleAssignment, admin_site=admin_site)
+
+    request = get_mock_request(request_factory, user=staff_user)
+    fields = admin.get_autocomplete_fields(request)
+    assert "role" not in fields
+
+
+def test_partner_role_assignment_admin_business_area_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    staff_user: User,
+):
+    admin = PartnerRoleAssignmentAdmin(model=RoleAssignment, admin_site=admin_site)
+
+    request = get_mock_request(request_factory, user=staff_user)
+    fields = admin.get_autocomplete_fields(request)
+    assert "business_area" not in fields
 
 
 def test_role_assignment_inline_has_permissions(
--- a/tests/unit/apps/account/test_admin.py
+++ b/tests/unit/apps/account/test_admin.py
@@ -285,6 +285,25 @@ def test_role_assignment_inline_formfield_for_foreignkey_business_area(
     assert business_area_ukr in field.queryset
 
 
+def test_role_assignment_inline_business_area_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    business_area_afg: BusinessArea,
+    business_area_ukr: BusinessArea,
+    partner: Partner,
+):
+    partner.allowed_business_areas.add(business_area_afg)
+
+    request = get_mock_request(request_factory, object_id=partner.id)
+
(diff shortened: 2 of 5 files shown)


src/hope/admin/user_role.py
```diff
@@ -23,6 +23,12 @@ class RoleAssignmentInline(AutocompleteForeignKeyMixin, admin.TabularInline):
     formset = RoleAssignmentInlineFormSet
     ordering = ["business_area__name"]
 
+    def get_autocomplete_fields(self, request: HttpRequest) -> list[str]:
```
pkujawa: Actually, role has the same problem.
It's filtered by `is_available_for_partner=True` and autocomplete broke it.
```

    _NON_AUTOCOMPLETE_FIELDS = frozenset({"business_area", "role"})

    def get_autocomplete_fields(self, request: HttpRequest) -> list[str]:
        return [f for f in super().get_autocomplete_fields(request) if f not in self._NON_AUTOCOMPLETE_FIELDS]
```

tests/unit/apps/account/test_admin.py
```diff
@@ -285,6 +285,25 @@ def test_role_assignment_inline_formfield_for_foreignkey_business_area(
...
+    business_area_afg: BusinessArea,
+    business_area_ukr: BusinessArea,
+    partner: Partner,
+):
+    partner.allowed_business_areas.add(business_area_afg)
+
+    request = get_mock_request(request_factory, object_id=partner.id)
+
+    admin = RoleAssignmentInline(parent_model=Partner, admin_site=admin_site)
+    assert "business_area" not in admin.get_autocomplete_fields(request)
```
pkujawa: This assertion can't fail. The `admin_site` fixture is `AdminSite()` with an empty `_registry`, with nothing registered, `get_autocomplete_fields` returns []

The mixin's tests hit this and use the real site for it, e.g. tests/unit/apps/core/test_admin.py under the "AutocompleteForeignKeyMixin":

```
    model_admin = HouseholdAdmin(Household, admin.site)
    fields = model_admin.get_autocomplete_fields(request)
    assert "business_area" in fields
```
Can you apply same way?

src/hope/admin/user_role.py
pkujawa: Same bug here, not fixed.  Filters `business_area` to `is_split=False`, but business_area is autocomplete -issue on both Role Assignment (User) and Role Assignment (Partner).

src/hope/admin/user_role.py
```diff
@@ -23,6 +23,12 @@ class RoleAssignmentInline(AutocompleteForeignKeyMixin, admin.TabularInline):
     formset = RoleAssignmentInlineFormSet
```
pkujawa: Actually, UserAdmin has the same bug.
It restricts the partner to non-parent partners, but partner is also an autocomplete field.
Can we fix as well? Exclude Partner from autocomplete in UserAdmin

src/hope/admin/user_role.py
```diff
@@ -23,6 +23,12 @@ class RoleAssignmentInline(AutocompleteForeignKeyMixin, admin.TabularInline):
...
     ordering = ["business_area__name"]
 
+    def get_autocomplete_fields(self, request: HttpRequest) -> list[str]:
+        # business_area is intentionally excluded: its choices are restricted to the
+        # partner's allowed_business_areas, and the autocomplete widget would bypass
+        # that restriction by listing every business area.
+        return [field for field in super().get_autocomplete_fields(request) if field != "business_area"]
```
pkujawa: role has the same problem.
It's filtered by `is_available_for_partner=True` for Partner Role Assignments
->
```
    def get_autocomplete_fields(self, request: HttpRequest) -> list[str]:
        return [
            field
            for field in super().get_autocomplete_fields(request)
            if field not in ("business_area", "role")
        ]
```
here

AND also needs excluding `role` from standard PartnerRoleAssignmentAdmin

src/hope/admin/user_role.py
pkujawa, line 87: Same bug here.  Filters `business_area` to `is_split=False`, but business_area is autocomplete - issue on both Role Assignment (User) and Role Assignment (Partner).
It will fixed both when applied here

tests/unit/apps/account/test_admin.py
pkujawa: This assertion can't fail. The `admin_site` fixture is `AdminSite()` with an empty `_registry`, with nothing registered, `get_autocomplete_fields` returns []

The mixin's tests hit this and use the real site for it, e.g. tests/unit/apps/core/test_admin.py under the "AutocompleteForeignKeyMixin":

```
    model_admin = HouseholdAdmin(Household, admin.site)
    fields = model_admin.get_autocomplete_fields(request)
    assert "business_area" in fields
```
Can you apply same way?

pkujawa (changes requested):
Actually, `UserAdmin` has the same bug.
It restricts the `partner` to non-parent partners, but partner is also an autocomplete field.
Can we fix as well? Exclude `partner` from autocomplete in `UserAdmin`

pkujawa:
Looks good but could you add tests for these new exclusions?
The role exclusion on the inline, business_area on the base, role on PartnerRoleAssignmentAdmin, and partner on UserAdmin are not checked

tests/unit/apps/account/test_admin.py
```diff
@@ -306,6 +306,21 @@ def test_role_assignment_inline_business_area_not_autocomplete(
     assert list(field.queryset) == [business_area_afg]
 
 
+def test_role_assignment_inline_role_not_autocomplete(
+    request_factory: RequestFactory,
+    partner: Partner,
+    role_available_for_partner: Role,
+    role_not_available_for_partner: Role,
+):
+    from django.contrib import admin
```
pkujawa: can we move at the top of the file? It's imported twice now in this file inside different tests

tests/unit/apps/account/test_admin.py
```diff
@@ -342,6 +357,44 @@ def test_role_assignment_inline_formfield_for_foreignkey_role_regular_partner(
...
 
 
+def test_user_role_assignment_admin_business_area_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    staff_user: User,
+    business_area_afg: BusinessArea,
+    role_1: Role,
+):
+    admin = UserRoleAssignmentAdmin(model=RoleAssignment, admin_site=admin_site)
```
pkujawa: same issue I raised for previous tests - empty registry for admin_site, so this doesn't test anything

tests/unit/apps/account/test_admin.py
```diff
@@ -342,6 +357,44 @@ def test_role_assignment_inline_formfield_for_foreignkey_role_regular_partner(
...
+    fields = admin.get_autocomplete_fields(request)
+    assert "role" not in fields
+
+
+def test_partner_role_assignment_admin_business_area_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    staff_user: User,
+):
+    admin = PartnerRoleAssignmentAdmin(model=RoleAssignment, admin_site=admin_site)
```
pkujawa: same

tests/unit/apps/account/test_admin.py
```diff
@@ -342,6 +357,44 @@ def test_role_assignment_inline_formfield_for_foreignkey_role_regular_partner(
...
+    fields = admin.get_autocomplete_fields(request)
+    assert "business_area" not in fields
+
+
+def test_partner_role_assignment_admin_role_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    staff_user: User,
+):
+    admin = PartnerRoleAssignmentAdmin(model=RoleAssignment, admin_site=admin_site)
```
pkujawa: same

tests/unit/apps/account/test_admin.py
```diff
@@ -342,6 +357,44 @@ def test_role_assignment_inline_formfield_for_foreignkey_role_regular_partner(
     assert role_not_available_for_partner not in field.queryset
 
 
+def test_user_role_assignment_admin_business_area_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    staff_user: User,
+    business_area_afg: BusinessArea,
+    role_1: Role,
```
pkujawa: not sure why these fixtures are used here

tests/unit/apps/account/test_admin.py
```diff
@@ -306,6 +306,21 @@ def test_role_assignment_inline_business_area_not_autocomplete(
     assert list(field.queryset) == [business_area_afg]
 
 
+def test_role_assignment_inline_role_not_autocomplete(
+    request_factory: RequestFactory,
+    partner: Partner,
+    role_available_for_partner: Role,
+    role_not_available_for_partner: Role,
```
pkujawa: these fixtures are not needed?

tests/unit/apps/account/test_admin.py
```diff
@@ -285,6 +285,42 @@ def test_role_assignment_inline_formfield_for_foreignkey_business_area(
...
+    assert list(field.queryset) == [business_area_afg]
+
+
+def test_role_assignment_inline_role_not_autocomplete(
+    request_factory: RequestFactory,
+    partner: Partner,
+    role_available_for_partner: Role,
+    role_not_available_for_partner: Role,
+):
+    from django.contrib import admin
```
pkujawa: can we move to the top of the file? It's imported twice now in this file inside different tests

tests/unit/apps/account/test_admin.py
```diff
@@ -321,6 +357,44 @@ def test_role_assignment_inline_formfield_for_foreignkey_role_regular_partner(
...
 
 
+def test_user_role_assignment_admin_business_area_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    staff_user: User,
+    business_area_afg: BusinessArea,
+    role_1: Role,
+):
+    admin = UserRoleAssignmentAdmin(model=RoleAssignment, admin_site=admin_site)
```
pkujawa, line 360: same issue from the initial review - empty registry for admin_site, so this doesn't test anything

tests/unit/apps/account/test_admin.py
```diff
@@ -321,6 +357,44 @@ def test_role_assignment_inline_formfield_for_foreignkey_role_regular_partner(
...
+    fields = admin.get_autocomplete_fields(request)
+    assert "business_area" not in fields
+
+
+def test_partner_role_assignment_admin_role_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    staff_user: User,
+):
+    admin = PartnerRoleAssignmentAdmin(model=RoleAssignment, admin_site=admin_site)
```
pkujawa, line 372: same - admin_site empty

tests/unit/apps/account/test_admin.py
```diff
@@ -321,6 +357,44 @@ def test_role_assignment_inline_formfield_for_foreignkey_role_regular_partner(
...
+    fields = admin.get_autocomplete_fields(request)
+    assert "role" not in fields
+
+
+def test_partner_role_assignment_admin_business_area_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    staff_user: User,
+):
+    admin = PartnerRoleAssignmentAdmin(model=RoleAssignment, admin_site=admin_site)
```
pkujawa, line 384: same as above

tests/unit/apps/account/test_admin.py
```diff
@@ -285,6 +285,42 @@ def test_role_assignment_inline_formfield_for_foreignkey_business_area(
...
+    form_set = model_admin.get_formset(request, partner)
+    field = form_set.form.base_fields["business_area"]
+    assert list(field.queryset) == [business_area_afg]
+
+
+def test_role_assignment_inline_role_not_autocomplete(
+    request_factory: RequestFactory,
+    partner: Partner,
+    role_available_for_partner: Role,
+    role_not_available_for_partner: Role,
```
pkujawa: these fixtures are not needed?

tests/unit/apps/account/test_admin.py
```diff
@@ -321,6 +357,44 @@ def test_role_assignment_inline_formfield_for_foreignkey_role_regular_partner(
     assert role_not_available_for_partner not in field.queryset
 
 
+def test_user_role_assignment_admin_business_area_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+    staff_user: User,
+    business_area_afg: BusinessArea,
+    role_1: Role,
```
pkujawa: I don't think these 2 fixtures are needed in this test

pkujawa (changes requested):

pkujawa (approved):

nerjan (approved):
Same problem exists in Partner for parent and Projects programme. It could be good idea to create mixin with "autocomplete_exclude_fields" and use it in this places, as it might repeat in the future.

PR #6315, opened 2026-08-17
[FEATURE] 331733: todo review

nerjan:
[AB#331733](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/331733)

Fixed TODOS or delete TODO if already done:
520c4190eb — Fixed # TODO: if passed a dict, it would go into infinite loop in the steficon jsonfy templatetag (plus dropped a stale React Query cache-invalidation TODO in EditVerificationPlan.tsx).
a706bf14a4 — Resolved # TODO: add 'program_id' arg or None? by logging the individual XLSX update with the uploader user and program.
cf68491c4f — Removed the PaymentPlan.PREPARING status marked "deprecated will remove it after data migrations", with the data migration.
1080bbe54b — Removed role_reassign_data from individual data update tickets, marked "deprecated will be removed in next release".
60eb49d9de — Applied # TODO: set to not nullable Program and on_delete=models.PROTECT to RegistrationDataImport.program.
77746ece00 (+ 764fb4b08d) — Resolved the # type: ignore # FIXME typing suppressions across admin, kobo API, steficon forms, grievance models and sentry middleware.
dc9dd9fd4c — Deleted the # TODO: this one is not added in snapshot markers on fields that genuinely aren't snapshotable.
86388528c1 — Cleared the # TODO: test needed / # TODO: not sure if this one is correct? in _prepare_kobo_asset_id_value by adding tests and dropping pragma: no cover.
609a286250 — Answered # TODO: not sure about deduplication for people?? — deduplication does run for people.
ce9603dd6a — Answered # TODO: update flex field for People in the RDI validators — current behaviour is correct.
9d83cf17c0 — Deleted calculate_children_fields_for_not_collected_individual_data_async_task_action and its # TODO: count differently or add all the fields for the new gender options, as it was unused.
7677106daa — Implemented # TODO: SANCTION LIST CHECK PER LIST as a per-program last-check key with possible-match reset.
3000c79c65 — Removed the run_from_migration parameter marked # TODO remove after migration from recalculate_data.
517da23c65 — Implemented # TODO exclude PDU here — PDU attributes no longer appear in FSP XLSX template flex-field choices.
3dfb7e4b88 — Fixed the two cache TODOs: the "bad approach, Count and Max are very expensive" list cache keys (now signal-driven versions) and "Enable cache on verification list".
9243bc1fcc — Deleted the stale # TODO: e2e failed probably because of cache here.
07b074b1b6 — Resolved the three // TODO Marcin make Type to this function in targetingUtils.ts.
90c3c773c6 (+ cb24c4553a) — Replaced the skipped ToDo: Filters feedback e2e class with unit tests, which surfaced and fixed the broken program/search filters; also removed sleep(0.2) # TODO: added just for test in CI.
e22f5f106d — Fixed # TODO: refactor the output of this function in the individual XLSX update service.
5c33983838 — Deleted the #FIXME: on CI we have 7 here instead of 5 query-count comments in the program list tests.
fec5933952 — Removed the partner-access "temporary solution" TODOs, since that solution is permanent.
4b319e2f2a — Removed TP_PROCESSING = "PROCESSING" # TODO: do we need this one? and migrated legacy rows to DRAFT.
1870b7c863 — Fixed the remaining # TODO: this one is not added in snapshot by adding identities to the payment snapshot so UNHCR/SCOPE ID columns export again.


MarekBiczysko (dismissed):

src/hope/models/registration_data_import.py
```diff
@@ -160,14 +160,11 @@ class RegistrationDataImport(TimeStampedUUIDModel, ConcurrencyModel, AdminUrlMix
...
     business_area = models.ForeignKey(BusinessArea, null=True, blank=True, on_delete=models.CASCADE)
-    # TODO: set to not nullable Program and on_delete=models.PROTECT
     program = models.ForeignKey(
         "program.Program",
-        null=True,
-        blank=True,
         db_index=True,
         related_name="registration_imports",
-        on_delete=models.SET_NULL,
+        on_delete=models.PROTECT,
```
johniak, line 167: `PROTECT` is the right call here — but deleting a draft program that has an RDI now returns a 500, because `ProgramViewSet.perform_destroy` (`program/api/views.py:394`) doesn't handle `ProtectedError`. Please catch it and return a friendly 400, e.g. "This programme has registration data imports and cannot be deleted." The frontend needs no change: `DeleteProgram.tsx` already pipes error responses through `showApiErrorMessages`, so the 400 message will show up in the snackbar as-is.

johniak (changes requested):

src/hope/apps/account/api/serializers.py
```diff
@@ -241,8 +241,6 @@ def get_partner_choices(self, *args: Any, **kwargs: Any) -> list[dict[str, Any]]
         )
 
     def get_partner_choices_temp(self, *args: Any, **kwargs: Any) -> list[dict[str, Any]]:
-        # TODO: can be removed after proper solution is applied; this is the temp solution to skip the user input in
-        #  program mutations and retrieve partners already with a role in BA
```
kosior, line 270: @nerjan Is it OK to remove this TODO? Has the "proper solution" been implemented and applied? If yes, then could we remove the `get_partner_choices_temp`?

I haven't reviewed the whole PR, but it's related to changes I'm making on a different ticket.

src/hope/models/registration_data_import.py
nerjan: There is actually no way to delete Program really, it always is soft deleted, unless for root users from admin where it has nice information about protected fields.

src/hope/apps/account/api/serializers.py
nerjan: Its temporary solution for over 1.5 year so its not actually temporary at all and would probably stay that way much longer.

tests/unit/apps/sanction_list/test_check_against_sanction_list_pre_merge.py
```diff
@@ -169,6 +170,80 @@ def test_execute(program, sanction_list, household_with_individuals, national_id
     assert result == expected
 
 
+@override_config(SANCTION_LIST_MATCH_SCORE=3.5)
+@override_config(IS_ELASTICSEARCH_ENABLED=True)
+def test_full_run_stores_last_check_for_its_own_program(
```
johniak, line 197: These four new tests build their data in the test body (`ProgramFactory`, `HouseholdFactory`, `Individual.objects.get(...)` + `save()`) instead of in fixtures. Could the "other program" and "stale individual" setup move into local fixtures so the tests stay Arrange-Act-Assert?

johniak (changes requested):

tests/unit/apps/sanction_list/test_check_against_sanction_list_pre_merge.py
nerjan: fixed!

johniak (dismissed):

kosior (dismissed):

domdinicola:
@nerjan pls resolve conflicts again

src/hope/apps/registration_data/tasks/rdi_xlsx_people_create.py
```diff
@@ -403,7 +403,6 @@ def execute(
 
         old_rdi_mis = RegistrationDataImport.objects.get(id=registration_data_import.id)
         if not self.business_area.postpone_deduplication:
-            # TODO: not sure about deduplication for people??
```
domdinicola, line 406: what's up here?

src/hope/apps/registration_data/validators.py
```diff
@@ -338,7 +338,6 @@ def get_combined_fields(self) -> dict:
             if not self.is_social_worker_program
             else FieldFactory.from_scopes([Scope.XLSX_PEOPLE])
         )
-        # TODO: update flex field for People
```
domdinicola, line 341: what's up here?

src/hope/apps/registration_data/tasks/rdi_xlsx_people_create.py
nerjan: removed because dedup does run for people

src/hope/apps/registration_data/validators.py
nerjan: it was already updated.

PR #6317, opened 2026-08-18
ActivityLog for Groups, flaky e2e fix, lock on PP actions

pkujawa:
- fix flaky e2e `test_household_data_update_admin_area_resolves_pcode_to_label` 
    Fix flaky dropdown clicks in the e2e helper - the option is clicked on the element and retried if the menu stays open. Covers 8 e2e files, including test_household_data_update_admin_area_resolves_pcode_to_label.
- apply changes found in review for https://github.com/unicef/hope/pull/6222 for pre-existing issues:
    - Payment plan rows are now locked before the status is checked, in acceptance                                                                                                                                                                                                                                             
      process, mark ready for closure, close and send back to finished. Without the lock two                                                                                                                                                                                                                                          
      people acting at the same time could both move the plan forward and both trigger                                                                                                                                                                                                                                         
      the email.                                                                                                                                                                                                                                                                                                               
    - The managerial bulk action endpoint now only accepts Approve, Authorize and                                                                                                                                                                                                                                              
      Review. It used to advertise all 18 actions while supporting three, so the rest                                                                                                                                                                                                                                          
      were accepted by validation and then came as 403 because of no permission mapped.
      The app only ever sends these 3, so nothing changes for users.                                                                                                                                                                                                                                                                                     
    - The payment notification tests no longer use loops. Each test now compares the                                                                                                                                                                                                                                           
      full list of recipients in one assertion.                                                                                                                                                                                                                                             
    - Query counts in tests for closure calls and on recipient resolution
- groups enhancement:
    - Activity log for payment plan groups.                                                                                                                                                                                                                                                                               
    - Activity Log table added to the group details page, behind the activity log permission.                                                                                                                                                                                                                                  
    - The group details header now shows the background action status while a group is busy. Before this the only clue was that the XLSX buttons were disabled.                                                                                                                                                                
    - Fixed a pre-existing bug: the activity log pager counted all logs in the business area, not just
  the ones for the specific record. This affects every page with an activity log


src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -343,31 +343,35 @@ def unlock_fsp(self) -> PaymentPlan | None:
...
-            "approval_process": approval_process,
-            "created_by": self.user,
-            "type": approval_type,
-            "comment": approval_comment,
-        }
-        Approval.objects.create(**approval_data)
-        log_payment_plan_approval(self.payment_plan, self.user, approval_type, approval_comment)
+        with transaction.atomic():
+            self.payment_plan = (
+                PaymentPlan.objects.select_for_update().select_related("program_cycle").get(pk=self.payment_plan.pk)
```
johniak: This `select_for_update()` has no `of=`, so Postgres also locks the joined `ProgramCycle` row for the whole transaction. Every approve/authorize/review across all plans in a cycle now serializes on that one row, and `bulk_action` holds these locks across its whole loop inside one outer `transaction.atomic` — two concurrent bulk requests over same-cycle plans can deadlock and 500 (T1 approves [A, B] and holds the cycle via A; T2 approves [B] and waits on the cycle; T1 then wants B). `select_for_update(of=("self",))` keeps the join but locks only the plan row — same pattern as `registration_data/tasks/rdi_merge.py:223`.

johniak (changes requested):

johniak (approved):

nerjan (approved):

PR #6318, opened 2026-08-18
[333803] USDC wallet image attachments

pkujawa:
[AB#333803](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/333803)                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                             
  The wallet number image moves off the individual's flex fields and onto the account.                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                    
  - New `AccountAttachment` model, same shape as `PaymentPlanSupportingDocument`. Max 10 files per                                                                                                                                                                                                                               
    account, 10 MB each.                                                                                                                                                                                                                                                                                                     
  - New USDC imports save the image as an attachment. `id_wallet_image_i_f` stays a flex field.                                                                                                                                                                                                                                
  - RDI upload API accepts attachments under accounts. Individual detail returns them under each                                                                                                                                                                                                                             
    account, behind the existing delivery mechanisms permission.                                                                                                                                                                                                                                                             
  - New upload, delete and download endpoints under                                                                                                                                                                                                                                                                          
    programs/{program_code}/individuals/{individual_pk}/accounts/{account_pk}/attachments/                                                                                                                                                                                                                                   
  - The individual page shows attachments in the accounts section.                                                                                                                                                                                                                                                           
               
               
Everything is implemented as the ticket describes, except: 
- Backfill is a one-time script, not a data migration.
- File size and count are also validated on the model's `save()`. RDI upload, Aurora parser and the admin all create attachments without going through the serializer mentioned in the ticket. 

                                                                                                                                                                                                                
  Existing images are moved by a one time script.                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                             
```
      from hope.one_time_scripts.migrate_wallet_images_to_account_attachments import (                                                                                                                                                                                                                                       
          migrate_wallet_images_to_account_attachments,                                                                                                                                                                                                                                                                      
      )                                                                                                                                                                                                                                                                                                                      
      migrate_wallet_images_to_account_attachments()             
```


src/hope/admin/account_attachment.py
```diff
@@ -0,0 +1,22 @@
...
+from django.contrib import admin
+from django.db.models import QuerySet
+from django.http import HttpRequest
+
+from hope.admin.utils import HOPEModelAdminBase
+from hope.models import AccountAttachment
+
+
+@admin.register(AccountAttachment)
+class AccountAttachmentAdmin(HOPEModelAdminBase):
```
MarekBiczysko, line 11: Account admin inline?

src/hope/models/account_attachment.py
```diff
@@ -0,0 +1,45 @@
+from typing import Any
+
+from django.conf import settings
+from django.core.exceptions import ValidationError
+from django.db import models
+
+from hope.models.account import Account
+
+
+class AccountAttachment(models.Model):
```
MarekBiczysko, line 19: shouldn't this be a merge status model?

MarekBiczysko (dismissed):

src/hope/models/account_attachment.py
pkujawa: Not sure we need it. It is only read through account.attachments, and the account already carries the status, so the attachment inherits visibility from it. Unless you see a case where the account's status is not enough?

MarekBiczysko: nope, just double checking if it was not missed unintentionally

MarekBiczysko (dismissed):

ntrncic (changes requested):
The only thing missing is the documentation about this.

ntrncic (dismissed):

src/hope/models/account_attachment.py
```diff
@@ -0,0 +1,45 @@
...
+    FILE_LIMIT = 10  # max 10 files per Account
+    FILE_SIZE_LIMIT = 10 * 1024 * 1024  # 10 MB
+
+    account = models.ForeignKey(Account, on_delete=models.CASCADE, related_name="attachments")
+    title = models.CharField(max_length=255, blank=True, default="")
+    file = models.FileField()
```
saxix: let start uploading files in a structured way; let's add upload_to and return ` {program_start_year}/ba/{ba_id}/programs/{program_id}/individuals/{individual_id}/accounts/{account_id}/.` 
(NOTE: HH is not part of this path)

src/hope/one_time_scripts/migrate_wallet_images_to_account_attachments.py
```diff
@@ -0,0 +1,77 @@
...
+            print(f"[SKIP] {individual.pk}: no digital wallet account, flex field left untouched")
+            skipped += 1
+            continue
+        if not source_name or not default_storage.exists(source_name):
+            print(f"[SKIP] {individual.pk}: source file '{source_name}' missing from storage")
+            skipped += 1
+            continue
+        copied_name = None
+        try:
+            with default_storage.open(source_name) as source:
```
saxix, line 55: warning: this is linked to the previous note

src/hope/api/endpoints/rdi/mixin.py
```diff
@@ -53,8 +54,16 @@ def save_document(member: PendingIndividual, doc: dict) -> None:
...
-    def save_account(member: PendingIndividual, doc: dict) -> None:
-        PendingAccount.objects.create(individual=member, **doc)
+    def save_account(member: PendingIndividual, doc: dict) -> PendingAccount:
+        attachments = doc.pop("attachments", [])
+        account = PendingAccount.objects.create(individual=member, **doc)
+        for attachment in attachments:
+            AccountAttachment.objects.create(
+                account=account,
+                title=attachment.get("title", ""),
+                file=PhotoMixin.get_photo(attachment["file"]),
```
saxix, line 64: this is affected by the `upload_to`

saxix (changes requested):

src/hope/apps/payment/api/serializers.py
```diff
@@ -70,6 +73,50 @@
...
+            raise serializers.ValidationError("Unsupported file type.")
+
+        return file
+
+    def validate(self, data: dict) -> dict:
+        account = self.context["account"]
+        data["account"] = account
+        data["created_by"] = self.context["request"].user
+
+        if account.attachments.count() >= AccountAttachment.FILE_LIMIT:
```
saxix: this is tricky. Add a note that we do not support parallel imports for the same individual.

saxix (changes requested):

src/hope/apps/payment/api/serializers.py
pkujawa: Actually double checked this and realized this check was not needed.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
It was copied from `PaymentPlanSupportingDocumentSerializer`, which needs it because it does not have validation on the model. `AccountAttachment` does it in `clean()`, which runs after the `select_for_update` in `create()`, so the count is under the lock  and it is refused before row or file are written.

domdinicola:
@pkujawa pls resolve conflicts

saxix (approved):

PR #6320, opened 2026-08-19
331730 slow queries sentry

MarekBiczysko:
[AB#331730](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/331730)

https://monitoring.hope.unicef.org/organizations/hope/issues/3424
https://monitoring.hope.unicef.org/organizations/hope/issues/696
https://monitoring.hope.unicef.org/organizations/hope/issues/2496
https://monitoring.hope.unicef.org/organizations/hope/issues/2412
https://monitoring.hope.unicef.org/organizations/hope/issues/2987
https://monitoring.hope.unicef.org/organizations/hope/issues/2989

  - Fix grievance ticket closing when a selected individual has no household/admin2; area access is checked only when an admin2 exists.
  - Optimize activity-log counts using EXISTS, preventing duplicate M2M results while retaining programless entries.
  - Optimize payment details with a lightweight parent serializer, avoiding expensive payment-plan aggregate queries and preloading required relations.
  - Batch account uniqueness resets for account types without unique fields, replacing repeated updates during RDI merge with one query.
  - Optimized linked needs-adjudication ticket lookup by replacing the expensive OR across direct and M2M relationships with two targeted queries combined using UNION ALL.
  - Removed redundant DISTINCT from Elasticsearch-backed household searches, avoiding costly deduplication of wide household list results.


src/hope/models/account.py
```diff
@@ -235,8 +237,23 @@ def update_unique_field(self) -> None:
 
     @classmethod
     def validate_uniqueness(cls, qs: QuerySet["Account"] | list["Account"]) -> None:
-        for dmd in qs:
-            dmd.update_unique_field()
+        accounts_without_unique_fields = []
+        for account in qs:
+            unique_fields = account.unique_fields if hasattr(account, "unique_fields") else None
+            if isinstance(unique_fields, list | tuple) and not unique_fields:
```
kosior: Suggestion: we have similar logic in `update_unique_field` above. Maybe we could add small `has_no_unique_fields` method to keep those two places in sync.

kosior (dismissed):

src/hope/apps/grievance/api/views.py
```diff
@@ -801,11 +801,16 @@ def _validate_status_change_preconditions(
         ):
             partner = user.partner
             for selected_individual in grievance_ticket.ticket_details.selected_individuals.select_related(
-                "household__admin2", "program"
+                "household"
             ).all():
-                if not partner.has_area_access(
-                    area_id=selected_individual.household.admin2.id,
-                    program_id=selected_individual.program.id,
+                household = selected_individual.household
```
johniak: The join is gone but the N+1 isn't: `has_area_access` runs `get_area_limits_for_program` once per selected individual, which is why the new test measures 2 queries for a single individual. Selected individuals on an NA ticket share a program, so resolve the limits once and drop the Household join too:

```python
for admin2_id, program_id in grievance_ticket.ticket_details.selected_individuals.values_list(
    "household__admin2_id", "program_id"
):
    if not admin2_id:
        continue
    allowed = limits_by_program.setdefault(
        program_id, set(partner.get_area_limits_for_program(program_id).values_list("id", flat=True))
    )
    if allowed and admin2_id not in allowed:
        raise PermissionDenied("Permission Denied: User does not have access to close ticket")
```

src/hope/apps/core/api/mixins.py
```diff
@@ -184,6 +184,21 @@ def get_queryset(self) -> QuerySet:
 class BusinessAreaProgramsAccessMixin(BusinessAreaMixin):
     """Applies BusinessAreaMixin and also filters the qs based on the user's partner's permissions across programs."""
 
+    program_model_field_is_many = False
+
+    def _filter_many_programs_access(self, queryset: QuerySet, program_ids: list[str]) -> QuerySet:
```
johniak, line 189: This helper fixes exactly the bug that `BusinessAreaVisibilityMixin.get_queryset` still has 60 lines below: it ORs `Q(programs__id__in=[program_id])` per program into one `.filter()`, so the m2m join duplicates a ticket once per program it belongs to. `GrievanceTicketGlobalViewSet` sets `program_model_field = "programs"` and `program_model_field_is_many = True`, but that mixin never reads the flag, and its `get_count_queryset` has no `.distinct()` - so `/count/` reports 3 for a ticket in 3 programs while the list shows 1. Worth wiring the helper in there too; the ticket asks for all occurrences.

src/hope/apps/payment/api/views.py
```diff
@@ -2531,12 +2531,7 @@ class PaymentViewSet(
     filter_backends = (DjangoFilterBackend,)
     filterset_class = PaymentSearchFilter
 
-    def get_object(self) -> Payment:
-        payment_id = self.kwargs["payment_id"]
-        return get_object_or_404(Payment, id=payment_id)
-
-    def get_queryset(self) -> QuerySet:
-        parent = PaymentPlan.objects.get(pk=self.kwargs["payment_plan_pk"])
+    def _with_related_data(self, queryset: QuerySet[Payment]) -> QuerySet[Payment]:
```
johniak: `_with_related_data` is only wired into `PaymentViewSet`. `PaymentVerificationRecordViewSet.retrieve` serializes with the same `PaymentDetailSerializer` off a bare `get_object_or_404(Payment, id=...)` with no `select_related` or `prefetch_related` at all - and this PR updates that endpoint's contract snapshot, so it's already in scope. The lean parent helps it, the prefetches don't reach it.

src/hope/apps/household/filters.py
```diff
@@ -174,7 +174,7 @@ def _search_es(self, qs: QuerySet, value: Any, program: Program) -> QuerySet:
             .execute()
         )
         es_ids = [x.meta["id"] for x in es_response]
-        return qs.filter(Q(id__in=es_ids) | inner_query).distinct()
+        return qs.filter(Q(id__in=es_ids) | inner_query)
```
johniak, line 177: I traced this and dropping `.distinct()` is safe - `inner_query` is only `detail_id__endswith`, the viewset's `get_queryset` uses `Exists` subqueries rather than joins, and `filter_survey_id`/`filter_message_id` match a single id so they can't fan out. But `_search_db` right below still ends in `.distinct()` on an equally join-free query, and nothing here tests the change. Either drop it there too or leave a line saying why the ES path differs.

johniak (changes requested):

src/hope/apps/core/api/mixins.py
kosior: I think this is something that I fixed in here https://github.com/unicef/hope/pull/6321

johniak (approved):

kosior (approved):

PR #6321, opened 2026-08-19
Grievance list query optimizations

kosior:
[AB#331051](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/331051)

The grievance list endpoints were unusable on a large business area.

- **Program visibility** (`BusinessAreaVisibilityMixin`) - match programs with `Exists()` instead of an M2M join, and group programs by their area limits. Also fixes `/count`, which used to return the number of ticket×program rows.
- **`DISTINCT` dropped** from the global list and from the `is_active_program` / `active_programs_only` filters, which now use `Exists()` too.
- **Permission filter** (`GrievancePermissionsMixin`) - `Q(id__in=<subquery>)` -> `Exists()` / `~Exists()`. At 500k the `id__in` form could not be hashed at the default `work_mem`, so it became a `Materialize` over a full `Seq Scan` of the 1.5M-row through table, rescanned per candidate row.
- **Business area filter** (`BusinessAreaMixin`) - filter on the FK (`business_area=self.business_area`) instead of joining `core_businessarea` on its slug.
- **`target_id` fallback and `related_tickets_count`** - resolved once per page by `GrievanceListBatchMixin` instead of per row. **`household_household` and `household_individual` leave the main list statement entirely** - a slow grievance list no longer holds `ACCESS SHARE` on them, which is what this ticket was opened for. The global list also stops issuing one `COUNT(*)` per row.
- **`programs__sanction_lists` prefetched**, killing one `exists()` per program per row from `Program.screen_beneficiary`.

Visibility is unchanged: every rewrite returns the same ticket ids as `develop` (symmetric difference 0 across `list`, `count` and `retrieve`, for wide / narrow / no-access users).

**User-visible change:** `/count` numbers will drop. It used to count ticket×program rows, so a ticket linked to three programs counted three times.

### Numbers - 500k fixture

500 000 tickets, 1 469 715 ticket -> program links, 50 programs, one business area; local PostgreSQL 14.3, `work_mem` 4 MB. Three users seeing 99.3% / 7.9% / 2.0% of the BA.

| request | `develop` | this branch |
|---|---|---|
| global list, 50 rows | did not complete -> 11 min cancelled by hand | **~0.1–0.5 s** |
| global count | 300 s statement timeout for all three users | **0.7–1.6 s** |

At the 10k tickets the whole list request goes **381 ms -> 88 ms**, its main statement **270 ms -> 51 ms** and its statement count **225 -> 40**.


src/hope/apps/grievance/api/views.py
```diff
@@ -607,13 +590,19 @@ def get_queryset(self) -> QuerySet:
             # the loop above already asks for some of these as plain lookups, which would clash
             to_prefetch = [path for path in to_prefetch if path not in annotated_paths]
             to_prefetch += [Prefetch(path, queryset=households) for path in annotated_paths]
+
+        if self.action == "list":
+            to_prefetch.append("linked_tickets")
+
```
MarekBiczysko, line 606: if self.action == "list":
      to_prefetch.append(
          Prefetch(
              "linked_tickets",
              queryset=GrievanceTicket.objects.only("id", "household_unicef_id"),
          )
      )

  This mirrors the program-specific list and loads only the fields used by get_related_tickets_count

kosior: Cool! Thanks! I added your suggestion.

src/hope/apps/grievance/filters.py
```diff
@@ -27,6 +27,16 @@ class IsNull(Func):
     template = "%(expressions)s IS NULL"
 
 
+def program_with_status_exists(program_status: str) -> Exists:
```
johniak, line 30: `filter_by_program` further down still does `qs.filter(programs__code=value)` - the last through-table join left in this filterset. Programme codes are only unique while `is_removed=False` (`models/program.py:365`), so a ticket linked to both a removed and a live programme with the same code now comes back twice on the global list, which no longer has `.distinct()`, while `/count` still reports one. Worth routing it through an `Exists` helper too.

src/hope/apps/core/api/mixins.py
```diff
@@ -218,17 +218,32 @@ class BusinessAreaVisibilityMixin(BusinessAreaMixin):
...
+
+        if self.program_model_field_is_many:
+            filter_q = self._visibility_q_for_many(queryset.model, accessible_program_ids, area_ids_by_program)
+            without_program_q = Q(~Exists(self._program_link_queryset(queryset.model)))
+        else:
+            filter_q = self._visibility_q_for_fk(accessible_program_ids, area_ids_by_program)
+            without_program_q = Q(**{f"{self.program_model_field}__isnull": True})
+
+        # filter_q empty if no access to any program
+        return queryset.filter(Q(filter_q) | without_program_q) if filter_q else queryset.none()
```
johniak, line 261: Now that visibility is expressed with `Exists`, the `.distinct()` in `GrievanceTicketGlobalViewSet.get_count_queryset()` (`grievance/api/views.py:560`) has nothing left to dedupe - but Django still compiles it to `SELECT COUNT(*) FROM (SELECT DISTINCT id ...)`, a unique/sort over every ticket the user can see. That is the endpoint measured at 0.7-1.6 s here, so dropping it should take most of that off.

src/hope/apps/grievance/api/mixins.py
```diff
@@ -24,16 +27,82 @@
...
+    existing_tickets_counts: dict[str, int] | None = None
+
+    def get_serializer_context(self) -> dict:
+        context = super().get_serializer_context()
+        context["fallback_individual_unicef_ids"] = self.fallback_individual_unicef_ids
+        context["existing_tickets_counts"] = self.existing_tickets_counts
+        return context
+
+    def list(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        page = self.paginate_queryset(self.filter_queryset(self.get_queryset()))
```
johniak, line 90: This drops DRF's `if page is not None` branch, so a viewset with `pagination_class = None` blows up with a `TypeError` inside the batch helper instead of returning an unpaginated response. The docstring names the constraint, but keeping the branch is cheaper than the footgun:

```python
if page is None:  # pagination disabled - nothing to batch per page
    return super().list(request, *args, **kwargs)
```

tests/extras/test_utils/sql.py
```diff
@@ -0,0 +1,75 @@
+"""Helpers for asserting on the *shape* of the SQL a list endpoint builds.
```
johniak, line 1: These three helpers are only used by the grievance list tests. Tests here keep helpers next to the file that needs them - `test_utils/` is for factories and fixtures. `list_queryset` and `joined_tables` are three lines each, so putting them in `test_grievance_list_global_visibility.py` and importing from there keeps the assertions next to what they pin.

tests/extras/test_utils/sql.py
```diff
@@ -0,0 +1,75 @@
...
+        response = api_client(user).get(list_url(afghanistan))
+
+    assert "grievance_grievanceticket_programs" not in joined_tables(list_queryset(response))
+    assert '"grievance_grievanceticket"."id" IN (SELECT' not in main_list_statement(
+        captured, "grievance_grievanceticket"
+    )
+
+Why assert on SQL at all:
+
+- The rewrites change neither the response body nor the number of queries, only how many
```
johniak: This says the rewrites change neither the response body nor the number of queries, but the same PR moves `GLOBAL_LIST_QUERY_COUNT` 46 -> 41 and the caching test 37 -> 38. Worth rewording so nobody builds an assertion on it.

src/hope/apps/core/api/mixins.py
```diff
@@ -218,17 +218,32 @@ class BusinessAreaVisibilityMixin(BusinessAreaMixin):
...
+
+    def _area_ids_by_program(self, partner: Any, program_ids: list) -> dict[str, list]:
```
johniak: `partner: Any` here, and `model: Any` on `_program_link_queryset` / `_visibility_q_for_many`. The file already has a `TYPE_CHECKING` block, so `"Partner"` and `type[Model]` would work without a runtime import and give mypy something to check.

johniak (changes requested):

tests/extras/test_utils/sql.py
kosior: These aren't grievance-list-only - sql.py has three consumers across two apps:

- tests/unit/apps/core/test_business_area_mixin.py - list_queryset, joined_tables, main_list_statement
- tests/unit/apps/grievance/test_grievance_list_global_visibility.py - all three
- tests/unit/apps/grievance/test_grievance_target_id.py - main_list_statement

(a fourth, test_grievance_search_scoping.py, is coming and uses main_list_statement too)

src/hope/apps/grievance/filters.py
kosior: Done.

src/hope/apps/core/api/mixins.py
kosior: Done.

src/hope/apps/grievance/api/mixins.py
kosior: Done.

tests/extras/test_utils/sql.py
kosior: Done. Good catch.

src/hope/apps/core/api/mixins.py
kosior: Done.

MarekBiczysko (approved):

johniak (approved):

PR #6324, opened 2026-08-20
Bump version to 4.20.0

domdinicola:
### Version bump
- `4.18.0` → `4.19.0`

### Mypy fixes
- Fixed `add_fields` override type hint in `account_forms.py` (`forms.Form` → `forms.ModelForm[Any]`)
- Added `action_location` parameter to `get_actions` overrides in admin classes (matches django-stubs 6.1.0 signature)

### DIFF
--- a/src/hope/apps/payment/api/views.py
+++ b/src/hope/apps/payment/api/views.py
@@ -891,7 +891,7 @@ class PaymentPlanViewSet(
     def create(self, request: Request, *args: Any, **kwargs: Any) -> Response:
         if "target_population_id" not in request.data:
             raise ValidationError("target_population_id is required")
-        payment_plan = get_object_or_404(PaymentPlan, id=request.data["target_population_id"])
+        payment_plan = get_object_or_404(PaymentPlan, id=request.data.get("target_population_id"))
         serializer = self.get_serializer(data=request.data, context={"payment_plan": payment_plan})
         serializer.is_valid(raise_exception=True)
         old_payment_plan = copy_model_object(payment_plan)
@@ -1624,7 +1624,9 @@ class PaymentPlanViewSet(
             raise ValidationError("Payment plan must be accepted to make a split")
 
         payments_no = request.data.get("payments_no")
-        split_type = request.data["split_type"]
+        split_type = request.data.get("split_type")
+        if not split_type:
+            raise ValidationError("split_type is required")
         if split_type == PaymentPlanSplit.SplitType.BY_RECORDS:
             if not payments_no:
                 raise ValidationError("Payment Number is required for split by records")
@@ -2265,7 +2267,7 @@ class TargetPopulationViewSet(
     @transaction.atomic
     def copy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
         user = request.user
-        request.data["target_population_id"] = kwargs.get("pk")
+        cast("dict[str, Any]", request.data)["target_population_id"] = kwargs.get("pk")
 
         serializer = self.get_serializer(
             data=request.data,
--- a/tests/unit/apps/sanction_list/test_sl_admin.py
+++ b/tests/unit/apps/sanction_list/test_sl_admin.py
@@ -1,13 +1,46 @@
 from typing import TYPE_CHECKING
+from unittest.mock import patch
 
+from django.contrib.admin.sites import AdminSite
+from django.http import HttpRequest
 from django.urls import reverse
+import pytest
 import responses
+
+from extras.test_utils.factories import UserFactory
+from hope.admin.sanction_list_uploaded_xlsx_file import UploadedXLSXFileAdmin
+from hope.models import UploadedXLSXFile
 
 if TYPE_CHECKING:
     from django_webtest import DjangoTestApp
     from responses import RequestsMock
 
     from hope.models import SanctionList, User
+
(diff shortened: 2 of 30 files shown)


src/hope/api/utils.py
```diff
@@ -6,6 +6,12 @@
 def humanize_errors(errors: dict) -> dict:
     try:
         households = errors.pop("households", [])
+        if isinstance(households, dict):
```
johniak: This handles the top level, but DRF 3.18 makes the same change to the nested `members` errors below: they are now index dicts too. `enumerate(members, 1)` then iterates the dict keys, so a member error humanizes to `{'Member #1': [1]}` — the real error is replaced by the index, and the member at index 0 is dropped by `if m`. Reproduced with `RDINestedSerializer` on this branch; it hits `UploadRDIView` and `PushLaxToRDIView`. Apply the same dict-to-list conversion to `members` before the `isinstance(members, list)` check.

src/hope/api/endpoints/rdi/push_people.py
```diff
@@ -258,4 +258,6 @@ def post(self, request: "Request", business_area: str, rdi: UUID) -> Response:
                 "people": people_ids,
             }
             return Response(response, status=status.HTTP_201_CREATED)
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+        errors = serializer.errors
+        error_list = list(errors.values()) if isinstance(errors, dict) else errors
```
johniak: `list(errors.values())` breaks the compatibility it is meant to restore: DRF 3.18 dicts contain only the failing indices (4 items with items 1 and 3 invalid gives `{1: ..., 3: ...}`), while 3.17 returned `[{}, err, {}, err]`. `values()` collapses that, so a mixed batch maps errors to the wrong people. It also turns a non-list payload's `{'non_field_errors': [...]}` into `[[...]]`. Rebuild positionally:
```python
if isinstance(errors, dict) and all(isinstance(k, int) for k in errors):
    errors = [errors.get(i, {}) for i in range(len(request.data))]
```

src/hope/api/utils.py
```diff
@@ -6,6 +6,12 @@
...
+            max_idx = max((int(k) for k in households), default=-1)
```
johniak: When the `households` payload is not a list, DRF 3.18 puts `{'non_field_errors': [...]}` under this key, so `int(k)` raises ValueError after the destructive `errors.pop("households")` — the except returns the mutated dict and the API answers 400 with `{}`, silently swallowing the only error (verified on this branch; before the upgrade this input failed loudly with a 500). Guard the branch with `all(isinstance(k, int) for k in households)` so non-index dicts pass through.

tests/unit/api/test_push_people.py
```diff
@@ -225,12 +225,12 @@ def test_upload_with_errors(token_api_client, push_people_url, program, afghanis
     assert response.json() == [
         {
             "birth_date": ["This field is required."],
-            "documents": [
-                {
+            "documents": {
```
johniak: This locks in the new nested shape — `documents` as a dict instead of a list — which is a breaking change for external clients, while the PR body says the endpoint was normalized for backward compatibility. The same shape change hits `accounts` here and `delegates`/`members` on the other RDI endpoints with no normalization at all. Either normalize recursively or state the nested contract change explicitly in the PR description.

johniak (changes requested):

domdinicola:
@johniak I reverted DRF changes (will implement in another PR)

tests/unit/apps/sanction_list/test_sl_admin.py
```diff
@@ -1,15 +1,44 @@
...
+
 if TYPE_CHECKING:
     from django_webtest import DjangoTestApp
     from responses import RequestsMock
 
     from hope.models import SanctionList, User
 
 
+def test_uploaded_xlsx_get_actions() -> None:
+    user = UserFactory(is_staff=True)
```
johniak: `UserFactory` is called directly in the test body here and in the test below. Our convention is factories only inside fixtures — move the user/request setup into a local fixture in this file.

johniak (changes requested):

johniak (dismissed):

ntrncic:
we need staging to be 4.19 and this one to be 4.20

src/hope/apps/payment/api/views.py
```diff
@@ -1624,7 +1624,9 @@ def split(self, request: Request, *args: Any, **kwargs: Any) -> Response:
             raise ValidationError("Payment plan must be accepted to make a split")
 
         payments_no = request.data.get("payments_no")
-        split_type = request.data["split_type"]
+        split_type = request.data.get("split_type")
+        if not split_type:
+            raise ValidationError("split_type is required")
```
johniak, line 1629: This branch is what makes codecov/patch fail (92.59% vs 100%) — nothing exercises the missing split_type case. It's also the real behavior change in this file: a KeyError 500 becomes a 400. One test in test_payment_plan_actions_views.py posting to split without split_type and asserting the 400 message pins it and turns the check green.

johniak (changes requested):

kosior (approved):

johniak (approved):

PR #6325, opened 2026-08-20
Fix admin search timeout on Individual/Household unicef_id

johniak:
[AB#335131](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/335131)

Searching the Individual admin by `unicef_id` timed out.

Django OR-ed `unicef_id` together with the name fields into one big `UPPER(col) LIKE UPPER('%term%')`. No index can serve that, so Postgres read all 20M rows. The cursor paginator added a second unbounded `COUNT(*)` on top.

Now a `unicef_id` search is handled on its own, separately from the name search: the term is upper-cased and matched with `startswith`, which uses the index that already exists. Searching by name keeps the old behaviour. The `COUNT(*)` is gone.

Same fix in the Household admin.

Searching by name is still slow, and it cannot be fixed the same way — `LIKE '%term%'` needs a trigram index. `pg_trgm` is available on the prod server but not installed, so fixing search on the other fields means installing the extension on the prod DB first.


src/hope/admin/utils.py
```diff
@@ -508,3 +508,24 @@ def get_related(self, user: Model, field: Any, manager: str, max_records: int =
...
+    Django turns `search_fields` into `UPPER(col) LIKE UPPER('%term%')` OR-ed together, which no
+    index can serve; on `household_individual` that means walking 20M rows. unicef_ids are generated
+    uppercase by the `create_*_unicef_id` triggers, so upper-casing the term lets a plain
+    `startswith` hit the existing `..._like` (varchar_pattern_ops) index.
+    """
+
+    # unicef_id prefix a user may paste -> field to prefix-match on
+    unicef_id_search_map: dict[str, str] = {}
+
+    def get_search_results(self, request: HttpRequest, queryset: QuerySet, search_term: str) -> tuple[QuerySet, bool]:
```
MarekBiczysko, line 555: A multi-term search like IND-24-0000.0001 Kowalski is incorrectly treated as one UNICEF ID prefix, producing no results.

Fix: Use the optimized ID lookup only for single-token searches; otherwise use Django’s default multi-term search.

  field = self.unicef_id_search_map.get(prefix) if rest and len(term.split()) == 1 else None

  Example:

  - IND-24-0000.0001 → optimized startswith search
  - IND-24-0000.0001 Kowalski → Django’s standard multi-term search

MarekBiczysko (dismissed):

nerjan (approved):

PR #6328, opened 2026-08-24
Used Kab in Dashboar

srugano:
- Report individuals, children_counts, and pwd_counts from the new kab_* household counters (true affected beneficiaries), falling back to the legacy denormalised fields when KAB is unknown; remove the fertility-rate extrapolation entirely.
- Refactor dashboard aggregation: dedupe the household-counting logic between the country and global paths, reuse _build_summary_results, and centralize the shared date/annotation expressions.
- Make update_dashboard_figures log and continue on per-BA/global failures (consistent with the recent-refresh task); return a generic message on trigger errors instead of leaking DB details.
- Replace magic payment-status strings with model enums and clean up dead code (unused router, Protocol base class, reduce for the pwd expression).
- Update dashboard tests for KAB semantics (incl. NULL-fallback and social cases) and add coverage for get_kab_pwd_count_expression.


src/hope/apps/dashboard/services.py
```diff
@@ -19,91 +19,109 @@
...
+    if not (household_id and isinstance(household_id, UUID) and household_id not in seen_set):
+        return
+    h_data = household_map.get(household_id, {})
+    current_summary["households"] += 1
+
+    kab_size = h_data.get("kab_size")
+    if kab_size is not None:
+        # KAB known: use the true affected-beneficiary counts.
+        current_summary["individuals"] += int(kab_size)
+        pwd_count = h_data.get("kab_pwd_count", 0)
```
johniak: This re-creates the Myanmar "0 PwD" bug (AB#322684). The KAB mirror path in `_recalculate_kab` copies the declared composition verbatim, including NULL `*_disabled_count` fields, and `get_kab_pwd_count_expression()` coalesces those NULLs to 0. For a non-`full_collection` household that declared age groups but not disabled counts, the legacy path counted disabled individuals via the `pwd_count_calc` subquery and reported a real number — this branch now reports 0 for the same household. Fall back to the legacy `pwd_count` when all ten `kab_*_disabled_count` fields are NULL, or confirm with product that "nothing declared ⇒ 0 known PwD" is intended despite AB#322684.

src/hope/apps/dashboard/services.py
```diff
@@ -19,91 +19,109 @@
...
+    """Add a household's affected-people counts to the summary (KAB-first, legacy fallback)."""
+    if not (household_id and isinstance(household_id, UUID) and household_id not in seen_set):
+        return
+    h_data = household_map.get(household_id, {})
+    current_summary["households"] += 1
+
+    kab_size = h_data.get("kab_size")
+    if kab_size is not None:
+        # KAB known: use the true affected-beneficiary counts.
+        current_summary["individuals"] += int(kab_size)
```
johniak: The KAB branch drops the social-program rule. Legacy forced `size` to 1 for SOCIAL households (`_get_household_data`: `1 if is_sw_program or size_value is None else size_value`) because in People programs only the registered person is the reached beneficiary — but `kab_size` is passed through raw and `_recalculate_kab` has no social special-case, so a social household whose KAB was mirrored from a declared composition adds its whole declared family size here. The tests only cover SOCIAL with `kab_size=None`. Either hoist `is_sw_program` above this branch and do `individuals += 1 if is_sw_program else int(kab_size)`, or add a SOCIAL + kab_size test pinning the intended behaviour.

src/hope/apps/dashboard/celery_tasks.py
```diff
@@ -36,15 +36,19 @@ def update_dashboard_figures() -> None:
         lock_acquired = cache.add(lock_key, True, timeout=60 * 60)
         try:
             DashboardDataCache.refresh_data(business_area.slug)
+        except Exception as e:
```
johniak, line 41: This blanket `except Exception` (here and in the global block) makes the task's `autoretry_for=(OperationalError, ProgrammingError, InvalidCursorName)` unreachable — no exception can escape the function anymore, so a transient DB error during the nightly run leaves that BA's dashboard stale for 24h where the whole task previously retried up to 3×. If log-and-continue is the intended contract (it matches `update_recent_dashboard_figures`), drop the now-dead `autoretry_for`/`retry_kwargs` from the decorator; otherwise re-raise those DB exceptions.

tests/unit/apps/dashboard/test_services_calculations.py
```diff
@@ -147,6 +145,38 @@ def test_pwd_count_expression(household_index: int, expected_pwd_count: int, hou
...
+        kab_female_age_group_0_5_disabled_count=1,
+        kab_female_age_group_6_11_disabled_count=None,
+        kab_male_age_group_18_59_disabled_count=3,
+    )
+    hh_kab_none = HouseholdFactory(business_area=ba)
+    return hh_kab_pwd, hh_kab_none
+
+
+@pytest.mark.django_db
+def test_kab_pwd_count_expression(households_for_kab_pwd_test) -> None:
```
johniak: This asserts two independent scenarios (partial-NULL sum and all-NULL) in one test body; the sibling `test_pwd_count_expression` above parametrizes exactly this shape. Splitting via `pytest.mark.parametrize` keeps one scenario per test and makes failures self-describing.

ruff.toml
```diff
@@ -75,7 +75,7 @@ skip-magic-trailing-comma = false
 line-ending = "auto"
 
 [lint.per-file-ignores]
-"tests/**.py" = ["S101", "PLR2004", "S", "SIM117", "D", "UP", "PLR0913", "PLR0915", "ANN", "N999"]
+"tests/**.py" = ["S101", "PLR2004", "S", "SIM117", "D", "UP", "PLR0913", "PLR0915", "PLR0917", "ANN", "N999"]
```
johniak: `PLR0917` is a preview rule and this config has no `preview = true`, so adding it to the ignore list is a no-op — PLR0913 already covers the arg-count complaints the new test signatures could trigger. Drop it, or note why it's pre-emptive.

johniak (changes requested):

domdinicola:
@srugano @ntrncic @saxix shall we implement this or not spend time and do this in the new dashboard repo?
@johniak

ntrncic:
> @srugano @ntrncic @saxix shall we implement this or not spend time and do this in the new dashboard repo? @johniak

let's leave this dashboard here but should also update the new one.

src/hope/apps/dashboard/services.py
srugano: Addressed.

src/hope/apps/dashboard/services.py
srugano: Reverted back

src/hope/apps/dashboard/celery_tasks.py
srugano: Done

src/hope/apps/dashboard/celery_tasks.py
```diff
@@ -36,15 +36,23 @@ def update_dashboard_figures() -> None:
         lock_acquired = cache.add(lock_key, True, timeout=60 * 60)
         try:
             DashboardDataCache.refresh_data(business_area.slug)
+        except (OperationalError, ProgrammingError, psycopg2.errors.InvalidCursorName):
+            raise
```
johniak, line 40: These two `raise` lines (here and in the global block) are the only uncovered lines in the patch — that's what keeps codecov/patch red. A test with `refresh_data` side-effecting `OperationalError` and `pytest.raises` around `update_dashboard_figures.apply(throw=True)` would both pin that the retry path is actually reachable (the point of this fix) and turn the gate green.

johniak (changes requested):

tests/unit/apps/dashboard/test_tasks.py
```diff
@@ -140,45 +140,69 @@ def test_generate_dash_report_task_business_area_not_found(mock_refresh_data: Mo
...
+
+
+@patch("hope.apps.dashboard.celery_tasks.DashboardDataCache.refresh_data")
+def test_update_dashboard_figures_reraises_ba_db_error(mock_ba_refresh: Mock, afghanistan: BusinessArea) -> None:
+    """update_dashboard_figures re-raises DB errors from the BA loop so celery can autoretry."""
+    mocked_error_message = "Mocked DB error"
+    mock_ba_refresh.side_effect = OperationalError(mocked_error_message)
+    BusinessArea.objects.exclude(slug=afghanistan.slug).update(active=False)
+
+    with pytest.raises(Exception, match=mocked_error_message):
```
johniak, line 185: Under eager `apply(throw=True)` what escapes here is `celery.exceptions.Retry` wrapping the `OperationalError`, and `refresh_data` runs exactly once; `match=` only passes because `str(Retry)` embeds the wrapped message. `pytest.raises(Retry)` would pin what the docstring claims and can't be satisfied by an unrelated exception. Same for the global test below.

johniak (approved):

domdinicola (approved):

PR #6329, opened 2026-08-24
[AB#330006]-Add Payment Purpose, Group, Cycle columns to household payments table

mmaciekk:
[AB#330006](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/330006)
## Summary
- Expose `payment_plan_cycle`, `payment_plan_group`, and `payment_plan_purposes` on `PaymentListSerializer`
- Preload the parent payment plan so the new columns don't cost a query per row
- Render three new columns — Payment Purpose, Group, Cycle — in `PaymentsHouseholdTable`, for both the household and People (individual) variants.


  ## Also in this PR                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                             
  - Make the columns sortable. The household payments endpoint was ignoring the sort                                                                                                                                                                                                                                         
    parameter, so every arrow on that table did nothing — the older columns included.                                                                                                                                                                                                                                        
    Payment Purpose stays unsorted; a plan can have several purposes.                                                                                                                                                                                                                                                        
  - Point the Head of Household column at the field it sorts on. Its label replacement key                                                                                                                                                                                                                                   
    in PaymentsHouseholdTable.tsx moved with it.                                                                                                                                                                                                                                                                             
  - Sort payment purposes by name. They had no set order and reshuffled between page loads.                                                                                                                                                                                                                                  
  - Replace the per-viewset select_related lists with one shared queryset in                                                                                                                                                                                                                                                 
    payment/api/querysets.py, so all payment list endpoints stop doing a query per row.                                                                                                                                                                                                                                      
  - Add tests for the household payments endpoint, including a check that the query count                                                                                                                                                                                                                                  
    doesn't grow with the number of rows.


pkujawa (dismissed):

src/hope/models/payment_plan_purpose.py
```diff
@@ -25,6 +25,7 @@ class PaymentPlanPurpose(TimeStampedUUIDModel, UnicefIdentifiedModel):
     class Meta:
         app_label = "payment"
         verbose_name = _("Payment Plan Purpose")
+        ordering = ["name"]
```
MarekBiczysko: pls remove this default ordering (and migration) and just order in 
serializers -> def get_payment_plan_purposes

MarekBiczysko (changes requested):

src/hope/models/payment_plan_purpose.py
domdinicola: I think it would be good to have ordering it here, why are you against @MarekBiczysko ?

mmaciekk: Done in fddf62f - removed the ordering from Meta and deleted the migration. The sorting now happens in `get_payment_plan_purposes`.

@domdinicola one thing to note: I sorted with plain `sorted()` in Python instead of `.order_by("name")`. The purposes are already prefetched in `payment/api/querysets.py`, and calling `.order_by()` on them would throw that prefetch away and run a query for every single payment row. That is the N+1 this PR was fixing in the first place.

MarekBiczysko (approved):

PR #6330, opened 2026-08-25
[AB#320656]-Fix Payments table column sorting

mmaciekk:
[AB#320656](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/320656)

## Summary
Sorting didn't work for several columns in the Payments table (Administrative Level 2, Collector, FSP, Entitlement, Delivered Quantity, FSP Auth Code, Reconciliation). This fixes it on the backend so those columns can be sorted, except Collector, which is now not sortable since the displayed value can't be sorted properly.

### DIFF
--- a/src/frontend/src/containers/tables/paymentmodule/PaymentsTable/PaymentsTableHeadCells.tsx
+++ b/src/frontend/src/containers/tables/paymentmodule/PaymentsTable/PaymentsTableHeadCells.tsx
@@ -76,7 +76,7 @@ export const headCells: HeadCell<PaymentList>[] = [
   {
     disablePadding: false,
     label: 'Reconciliation',
-    id: 'mark',
+    id: 'reconciliation_rank',
     numeric: false,
   },
 ];
@@ -150,7 +150,7 @@ export const headCellsPeople: HeadCell<PaymentList>[] = [
   {
     disablePadding: false,
     label: 'Reconciliation',
-    id: 'mark',
+    id: 'reconciliation_rank',
     numeric: false,
   },
 ];
--- a/src/frontend/src/containers/tables/paymentmodule/PaymentsTable/PaymentsTableHeadCells.tsx
+++ b/src/frontend/src/containers/tables/paymentmodule/PaymentsTable/PaymentsTableHeadCells.tsx
@@ -39,6 +39,7 @@ export const headCells: HeadCell<PaymentList>[] = [
     label: 'Collector',
     id: 'collector_id',
     numeric: false,
+    disableSort: true,
   },
   {
     disablePadding: false,
--- a/src/hope/apps/payment/api/filters.py
+++ b/src/hope/apps/payment/api/filters.py
@@ -1,6 +1,6 @@
 from typing import Any
 
-from django.db.models import Q, QuerySet
+from django.db.models import Case, IntegerField, Q, QuerySet, Value, When
 import django_filters
 from django_filters import FilterSet, OrderingFilter
 
@@ -297,9 +297,31 @@ class PaymentSearchFilter(FilterSet):
             ("household__size", "household__size"),
             ("collector__full_name", "collector__full_name"),
             ("created_at", "created_at"),
+            ("household__admin2__name", "household__admin2__name"),
+            ("collector_id", "collector_id"),
+            ("financial_service_provider__name", "financial_service_provider__name"),
+            ("entitlement_quantity_usd", "entitlement_quantity_usd"),
+            ("delivered_quantity", "delivered_quantity"),
+            ("fsp_auth_code", "fsp_auth_code"),
+            ("mark", "mark"),
         )
     )
 
     class Meta:
         model = Payment
         fields = []
+
+    def filter_queryset(self, queryset: QuerySet) -> "QuerySet[Payment]":
+        queryset = queryset.annotate(
+            mark=Case(
+                When(status=Payment.STATUS_DISTRIBUTION_SUCCESS, then=Value(1)),
+                When(status=Payment.STATUS_DISTRIBUTION_PARTIAL, then=Value(2)),
+                When(status=Payment.STATUS_NOT_DISTRIBUTED, then=Value(3)),
+                When(status=Payment.STATUS_ERROR, then=Value(4)),
+                When(status=Payment.STATUS_FORCE_FAILED, then=Value(5)),
+                When(status=Payment.STATUS_MANUALLY_CANCELLED, then=Value(6)),
+                When(status=Payment.STATUS_PENDING, then=Value(7)),
+                output_field=IntegerField(),
+            )
+        )
+        return super().filter_queryset(queryset)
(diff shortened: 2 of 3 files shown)


src/hope/apps/payment/api/filters.py
```diff
@@ -297,9 +297,31 @@ class PaymentSearchFilter(FilterSet):
...
     )
 
     class Meta:
         model = Payment
         fields = []
+
+    def filter_queryset(self, queryset: QuerySet) -> "QuerySet[Payment]":
+        queryset = queryset.annotate(
+            mark=Case(
+                When(status=Payment.STATUS_DISTRIBUTION_SUCCESS, then=Value(1)),
```
Jeeziorny: how about STATUS_SUCCESS, STATUS_SENT_TO_PG, STATUS_SENT_TO_FSP? They will have `mark = NULL` and sort order for them will be undefined

src/hope/apps/payment/api/filters.py
```diff
@@ -1,6 +1,6 @@
 from typing import Any
```
Jeeziorny, line 1: 1. User orders ONLY by `mark`.
2. User gets a response with results sorted by mark, let's say he gets a record X on 49th place.
3. In the meantime something happens to db planner, state is changing.
4. User requests second page, still ordered ONLY by mark.
5. User _can_ get record X on second page, query is not being kept between the request.

@nerjan am I halucinating or is it the issue with all filters here?

src/frontend/src/containers/tables/paymentmodule/PaymentsTable/PaymentsTableHeadCells.tsx
```diff
@@ -39,6 +39,7 @@ export const headCells: HeadCell<PaymentList>[] = [
     label: 'Collector',
     id: 'collector_id',
     numeric: false,
+    disableSort: true,
```
Jeeziorny, line 42: collector id is disabled from sort on FE, but it's included in the filter, looks like a bug

src/hope/apps/payment/api/filters.py
```diff
@@ -297,9 +297,31 @@ class PaymentSearchFilter(FilterSet):
...
         )
     )
 
     class Meta:
         model = Payment
         fields = []
+
+    def filter_queryset(self, queryset: QuerySet) -> "QuerySet[Payment]":
+        queryset = queryset.annotate(
+            mark=Case(
```
Jeeziorny: WDYT about changing `mark` to `reconciliation_rank` or `delivery_status_order`? Seems to be more verbose

tests/unit/apps/payment/test_payment_list_ordering.py
```diff
@@ -41,3 +46,72 @@ def test_payment_search_filter_orders_by_created_at_descending(
     result = PaymentSearchFilter(data={"ordering": "-created_at"}, queryset=Payment.objects.all()).qs
 
     assert list(result.values_list("pk", flat=True)) == [newest.pk, middle.pk, oldest.pk]
+
+
+def test_payment_search_filter_orders_by_admin2_name() -> None:
```
Jeeziorny: Look for the tests that could be folded into one and implemented with pytest parametrisation

src/hope/apps/payment/api/filters.py
nerjan: yes, I think that is reall.

src/hope/apps/payment/api/filters.py
mmaciekk: Good catch, fixed. Those three now get a rank instead of NULL.

src/hope/apps/payment/api/filters.py
mmaciekk: Added StableOrderingFilter which always appends pk. Applied to PaymentSearchFilter here; the other filters in this file have the same issue but are pre-existing, so I left them for a follow-up.

src/frontend/src/containers/tables/paymentmodule/PaymentsTable/PaymentsTableHeadCells.tsx
mmaciekk: You are right, removed collector_id from the ordering list.

src/hope/apps/payment/api/filters.py
mmaciekk: Renamed to reconciliation_rank, matches the column label.

tests/unit/apps/payment/test_payment_list_ordering.py
mmaciekk: Done, the five per-column tests are now one parametrised test.

src/hope/apps/payment/api/filters.py
```diff
@@ -271,6 +271,19 @@ def search_filter(self, qs: QuerySet, name: str, value: str) -> "QuerySet[Paymen
...
+class StableOrderingFilter(OrderingFilter):
+    """OrderingFilter that always appends a unique tie-breaker.
+
+    `order_by()` replaces the model's `Meta.ordering`, so rows tied on the requested column have
+    no defined relative order. With LIMIT/OFFSET pagination that lets a row appear on two pages
+    or on none, because each page is a separate query and Postgres may re-plan it.
+    """
+
+    def filter(self, qs: QuerySet, value: Any) -> QuerySet:
+        ordering = [self.get_ordering_value(param) for param in value] if value else ["-created_at"]
```
Jeeziorny, line 283: Why have you choose `created_at` as a criteria for stable ordering? I'm not judging. When I was analysing this issue I thought that we should use unicef id that is forced to be unique, but I have no arguments why it could be better than `created_at`.

Jeeziorny (approved):

src/hope/apps/payment/api/filters.py
mmaciekk: I thought created_at will also be unique realistically. Probably unicef_id would be the best choice.

src/hope/apps/payment/api/filters.py
Jeeziorny: https://dev.azure.com/unicef/ICTD-HCT-MIS/_backlogs/backlog/HOPE%20Team/Stories?workitem=338034

src/hope/apps/payment/api/filters.py
mmaciekk: Why created_at for the default, and pk only as tie-breaker:

- pk is a random UUID (via TimeStampedUUIDModel) — it has no relation to when a row was created, so sorting by it alone would look arbitrary to the user.
- created_at gives a meaningful default: newest payments first.
- But timestamps can collide (e.g. bulk-created rows), so created_at alone doesn't guarantee a stable order across paginated pages.
- pk is added after created_at purely to break those ties — it's unique and never null, so it guarantees a deterministic, stable order without needing to mean anything itself.

Why pk and not unicef_id as that tie-breaker: unicef_id is nullable and not unique, so it can't guarantee uniqueness the way pk can — it wouldn't reliably break ties.

mmaciekk: So not unicef_id but pk and created_at combined

nerjan (approved):

PR #6331, opened 2026-08-25
336378_User_BusinessArea_Timezone

MarekBiczysko:
[AB#336378](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/336378)

## What this PR does

Adds timezone support so that dates and times shown in HOPE reflect a User's own timezone (or, if they haven't set one, their Business Area's timezone) instead of always being UTC.

## Backend

- Added a `timezone` field to `BusinessArea` and to `User`.
  - A new Business Area automatically gets its timezone from its office country when one isn't set explicitly (falls back to UTC if the country is missing/unmapped).
  - A User's timezone is optional — if not set, they inherit their current Business Area's timezone.
- Added `hope/apps/core/timezones.py` with helpers to work out the "effective" timezone for a User/Business Area and to format dates/times in that timezone.
- Added new endpoints for the logged-in user to view/update their own timezone preference and to get the list of valid timezone choices.
- Updated the profile endpoint to return both the User's chosen `timezone` and the resolved `effective_timezone`.
- Updated places that generate dates/times shown to users (activity logs, notifications, exports, xlsx templates, payment services, etc.) to use the new effective-timezone logic instead of always assuming UTC.
- Added a data migration to backfill timezones for existing Business Areas based on their office country.
- Added documentation explaining how timezone resolution works (`docs/guide-adm/timezone.md`).

## Frontend

- Added a timezone picker to the user profile menu, letting a user either pick their own timezone or use the current Business Area's timezone, and see the current local time update live.
- Added a `timezoneContext` and `utils/timezone.ts` helpers so components can format dates/times consistently using the effective timezone.
- Updated activity log tables and other date-displaying components to use the new timezone-aware formatting.

## Tests

- Added backend unit tests for the new timezone resolution logic, the Business Area/User timezone fields, the new profile-timezone and timezone-choices endpoints, and updated affected tests (notifications, exports, celery tasks, serializers) to account for timezone-aware output.
- Added frontend unit tests for the timezone utilities and the updated `UserProfileMenu` component.
- Added a new Selenium end-to-end test (`tests/e2e/new_selenium/user_profile/test_timezone_preference.py`) covering changing the user's timezone preference through the UI.
- Updated API contract fixtures/snapshots to include the new timezone fields.

Screenshots:
When backend sends datetime, tooltip appears on hover
image
When backend sends date, show plain date adjusted to the effective timezone:
image
Allow user to change the timezone and display current time in the top right corner menu
image


src/frontend/src/components/core/UniversalMoment.tsx
```diff
@@ -1,27 +1,41 @@
...
+    );
+  }
+
+  const date = parseInstant(children);
+  if (!date) {
+    return <>-</>;
+  }
+
+  const mode = withTime ? 'dateTime' : 'date';
+  const formattedDate = formatInstant(date, timezone, mode);
```
johniak, line 33: This now renders `household.lastRegistrationDate` / `firstRegistrationDate` through the effective timezone, but those are `DateTimeField`s holding a calendar date: the XLSX import stores the date cell as UTC midnight (`_handle_datetime` → `timezone_datetime`), and the household serializers still emit them as datetimes. A household registered on 5 Mar shows as 4 Mar for anyone west of UTC, i.e. every LAC business area. Before this PR the date was sliced from the ISO string. Give those fields `UTCDateField` like `delivery_date` / `start_date`, or keep the old slice for datetimes rendered without `withTime`.

src/hope/apps/payment/notifications.py
```diff
@@ -92,24 +94,60 @@ def _prepare_user_recipients(self) -> QuerySet[User]:
...
+                        variables=self._prepare_body_variables(action_user_timezone),
+                    )
+                )
+            return emails
+
+        return [
+            MailjetClient(
+                mailjet_template_id=config.MAILJET_TEMPLATE_PAYMENT_PLAN_NOTIFICATION,
+                subject=self.email_subject,
+                recipients=[],
```
johniak: With no other recipients this builds an email with an empty `To` and the action user in `Cc`, but `MailjetClient.send_email` iterates `chunks(self.recipients, 50)`, so an empty `To` yields no batch and nothing is sent. Ten lines above, the same "action user alone" case gets a real `To` email. Use `recipients=[self.action_user.email]` here too. Same in `periodic_data_update/notifications.py` L125.

src/hope/apps/grievance/celery_tasks.py
```diff
@@ -93,32 +123,59 @@ def daily_grievance_digest_async_task_action(job: PeriodicAsyncJob) -> None:
...
-            description=f"Send the {digest_date} grievance digest for {name}",
-        )
+    now = timezone.now()
+    notification_hour = get_grievance_notification_hour()
+    for business_area in BusinessArea.objects.filter(enable_email_notification=True).only("id", "name", "timezone"):
+        for timezone_name in DailyDigestService.recipient_timezone_names(business_area):
+            notification_date, _ = latest_local_schedule_time(timezone_name, now, notification_hour)
+            digest_date = notification_date - timedelta(days=1)
+            delivery_key = _daily_digest_delivery_key(str(business_area.id), timezone_name, digest_date)
+            if _daily_digest_dispatch_exists(delivery_key):
```
johniak, line 160: The dedup only recognises jobs with `config.delivery_key`; the digests already sent by the current code are keyed by `config.completed_for`. On the first hourly run after deploy, every BA whose latest local window has passed gets its previous-day digest dispatched again, so everyone receives yesterday's digest twice. Check the legacy `completed_for` key for one release, or backfill `delivery_key` on the existing rows.

src/hope/apps/payment/celery_tasks.py
```diff
@@ -1370,22 +1370,22 @@ def send_payment_notification_emails_async_task_action(job: AsyncJob) -> None:
         payment_plan,
         job.config["action"],
         action_user,
-        job.config["action_date_formatted"],
+        datetime.datetime.fromisoformat(job.config["action_date"]),
```
johniak, line 1373: `job.config["action_date"]` raises `KeyError` for any job queued before deploy, which still carries `action_date_formatted`. `AsyncRetryJob` rows waiting for retry on the PDU side (`periodic_data_update/celery_tasks.py` L194) die on the first run after the release. Read both keys for one release, or keep the old key as a fallback.

src/hope/apps/account/api/views.py
```diff
@@ -152,6 +164,30 @@ def profile(self, request: "Request", *args: Any, **kwargs: Any) -> Response:
...
+
+    @extend_schema(responses=TimezoneChoiceSerializer(many=True), filters=False)
+    @action(
+        detail=False,
+        methods=["get"],
+        url_path="timezone-choices",
+        url_name="timezone-choices",
+        pagination_class=None,
+    )
+    def timezone_choices(self, request: "Request", *args: object, **kwargs: object) -> Response:
```
johniak: This is a static, BA-independent list, but it lives on the BA-nested `UserViewSet`; every other static list is on `ChoicesViewSet` under `/api/rest/choices/<name>/`. Also `available_timezones()` walks the tzdata directory on every request. Compute the sorted list once at import and expose it from `ChoicesViewSet`.

src/hope/apps/grievance/services/daily_digest_service.py
```diff
@@ -48,34 +51,72 @@ class DailyDigestService:
     # limit number of tickets that can be send in one email, and how many are ever held in memory
     ROW_LIMIT = 50
 
-    def __init__(self, business_area: "BusinessArea", digest_date: date) -> None:
+    @staticmethod
+    def recipient_timezone_names(business_area: "BusinessArea") -> set[str]:
+        user_timezones = (
+            User.objects.filter(
+                Q(assigned_tickets__business_area=business_area) | Q(created_tickets__business_area=business_area)
```
johniak: `Q(assigned_tickets__...) | Q(created_tickets__...)` OR-joins the whole ticket table twice per BA, and this now runs every hour for every BA with notifications on. Two `Exists()` subqueries on `User` filtered by `timezone__isnull=False` first keep it to an index probe. The grievance list already had a production hang from an OR'd join of this shape.

src/hope/apps/grievance/services/notification_schedule.py
```diff
@@ -0,0 +1,28 @@
...
+from constance import config
+
+from hope.apps.core.timezones import latest_local_schedule_time, resolve_timezone_name
+from hope.apps.grievance.models import GrievanceTicket
+
+
+def get_grievance_notification_hour() -> int:
+    notification_hour = int(config.GRIEVANCE_NOTIFICATION_HOUR)
+    if not 0 <= notification_hour <= 23:
+        raise ValueError("GRIEVANCE_NOTIFICATION_HOUR must be between 0 and 23")
```
johniak: A bad `GRIEVANCE_NOTIFICATION_HOUR` (someone types 24 in constance) raises here and takes down both beat actions: no digests and no reminders for every BA until someone reads the worker log. Clamp with a logged warning, or validate on the constance field instead of at send time.

johniak (changes requested):

pkujawa (dismissed):

src/hope/apps/grievance/celery_tasks.py
MarekBiczysko: config.SEND_GRIEVANCES_NOTIFICATION is False on prod, so there are no legacy entries, so nothing to backfill

johniak (approved):

pkujawa (approved):

PR #6333, opened 2026-08-25
Really soft-delete Top-Up / Follow-Up / Top-Up Amendment payment plans on delete

johniak:
[AB#337306](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/337306)

Deleting a child payment plan (OPEN status) used to push it back to DRAFT, leaving it on the Targeting list and in the "Create Payment Plan" flow, while its payments kept blocking those households from any new child plan (AFG case: 97 of 1542 records missing from a new top-up, [AB#337009](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/337009)).

- Delete of a Top-Up / Follow-Up / Top-Up Amendment now soft-deletes the plan together with its payments, releasing the households back to the eligibility pools
- Only the most recent non-deleted plan of its category (source plan + plan type) can be deleted
- A Top-Up with a non-deleted Amendment cannot be deleted — the Amendment goes first
- Regular plans (OPEN → DRAFT) and targeting (TP_OPEN) behavior unchanged

### DIFF
--- a/src/hope/apps/payment/services/payment_plan_services.py
+++ b/src/hope/apps/payment/services/payment_plan_services.py
@@ -945,6 +945,9 @@ class PaymentPlanService:
         ]:
             raise ValidationError("Deletion is only allowed when the status is 'Open'")
 
+        if self.payment_plan.plan_type != PaymentPlan.PlanType.REGULAR:
+            return self._delete_child_plan()
+
         if self.payment_plan.status == PaymentPlan.Status.OPEN:
             if self.payment_plan.program_cycle.payment_plans.count() == 1:
                 # if it's the last Payment Plan in this Cycle need to update Cycle status
@@ -962,6 +965,39 @@ class PaymentPlanService:
         self.payment_plan.save()
 
         return self.payment_plan
+
+    def _delete_child_plan(self) -> PaymentPlan:
+        """Soft-delete a child plan (Top-Up / Follow-Up / Top-Up Amendment) together with its payments.
+
+        Child plans were never a targeting, so pushing them back to DRAFT (the Regular-plan delete
+        path) would leave them on the Targeting list and keep their payments blocking the source
+        plan's eligibility pools. Only the most recent non-removed plan of its category may be
+        deleted, so the remaining plans stay chronologically consistent.
+        """
+        payment_plan = self.payment_plan
+        if (
+            PaymentPlan.objects.filter(
+                source_payment_plan=payment_plan.source_payment_plan,
+                plan_type=payment_plan.plan_type,
+                created_at__gt=payment_plan.created_at,
+            )
+            .exclude(pk=payment_plan.pk)
+            .exists()
+        ):
+            raise ValidationError(
+                f"Only the most recent {payment_plan.get_plan_type_display()} of this Payment Plan can be deleted"
+            )
+        if (
+            payment_plan.plan_type == PaymentPlan.PlanType.TOP_UP
+            and PaymentPlan.objects.filter(
+                source_payment_plan=payment_plan, plan_type=PaymentPlan.PlanType.TOP_UP_AMENDMENT
+            ).exists()
+        ):
+            raise ValidationError("Cannot delete a Top Up that has a Top Up Amendment; delete the Amendment first")
+
+        payment_plan.payment_items.all().delete()
+        payment_plan.delete()
+        return payment_plan
 
     def export_xlsx(self, user_id: str) -> PaymentPlan:
         flow = PaymentPlanFlow(self.payment_plan)
--- a/src/hope/apps/payment/services/payment_plan_services.py
+++ b/src/hope/apps/payment/services/payment_plan_services.py
@@ -967,6 +967,7 @@ class PaymentPlanService:
         return self.payment_plan
 
     def _delete_child_plan(self) -> PaymentPlan:
+        """Soft-delete the most recent child plan (Top-Up / Follow-Up / Amendment) together with its payments."""
         payment_plan = self.payment_plan
         if (
             PaymentPlan.objects.filter(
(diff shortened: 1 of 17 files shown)


src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -963,6 +966,33 @@ def delete(self) -> PaymentPlan:
 
         return self.payment_plan
 
+    def _delete_child_plan(self) -> PaymentPlan:
```
Jeeziorny, line 969: Optional: instead of having those big ifs can we make some helper predicates functions so everything can be read as a book?

Jeeziorny (dismissed):
Very cool Jan

src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -963,6 +966,39 @@ def delete(self) -> PaymentPlan:
...
         return self.payment_plan
 
+    def _has_newer_sibling_plan(self) -> bool:
+        return (
+            PaymentPlan.objects.filter(
+                source_payment_plan=self.payment_plan.source_payment_plan,
+                plan_type=self.payment_plan.plan_type,
+                created_at__gt=self.payment_plan.created_at,
+            )
+            .exclude(pk=self.payment_plan.pk)
```
pkujawa: it does not exclude anything, since `created_at__gt=self.payment_plan.created_at,`

src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -963,6 +966,39 @@ def delete(self) -> PaymentPlan:
...
+            .exists()
+        )
+
+    def _is_top_up_with_amendment(self) -> bool:
+        return (
+            self.payment_plan.plan_type == PaymentPlan.PlanType.TOP_UP
+            and PaymentPlan.objects.filter(
+                source_payment_plan=self.payment_plan, plan_type=PaymentPlan.PlanType.TOP_UP_AMENDMENT
+            ).exists()
+        )
```
pkujawa: Actually this case will never exist - only OPEN plans can be deleted and an OPEN one cannot have an amendment.

tests/unit/apps/payment/test_delete_child_payment_plans.py
```diff
@@ -0,0 +1,285 @@
...
+
+    assert deleted.is_removed is True
+
+
+def test_delete_top_up_with_active_amendment_raises(top_up: PaymentPlan, amendment: PaymentPlan) -> None:
+    with pytest.raises(ValidationError) as error:
+        PaymentPlanService(payment_plan=top_up).delete()
+
+    assert "delete the Amendment first" in str(error.value)
+    assert PaymentPlan.objects.filter(pk=top_up.pk).exists()
```
pkujawa: this tests the case for `_is_top_up_with_amendment` but this is not a possible case - an OPEN Top Up cannot have an amendment

src/hope/apps/payment/services/payment_plan_services.py
```diff
@@ -963,6 +966,39 @@ def delete(self) -> PaymentPlan:
...
+        """Soft-delete the most recent child plan (Top-Up / Follow-Up / Amendment) together with its payments."""
+        payment_plan = self.payment_plan
+        if self._has_newer_sibling_plan():
+            raise ValidationError(
+                f"Only the most recent {payment_plan.get_plan_type_display()} of this Payment Plan can be deleted"
+            )
+        if self._is_top_up_with_amendment():
+            raise ValidationError("Cannot delete a Top Up that has a Top Up Amendment; delete the Amendment first")
+
+        payment_plan.payment_items.all().delete()
```
pkujawa: I think this can race with `prepare_child_payment_plan_async_task_action`. The task runs after the plan is created, and `delete()` only looks at status, so you can delete a plan while its payments are still being copied. The task takes the plan at the start and then locks the source plan, not this one, so nothing blocks the delete. It commits, and then the re-entry guard in `_copy_payments` sees no payments and `bulk_creates` them onto a plan that's already removed. You end up with live payments hanging off a deleted plan, invisible in the UI, still blocking those households from any new child plan. Which is the thing we're fixing here.

Taking the same lock here would stop the two overlapping:

```
    if payment_plan.source_payment_plan_id:
        PaymentPlan.objects.select_for_update().get(pk=payment_plan.source_payment_plan_id)
```

and the task should re-read `is_removed` once it has the lock, so it gives up if the plan is gone.

pkujawa (changes requested):

pkujawa:
The delete button is shown for any OPEN plan — top-ups, amendments and follow-ups all render `FollowUpPaymentPlanDetailsHeader`, where `canRemove` is just `hasPermissions(PM_CREATE)`.
Maybe we could add a can_delete flag?

pkujawa:
Also, I think neither of the two new error messages actually reaches the user. The BE is fine, but `DeletePaymentPlan.tsx `does `showMessage(e.message)`, and the generated client puts 'Bad Request' in message and keeps the real text in `e.body`.         
                                                                                                                                                                                                                                                                                                                             
  Everywhere else we go through `showApiErrorMessages` from utils, which digs into e.body. This dialog just doesn't.                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                             
  ```js                                                                                                                                                                                                                                                                                                                      
  import { showApiErrorMessages } from '@utils/utils';                                                                                                                                                                                                                                                                       
  ...                                                                                                                                                                                                                                                                                                                        
  showApiErrorMessages(e, showMessage, t('Failed to delete the Payment Plan'));                                                                                                                                                                                                                                              
  ```

pkujawa (dismissed):

PR #6337, opened 2026-08-26
[FEATURE] 334130: Join external URLs through a single helper

nerjan:
[AB#334130](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/334130)
**Added `hope.apps.utils.external_urls`** with `normalize_base_url` and `build_url`, which join a base and a path with exactly one slash while leaving the path's trailing slash and query string intact.

**Routed every external URL join through it** — KoBo, RapidPro, Vision, the `BaseAPI` subclasses, the Aurora fetch form in `RecordAdmin`, the `export_encodings` dedup client, and the biometric-deduplication notification URL.


tests/unit/admin/test_record_admin.py
```diff
@@ -0,0 +1,68 @@
...
+        "username": "aurora-user",
+        "password": "aurora-password",
+        "registration": 2,
+        "start": 10,
+        "end": 20,
+    }
+
+
+@patch("hope.admin.record.requests.get")
+def test_fetch_builds_url_from_host_without_trailing_slash(
```
Jeeziorny: maybe we could use pytest parametrisation here?

tests/unit/apps/utils/test_external_urls.py
```diff
@@ -0,0 +1,63 @@
+from hope.apps.utils.external_urls import build_url, normalize_base_url
+
+
+def test_normalize_base_url_strips_trailing_slash() -> None:
```
Jeeziorny: Looks like good candidates for parametrisation as well

Jeeziorny:
Other places to be considered as well:
1. `src/hope/contrib/aurora/utils.py:163` - we assume that `record_url` ends with ‘/’
2. `src/hope/contrib/aurora/utils.py:174` - SAA
3. `src/hope/apps/core/exchange_rates/api.py:40` - it’s not a child of BaseAPI, but maybe we could normalise api_url as well here?
4. `src/hope/apps/payment/utils.py:323`,
5. `src/hope/apps/grievance/utils.py:33`,
6. `src/hope/apps/payment/notifications.py:119`
7. `src/hope/apps/periodic_data_update/notifications.py:109`

Questions
1. Should we normalise API_URL
2. WDYT about adding warning log at `build_url`, when base_url is empty? I feel like this could improve traceability, but I’m looking for edge cases that this could make too many logs in some valid cases
3. Do you think that some pre commit hook to guard reintroducing such pattern in codebase is worth considering? It does not necessarily needs to be done in this PR, but.. Considering we put an effort to introduce convention and fix some stuff it might be reasonable to have some pygrep hook or ruff TID251 rule enforment in precommit hook. @kosior @nerjan wdyt

kosior (dismissed):
Awesome! Great idea with the `frontend_url`!

Jeeziorny (changes requested):
Seems that test changes and log in build_url when base is empty are not done yet. Let's wait for @nerjan to explicitly tell this PR is ready to review. FYI @domdinicola

nerjan:
@Jeeziorny  In the log — I dug into it and I don't think it can fire. BaseAPI.__init__ already raises on an empty api_url before build_url is reached, and the other callers' settings have non-empty defaults in env.py/ should never be empty inr eality.
According to tests I parametrised them.
PR is ready for review/merge.

kosior (approved):

Jeeziorny (approved):

PR #6338, opened 2026-08-26
Scope the program whose field attributes are read, per action

johniak:
Split out of #6309. **Merge this one first** — #6309's accountability and payment splits build on it.

## Problem

Two separate holes in the shared layer.

**Field attributes.** `all_fields_attributes` took a `program_id` from the query string and passed it to the generator without comparing it with the business area of the path, so the field attributes of any program could be read.

**Wrong permission set.** `BusinessAreaProgramsAccessMixin` narrowed the queryset to the programs the user holds `self.PERMISSIONS` in — that is the *list* permission. A detail or write action requires a different one, so the narrowing was done against the wrong set. It now uses the permissions of the action actually being served.

## Fix

- the program named in the query string is checked against the business area of the path before it is used
- the mixin narrows by `get_permissions_for_action()` instead of the class-level list

The second change is not cosmetic: the feedback update endpoint answers 404 for a legitimate request without it. Used by `payment`, `accountability` and `activity_log`.

Part of GHSA-2xf8-jjc2-9pmv.

### DIFF
--- a/src/hope/apps/core/api/mixins.py
+++ b/src/hope/apps/core/api/mixins.py
@@ -204,7 +204,7 @@ class BusinessAreaProgramsAccessMixin(BusinessAreaMixin):
 
         program_ids = self.request.user.get_program_ids_for_permissions_in_business_area(
             self.business_area.id,
-            self.PERMISSIONS,
+            self.get_permissions_for_action(),
         )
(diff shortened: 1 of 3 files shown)


johniak:

PR #6339, opened 2026-08-26
Scope a periodic data update upload to the program of the url path

johniak:

## Problem

The upload reads the template id out of the **uploaded workbook**, not out of a serializer field, so nothing compared it with the program in the url path.

- a template exported from another program was accepted
- the rows in that file were then applied to the individuals it named
- the id never passes through a relation field, so no field-level scoping could have caught it

## Fix

- the template lookup is scoped to the program of the url path
- the individual lookup in each row is scoped to the program of the template

The second one matters on its own: without it a template of the right program could still name an individual of another one.

Both new tests fail when the matching filter is reverted.

Part of GHSA-2xf8-jjc2-9pmv.


johniak:

PR #6340, opened 2026-08-26
Scope the beneficiary that names the program in partner choices

johniak:

## Problem

`UserViewSet.partner_for_grievance_choices` takes the program from an id the caller supplies, with a global `.get()` that never compared the object with the business area in the url path.

- a foreign `household` or `individual` id named a foreign program, and that program shaped the response
- a missing id raised `DoesNotExist` rather than answering 404, so the endpoint returned a 500

## Fix

Both lookups are scoped to the business area of the path and use `get_object_or_404`. An id outside the path now fails the same way an id that does not exist fails.

Part of GHSA-2xf8-jjc2-9pmv.

### DIFF
--- a/src/hope/apps/account/api/views.py
+++ b/src/hope/apps/account/api/views.py
@@ -183,11 +183,11 @@ class UserViewSet(
         individual_id = request.query_params.get("individual")
 
         if program_code:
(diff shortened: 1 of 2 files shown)


johniak:

PR #6341, opened 2026-08-26
Scope the program and the import file a registration import starts from

johniak:

## Problem

Three lookups took a caller-supplied id and never compared it with the business area in the url path.

| Action | Id from the body | What it reached |
|---|---|---|
| `create-program-population-import` | `import_from_program_id` | a program of another business area, whose population was then pulled in |
| `create` (xlsx) | `import_data_id` | an `ImportData` uploaded by another business area |
| `create-kobo` | `import_data_id` | a `KoboImportData` uploaded by another business area |

## Fix

- the source program is resolved in `validate_import_from_program_id`, where the business area is already in the serializer context, and the view reads the resolved object instead of looking it up again
- both import-data lookups are filtered by the `business_area_slug` of the path
- the second `.get()` of the same import data in each action is dropped — the row was already fetched, and fetching it again unfiltered would have undone the scoping

Part of GHSA-2xf8-jjc2-9pmv.

### DIFF
--- a/src/hope/apps/registration_data/api/serializers.py
+++ b/src/hope/apps/registration_data/api/serializers.py
@@ -4,12 +4,13 @@ from typing import Any
 
 from django.db.models import Count, Q
 from rest_framework import serializers
+from rest_framework.generics import get_object_or_404
 from rest_framework.validators import UniqueValidator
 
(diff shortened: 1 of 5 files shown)


johniak:

PR #6343, opened 2026-08-26
Scope a delegation to the import of the url path

johniak:

## Problem

The delegate-people endpoint took individual ids from the body and reassigned the primary collector role by id alone, so ids belonging to another import — and so to another business area — were reassigned through a url scoped to the caller's.

## Fix

- the RDI of the path is resolved once, and the role queryset is filtered by it
- delegates are validated against that import in `validate_delegates`, before any role is written, which also collapses one `exists()` per delegate into one query for the batch
- the writes run in one transaction

The last two points address @MarekBiczysko's review on #6309: the original check sat inside the write loop, so a bad delegate in the middle of a batch left the earlier ones already reassigned behind a 400.

Part of GHSA-2xf8-jjc2-9pmv.

### DIFF
--- a/src/hope/api/endpoints/rdi/delegate_people.py
+++ b/src/hope/api/endpoints/rdi/delegate_people.py
@@ -1,13 +1,16 @@
 from typing import TYPE_CHECKING
 from uuid import UUID
 
+from django.db import transaction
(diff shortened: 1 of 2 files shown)


johniak:

PR #6344, opened 2026-08-26
Scope the beneficiaries and the program a feedback points at

johniak:
Split out of #6309. **Based on `feature/fix-cross-ba-idor-core`** — the base retargets to develop once that merges.

## Problem

`FeedbackCrudServices._apply_fields` resolved `household_lookup` and `individual_lookup` with a global `get_object_or_404`. A feedback created in one business area attached beneficiaries of another and was then dragged into their program — this is the PoC @Jeeziorny posted on #6309.

The program was resolved the same unscoped way, from both the `program_id` of the body and the `program_code` of the path, and `get_object` bypassed the queryset that scopes the list.

## Fix

- `household_lookup` and `individual_lookup` are scoped to the feedback's business area
- `admin2` stays global on purpose — admin areas are shared reference data, not owned by a business area
- both program lookups are scoped to the business area of the path
- `get_object` goes through `get_queryset()`
- the message and survey actions read `self.business_area` / `self.program` instead of re-fetching them unscoped

Part of GHSA-2xf8-jjc2-9pmv.

### DIFF
--- a/src/hope/apps/accountability/api/views.py
+++ b/src/hope/apps/accountability/api/views.py
@@ -116,7 +116,7 @@ class FeedbackViewSet(
     program_model_field = "program"
 
     def get_object(self) -> Feedback:
-        return get_object_or_404(Feedback, id=self.kwargs.get("pk"))
+        return get_object_or_404(self.get_queryset(), id=self.kwargs.get("pk"))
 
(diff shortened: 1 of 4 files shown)


johniak:

PR #6345, opened 2026-08-26
Scope payment plans, verifications and their documents to the url path

johniak:
Split out of #6309. **Based on `feature/fix-cross-ba-idor-core`** — the base retargets to develop once that merges.

## Problem

Every lookup under a payment plan took an id and never compared it with the program of the path: the plan itself, its verification plans, its verification records, and its supporting documents.

`PaymentPlanService.create` took a business area slug and read the program from the cycle the caller named — so a cycle of another program decided which program the plan landed in.

`update` was not atomic: a foreign provider rejected *after* the targeting rules were rewritten left the rules replaced and the plan unsaved.

## Fix

- `ScopedPaymentPlanMixin` resolves the plan of the path once, and the verification plans, records and documents hang off it
- `PaymentPlanService.create` takes the program of the path and scopes the cycle to it — which also drops two queries per call
- `update` is wrapped in a transaction, addressing @MarekBiczysko's review on #6309
- `PaymentVerificationLogEntryFilter` is deleted; no view referenced it

The one payment change that needs the `ScopedRelatedField` mechanism — the `cycle` field on `PaymentPlanGroupCreateSerializer` — stays in #6309.

Part of GHSA-2xf8-jjc2-9pmv.

### DIFF
--- a/src/hope/apps/payment/api/serializers.py
+++ b/src/hope/apps/payment/api/serializers.py
@@ -89,8 +89,7 @@ class PaymentPlanSupportingDocumentSerializer(serializers.ModelSerializer):
         return file
 
     def validate(self, data: dict) -> dict:
-        payment_plan_id = self.context["request"].parser_context["kwargs"]["payment_plan_pk"]
-        payment_plan = get_object_or_404(PaymentPlan, id=payment_plan_id)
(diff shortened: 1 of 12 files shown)


johniak:

PR #6346, opened 2026-08-27
Feature/partner assignment enhancement

domdinicola:


src/hope/admin/utils.py
```diff
@@ -150,6 +150,22 @@ def get_raw_id_fields(self, request: HttpRequest) -> list[str]:
         return list(result)
 
 
+class AutocompleteExcludeFieldsMixin:
```
pkujawa: Every class using this will always inherit `AutocompleteForeignKeyMixin`, so could we implement it in that mixin and this one goes away? I don't see a need for a new one

```
    class AutocompleteForeignKeyMixin:
        autocomplete_exclude_fields: tuple[str, ...] = ()

        def get_autocomplete_fields(self, request: HttpRequest) -> list[str]:
            result = set(super().get_autocomplete_fields(request))
            autocomplete, _ = self._related_admin_fields(request)
            result.update(autocomplete)
            return list(result - set(self.autocomplete_exclude_fields))
```
(also safer because if someone put the new mixin after `AutocompleteForeignKeyMixin`, then the exclusion is silently undone — `AutocompleteForeignKeyMixin.get_autocomplete_fields` re-adds every FK with `search_fields`, including the excluded ones.)

pkujawa (changes requested):

pkujawa (approved):

tests/unit/apps/account/test_admin.py
```diff
@@ -389,6 +390,28 @@ def test_partner_role_assignment_admin_business_area_not_autocomplete(
...
+    request = get_mock_request(request_factory)
+    fields = admin.get_autocomplete_fields(request)
+    assert "parent" not in fields
+
+
+def test_project_admin_programme_not_autocomplete(
+    request_factory: RequestFactory,
+    admin_site: AdminSite,
+):
+    from hope.contrib.aurora.models import Project
```
nerjan, line 407: import should go to top of the file.

nerjan (approved):

PR #6353, opened 2026-08-28
Add ScopedRelatedField and scope core field attributes per action

johniak:
Split out of #6309 — bottom of the stack. The layers above build on `ScopedRelatedField` introduced here.

## The mechanism

`ScopedRelatedField` narrows a writable related field to the scope of the url path — business area or program — so an id outside it fails like an id that does not exist (400, same message). The layers above use it on 35 grievance fields, 8 accountability fields, the token API serializers and one payment field. A contract test that walks every REST route and fails on any unscoped writable relation field sits at the top of the stack, where it can pass.

## Two holes in the shared layer

**Field attributes.** `all_fields_attributes` took a `program_id` from the query string and passed it to the generator without comparing it with the business area of the path, so the field attributes of any program could be read.

**Wrong permission set.** `BusinessAreaProgramsAccessMixin` narrowed the queryset to the programs the user holds `self.PERMISSIONS` in — that is the *list* permission. A detail or write action requires a different one, so the narrowing was done against the wrong set. It now uses the permissions of the action actually being served. Not cosmetic: the feedback update endpoint answers 404 for a legitimate request without it. Used by `payment`, `accountability` and `activity_log`.

Part of GHSA-2xf8-jjc2-9pmv. Replaces closed #6338.

### DIFF
--- a/src/hope/apps/core/api/fields.py
+++ b/src/hope/apps/core/api/fields.py
@@ -1,0 +1,34 @@
+from typing import Any
+
+from django.db.models import QuerySet
+from rest_framework import serializers
+
+
+class ScopedRelatedFieldMixin:
+    """Narrows the queryset to the scope of the url path, so an id from outside of it fails like a nonexistent one.
+
+    The field relies on two things: the view has to put the scope object in the serializer context
+    under the `scope` key (`BusinessAreaMixin`/`ProgramMixin` do), and `scope_path` has to be a valid
+    ORM path from the queryset model to that object (defaults to the `scope` name itself).
+    """
+
+    def __init__(self, *args: Any, scope: str = "business_area", scope_path: str | None = None, **kwargs: Any) -> None:
+        self.scope = scope
+        self.scope_path = scope_path or scope
+        super().__init__(*args, **kwargs)
+
+    def get_queryset(self) -> QuerySet:
+        if self.scope not in self.context:
+            raise KeyError(
+                f"'{self.scope}' is missing from the serializer context;"
+                f" the view has to provide it for {type(self).__name__}"
+            )
+        return super().get_queryset().filter(**{self.scope_path: self.context[self.scope]})
+
+
+class ScopedRelatedField(ScopedRelatedFieldMixin, serializers.PrimaryKeyRelatedField):
+    pass
+
+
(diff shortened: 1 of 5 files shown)


MarekBiczysko (approved):

src/hope/apps/core/api/fields.py
```diff
@@ -0,0 +1,24 @@
+from typing import Any
+
+from django.db.models import QuerySet
+from rest_framework import serializers
+
+
+class ScopedRelatedFieldMixin:
+    """Narrows the queryset to the scope of the url path, which the view has to put in the context."""
+
+    def __init__(self, *args: Any, scope: str = "business_area", scope_path: str | None = None, **kwargs: Any) -> None:
```
MarekBiczysko, line 15: I'm not user about this "business_area" default argument, maybe it should be nullable and flter qs only if provided explicitly

src/hope/apps/core/api/fields.py
```diff
@@ -0,0 +1,24 @@
...
+        self.scope = scope
+        self.scope_path = scope_path or scope
+        super().__init__(*args, **kwargs)
+
+    def get_queryset(self) -> QuerySet:
+        return super().get_queryset().filter(**{self.scope_path: self.context[self.scope]})
```
MarekBiczysko, line 26: This requires two things:

serializer context with the proper key (by default, "business_area")
get_queryset() base model must have existing relation (by default, "business_area")

Maybe this should be documented or validated to ensure these two requirements are met.
Otherwise, we have 3 different parts of code that are intertwined in a not transparent way.

MarekBiczysko (changes requested):

src/hope/apps/core/api/fields.py
johniak: I would prefer to keep it like this. If default is no filtering then ScopedRelatedField with no args works same as plain PrimaryKeyRelatedField, so silently unscoped, and this is exactly the bug this whole series is fixing. Default scope is on purpose, so you cannot forget to scope it. If some field really need global lookup you just use plain DRF field, and contract test on top of the stack will force you to add it to allow list with a reason. So it is always explicit decision, not accident.

MarekBiczysko (approved):

PR #6342, opened 2026-08-26
Bind a grievance document and a bulk action to the ticket of the url path

johniak:
Split out of #6309. Independent — touches no shared code.

## Problem

**Documents.** `update_grievance_documents` took a document id straight from the body and looked it up globally. An edit on your own ticket could point at a document of someone else's ticket — the file on disk was removed and replaced.

**Bulk actions.** Assign, priority, urgency and add-note filtered tickets by `id__in` alone. Ids of another business area, posted to an endpoint scoped to the caller's, were updated.

## Fix

- `update_grievance_documents` takes the ticket id and filters on it
- the four bulk actions share one `_open_tickets()` helper that filters by `business_area__slug` of the path

Part of GHSA-2xf8-jjc2-9pmv.

PR #6354, opened 2026-08-28
Scope grievance tickets, documents and bulk actions to the url path

johniak:
Split out of #6309 — layer 2 of the stack, on the `ScopedRelatedField` layer below.

## Problem

**Ticket create and update.** The grievance ticket serializer takes 35 writable relation fields — individuals, households, payments, documents, linked tickets, linked feedback — all looked up by bare id. Proven exploitable from the body: a ticket created with a foreign individual (201), a ticket linked to a foreign feedback (201, and the foreign feedback row was written to).

**Documents.** `update_grievance_documents` took a document id straight from the body and looked it up globally. An edit on your own ticket could point at a document of someone else's ticket — the file on disk was removed and replaced.

**Bulk actions.** Assign, priority, urgency and add-note filtered tickets by `id__in` alone. Ids of another business area, posted to an endpoint scoped to the caller's, were updated — the business area of the url was used only to clear a cache.

## Fix

- the 35 relation fields become `ScopedRelatedField`s, scoped to the business area or program of the path
- `update_grievance_documents` takes the ticket id and filters on it
- the four bulk actions share one `_open_tickets()` helper that filters by `business_area__slug` of the path

Part of GHSA-2xf8-jjc2-9pmv. Replaces closed #6342.

### DIFF
--- a/src/hope/apps/grievance/api/mixins.py
+++ b/src/hope/apps/grievance/api/mixins.py
@@ -483,7 +483,7 @@ class GrievanceMutationMixin:
             delete_grievance_documents(ticket.id, ids_to_delete)
         if documents_to_update := input_data.pop("documentation_to_update", None):
             validate_grievance_documents_size(ticket.id, documents_to_update, is_updated=True)
-            update_grievance_documents(documents_to_update)
+            update_grievance_documents(ticket.id, documents_to_update)
         if documents := input_data.pop("documentation", None):
             validate_grievance_documents_size(ticket.id, documents)
             create_grievance_documents(approver, ticket, documents)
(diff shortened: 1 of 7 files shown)


src/hope/apps/grievance/api/views.py
```diff
@@ -398,6 +398,9 @@ class GrievanceTicketGlobalViewSet(
 
     parser_classes = (DictDrfNestedParser, JSONParser)
 
+    def get_serializer_context(self) -> dict:
```
MarekBiczysko: why don't extend BusinessAreaMixin to propagate context like in ProgramMixin

MarekBiczysko (changes requested):

PR #6356, opened 2026-08-28
Scope feedback, surveys and messages to the url path

johniak:
Split out of #6309 — layer 4 of the stack, on the `ScopedRelatedField` layer below.

## Problem

**Feedback.** `FeedbackCrudServices._apply_fields` resolved `household_lookup` and `individual_lookup` with a global `get_object_or_404`. A feedback created in one business area attached beneficiaries of another and was then dragged into their program — this is the PoC @Jeeziorny posted on #6309. The program was resolved the same unscoped way, from both the `program_id` of the body and the `program_code` of the path, and `get_object` bypassed the queryset that scopes the list.

**Surveys and messages.** Both accepted foreign household ids in the body — 201, and the message was already on its way to RapidPro.

## Fix

- the survey and message serializers scope their household and payment-plan fields with `ScopedRelatedField`
- `household_lookup` and `individual_lookup` are scoped to the feedback's business area
- `admin2` stays global on purpose — admin areas are shared reference data, not owned by a business area
- both program lookups are scoped to the business area of the path
- `get_object` goes through `get_queryset()`
- the message and survey actions read `self.business_area` / `self.program` instead of re-fetching them unscoped

Part of GHSA-2xf8-jjc2-9pmv. Replaces closed #6344.

### DIFF
--- a/src/hope/apps/accountability/api/views.py
+++ b/src/hope/apps/accountability/api/views.py
@@ -114,7 +114,7 @@ class FeedbackViewSet(
     program_model_field = "program"
 
     def get_object(self) -> Feedback:
-        return get_object_or_404(Feedback, id=self.kwargs.get("pk"))
+        return get_object_or_404(self.get_queryset(), id=self.kwargs.get("pk"))
 
     def get_queryset(self) -> QuerySet[Feedback]:
         queryset = super().get_queryset()
@@ -136,10 +136,10 @@ class FeedbackViewSet(
         program_code = self.kwargs.get("program_code")
         program = None
         if program_code:
-            program = Program.objects.get(code=program_code)
+            program = get_object_or_404(Program, code=program_code, business_area=business_area)
 
         if program_id := serializer.validated_data.get("program_id"):
-            program = Program.objects.get(id=program_id)
+            program = get_object_or_404(Program, id=program_id, business_area=business_area)
 
         if program and program.status == Program.FINISHED:
             raise ValidationError("It is not possible to create Feedback for a Finished Program.")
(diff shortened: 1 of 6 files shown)


MarekBiczysko (dismissed):

MarekBiczysko (dismissed):

src/hope/apps/accountability/api/views.py
```diff
@@ -136,10 +136,10 @@ def create(self, request: Request, *args: Any, **kwargs: Any) -> Response:
         program_code = self.kwargs.get("program_code")
         program = None
         if program_code:
-            program = Program.objects.get(code=program_code)
+            program = get_object_or_404(Program, code=program_code, business_area=business_area)
 
         if program_id := serializer.validated_data.get("program_id"):
-            program = Program.objects.get(id=program_id)
+            program = get_object_or_404(Program, id=program_id, business_area=business_area)
```
nerjan, line 141: This path is unsafe, as if it is program path (with program_code in url) to which permissions are checked AND program_id in the body, then program is taken from the body even when User has no permission to this program.
I think simple elif would fix this.

src/hope/apps/accountability/services/feedback_crud_services.py
```diff
@@ -49,7 +51,7 @@ def create(
             issue_type=input_data["issue_type"],
             description=input_data["description"],
         )
-        cls._apply_fields(obj, input_data)
+        cls._apply_fields(obj, input_data, business_area)
 
         if obj.household_lookup:
             obj.program = obj.household_lookup.program or obj.household_lookup.programs.first()
```
nerjan: Similar situation as in previous comment, this program WAS NOT authorized in the view.

nerjan (commented):
I think comments were not addressed

nerjan (approved):

PR #6363, opened 2026-09-01
[Grievance] Fix data loss on ticket update, 500 on household data tickets, and the programme filter

pkujawa:
Six pre-existing bugs in the grievance update endpoint and the grievance list.                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                             
  ### Fixed                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                             
  - **500 on saving a Household Data Update ticket.** The edit form stopped sending household data extras in Aug 2025; the service read them without checking.                                                                                                                                                               
  - **Updates deleted the ticket's linked tickets** when the request didn't include them - Household Data Update, Payment Verification and Biometric Photo Error tickets.                                                                                                      
  - **Updates unassigned the ticket** when the request didn't include `assigned_to` - Payment Verification tickets.                                                                           
  -  **Any edit of the ticket cleared the approvals on its requested data**, even though the edit form cannot change that data — Individual Data Update and Add Individual tickets.                                                                                                                                                                                                                                                
  - **The Programme filter on the grievance list did nothing.** The page sent it under the wrong key, and the dropdown gives a programme id while the backend matched on the code.                                                                                   
  - **No backend validation on the submission channel.** The dropdown already hides "HOPE Generated" on user-generated tickets; this adds the backend check.
                                                                                                                                                                                                                                                                                                                             
  ### How                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                             
  - The data-change services leave the stored request alone when the update carries no data change. A real change behaves as before, approvals included.                                                                                                                                                                     
  - The edit form stops sending requested data it cannot change — every control for it is hidden on the edit page. Photo replacement has its own page and still works.                                                                                                                                                       
  - Linked tickets and the assignee only change when the request actually contains them. An empty list still clears links; null still unassigns.                                                                                                                                                                             
  -  The list now sends the selected programme in the `program` filter as an id, and the backend filters by id.                                                                                                                                                                                                                                                                                                                                                                      
  - "HOPE Generated" is rejected unless the ticket is system-generated, via a new `GrievanceTicket.is_system_generated`.


johniak (dismissed):

src/frontend/src/components/grievances/utils/editGrievanceUtils.ts
```diff
@@ -303,81 +289,13 @@ function prepareDeleteIndividualVariables(requiredVariables, values) {
...
-      }
-      // Handle direct structure {country, key, number} for backward compatibility
-      return item;
-    });
-  };
-
   return {
     ...requiredVariables,
     linkedTickets: values.selectedLinkedTickets,
-    extras: {
```
kosior, line 336: Should we also update [UPDATE_EXTRAS_OPTIONS](https://github.com/unicef/hope/blob/fix/hh-data-update-edit-500/src/hope/apps/grievance/api/mixins.py#L434)?

or even better remove it completely?
[This condition](https://github.com/unicef/hope/blob/fix/hh-data-update-edit-500/src/hope/apps/grievance/api/views.py#L573) looks like a dead code. The `issue_type` is not a field on UpdateGrievanceTicketSerializer and I think that's the only place where we use `UPDATE_EXTRAS_OPTIONS`

This could be in a follow-up PR, but I'm not sure so I decided to leave a comment.

pkujawa: Good catch - confirmed dead and removed in this PR.                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                             
(I dug deeper and found that [PR #2463](https://github.com/unicef/hope/pull/2463) deliberately disabled editing the requested-data sections on the edit page for household, individual and add-individual tickets.
The payload stayed as it was, and that's where both bugs here come from: the household form later stopped sending its extras while the backend still read them, hence the 500, and the individual and add-individual forms kept sending them while the backend kept rebuilding the requested data from that echo, hence the approvals reset.)

kosior (approved):

johniak (approved):

PR #6369, opened 2026-09-02
Upgrade djangorestframework to 3.18

domdinicola:
## Summary

Upgrades `djangorestframework` from `>=3.13.1,<3.18` to `>=3.13.1` (now resolving to 3.18.0) and adapts the codebase to the DRF 3.18 breaking change.

## Breaking change addressed

DRF 3.18 changed how `ListSerializer` (`many=True`) reports validation errors. The error response flipped from a **list padded with empty dicts** to a **dict keyed by failing item index**:

```json
3.17.x: [{}, {"code": ["This field is required."]}, {}]
3.18.0: {"1": {"code": ["This field is required."]}}
```

Only failing indices are present, so the JSON type of affected error responses changes from array to object.

## Changes

- **`src/hope/api/utils.py`**: refactored `humanize_errors()` into `_humanize_households` / `_humanize_members` / `_humanize_members_info` helpers, which now handle the new dict-keyed `households`/`members` error shape (accurate item indices) while remaining backward-compatible with the list form used by the lax RDI path.
- **Tests**: updated assertions that pinned the old list format:
  - `test_upload.py` — upload validation errors (`Household #N` now keyed dict)
  - `test_push_people.py` — `many=True` people errors, including nested `documents`
  - `test_push_people_country_workspace_id.py` — skipped CW error-format tests
  - `test_humanize.py` — added coverage for the new dict format
- **`pyproject.toml` / `uv.lock`**: unpin DRF `<3.18` and refresh lockfile to DRF 3.18.0.

## Verification

- Affected test modules: `61 passed, 5 skipped`
- `ruff check`, `ruff format`, and `mypy` all pass on changed files.


src/hope/api/utils.py
```diff
@@ -31,6 +16,46 @@ def humanize_errors(errors: dict) -> dict:
...
+                _humanize_members(h)
+            if h:
+                hh_info_dict[f"Household #{idx}"] = [h]
+        return hh_info_dict
+    hh_info_list: list[dict[str, Any]] = []
+    for i, h in enumerate(households, 1):
+        if h and isinstance(h, dict):
+            _humanize_members(h)
+        if h:
+            hh_info_list.append({f"Household #{i}": [h]})
```
kosior, line 48: Was the household renumbering intentional? The first household used to be Household #1, now it's Household #0. Is that label user-facing?

pyproject.toml
```diff
@@ -66,7 +66,7 @@ dependencies = [
   "django-strategy-field>=3,<4",
   "django-sysinfo>=2.6,<3",
   "django-viewflow>=2.2.15",
-  "djangorestframework>=3.13.1,<3.18",
+  "djangorestframework>=3.13.1",
```
kosior:
```suggestion
  "djangorestframework>=3.18.0,<3.19",
```

I think it would be safer to keep the upper bound.

kosior (changes requested):

src/hope/api/utils.py
```diff
@@ -7,22 +7,7 @@ def humanize_errors(errors: dict) -> dict:
...
-                    mm_info: list | dict
-                    if isinstance(members, list) and len(members) == 1 and isinstance(members[0], str):
-                        mm_info = members
-                    else:
-                        mm_info = {f"Member #{i}": [m] for i, m in enumerate(members, 1) if m}
-                    if mm_info:
-                        h["members"] = mm_info
-                if h:
-                    hh_info.append({f"Household #{i}": [h]})
+        hh_info = _humanize_households(households)
```
johniak, line 18: `PushToRDIView` (`rdi/base.py:123`) passes a top-level `many=True` `serializer.errors` here. On 3.18 that's a `ReturnDict` with int keys, so `errors.pop("households", [])` finds nothing and `errs.update(**errors)` raises `TypeError: keywords must be strings`. `POST /rdi/<id>/push/` with `[{"village": "x"}]` returns 500. It was already a 500 on 3.17 (`list.pop("households", [])`), but this is the place to fix it: handle int-keyed top-level errors like the nested dict, and add a test posting an invalid list to `api:rdi-push`.

src/hope/api/utils.py
```diff
@@ -31,6 +16,46 @@ def humanize_errors(errors: dict) -> dict:
...
+        return [households]
+    if isinstance(households, list) and len(households) == 1 and isinstance(households[0], str):
+        return households
+    if isinstance(households, dict):
+        hh_info_dict: dict[str, Any] = {}
+        for idx, h in households.items():
+            if h and isinstance(h, dict):
+                _humanize_members(h)
+            if h:
+                hh_info_dict[f"Household #{idx}"] = [h]
```
johniak: `Household #{idx}` now uses DRF's 0-based index, so the upload API's first household changed from `Household #1` to `Household #0` (see the flipped assertions in `test_upload.py`), while the list branch below and the lax endpoint (`test_soft.py:282`) stay 1-based. Was the numbering change intended? `int(idx) + 1` here and on the `Member #` line keeps the old contract and makes both paths agree.

tests/unit/api/test_humanize.py
```diff
@@ -100,3 +100,55 @@ def test_empty_members_list(household_error: dict) -> None:
 
     hh = result["households"][0]["Household #1"][0]
     assert "members" not in hh
+
+
+def test_households_dict_single_error(household_error: dict) -> None:
+    errors = {"households": {"0": household_error}}
```
johniak: These dict-shape tests use string keys (`{"0": ...}`), which `ListSerializer.errors` never produces in-process – it's int keys, strings only appear after JSON serialisation. Use int keys so the tests exercise the real shape; with string keys they can't catch the top-level `TypeError` from `errs.update(**errors)`.

johniak (changes requested):

johniak (dismissed):

src/hope/api/utils.py
```diff
@@ -31,6 +24,46 @@ def humanize_errors(errors: dict) -> dict:
...
+        return [households]
+    if isinstance(households, list) and len(households) == 1 and isinstance(households[0], str):
+        return households
+    if isinstance(households, dict):
+        hh_info_dict: dict[str, Any] = {}
+        for idx, h in households.items():
+            if h and isinstance(h, dict):
+                _humanize_members(h)
+            if h:
+                hh_info_dict[f"Household #{int(idx) + 1}"] = [h]
```
kosior: I'm thinking if we need to worry about strings like `non_field_errors` here. Casting to int would give use ValueError. It would be caught by the try.except in the `humanize_errors`, but I think we wouldn't get a valid error information.

The same could happen in `_humanize_members_info`.

kosior (approved):

PR #6370, opened 2026-09-02
[HOTFIX] [338764] Fix exclusion crash on cycles with no end date 

pkujawa:
[AB#338764](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/338764)

  **Issues found**                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 

- Removing a household from the exclusion list crashed when the programme cycle had no end date.
- The crash put the plan into Exclude Beneficiaries Error but saved no message, so nothing in the UI or in admin said what went wrong.                                                                                                                                                                
- The error status disabled the Save and Apply buttons, so the user could not retry, undo, or clear the list. The backend accepts a retry, the UI just never sent one.                                                                                                               
                                                                                                                                                                                                                                                                                                                             
Two more issues found while investigating, not related to the crash

  - Saving without an exclusion reason returned 400. Empty should be accepted.                                                                                                                                                                                                                                               
  - A stored error message is not shown if no reason was filled in, or while the user is editing the list.                                                                                                                                                                                                               

                                                                                                                                                                                                                                                                                                                             
  **Applied fixes**                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          

  - Removed the two date comparisons in the hard conflict query. The query is already filtered to one cycle, so those lines compared the cycle against itself and never filtered anything. They only crashed when the end date was null.
  - Unexpected failures now save an error message instead of leaving the field empty.  
  - Unexpected failures are no longer retried. The job used to keep retrying for about 3 minutes, while the plan already showed the error status and Save was available again. A retry could then re-apply its own old list of IDs over whatever the user saved in the meantime. The task already treats a failure as final, it sets the error status and writes a message, so retrying underneath that was wrong. Transient failures no longer recover on their own, the user retries from the UI instead.
  it sets the error status and writes a message, so retrying underneath that was wrong. Transient failures no longer recover on their own, the user retries from the UI instead.                                                                                                                                                                                                                                    
  - Save and Apply now follow the same rule as the backend: an exclusion can start when there is no background action, or when the last exclusion failed. Added a tooltip explaining why the button is disabled.                                                                                                             
  - Empty exclusion reason is accepted.                                                                                                                                                                                                                                                                                      
  - The stored error message now always shows. It used to require an exclusion reason and disappear while editing the list.                                                                                                                                                                                                                                 
  - Moved the duplicated error status list into utils/constants.ts.

### DIFF
--- a/src/frontend/src/components/paymentmodule/PaymentPlanDetails/ExcludeSection/ExcludeSection.tsx
+++ b/src/frontend/src/components/paymentmodule/PaymentPlanDetails/ExcludeSection/ExcludeSection.tsx
@@ -11,6 +11,7 @@ import { ReactElement, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import * as Yup from 'yup';
 import { PaymentPlanStatusEnum } from '@restgenerated/models/PaymentPlanStatusEnum';
+import { PaymentPlanDetailBackgroundActionStatusEnum } from '@restgenerated/models/PaymentPlanDetailBackgroundActionStatusEnum';
 import { PERMISSIONS, hasPermissions } from '../../../../config/permissions';
 import { usePermissions } from '@hooks/usePermissions';
 import { useSnackbar } from '@hooks/useSnackBar';
@@ -91,6 +92,12 @@ function ExcludeSection({
     PERMISSIONS.PM_EXCLUDE_BENEFICIARIES_FROM_FOLLOW_UP_PP,
     permissions,
   );
+  // mirrors flows.py: an exclusion can only start when no background action is set,
+  // or when the previous exclusion failed.
+  const canRunExclusion =
+    !backgroundActionStatus ||
+    backgroundActionStatus ===
+      PaymentPlanDetailBackgroundActionStatusEnum.EXCLUDE_BENEFICIARIES_ERROR;
   const hasOpenOrLockedStatus =
     status === PaymentPlanStatusEnum.LOCKED ||
     status === PaymentPlanStatusEnum.OPEN;
@@ -101,6 +108,11 @@ function ExcludeSection({
     }
     if (!hasExcludePermission) {
       return t('Permission denied');
+    }
+    if (!canRunExclusion) {
+      return t(
+        'Another background action is currently running on this Payment Plan',
+      );
     }
     return '';
   };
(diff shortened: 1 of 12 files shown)


src/hope/apps/payment/celery_tasks.py
```diff
@@ -1173,7 +1174,7 @@ def payment_plan_exclude_beneficiaries_async_task_action(job: AsyncRetryJob) ->
         )
         if error_msg:
             return
-        raise
+        raise NonRetriableTaskError(str(exc)) from exc
```
johniak, line 1177: This will disable retries for unexpected errors. Is this change intended, It is not described inside PR summary

src/frontend/src/utils/constants.ts
```diff
@@ -313,3 +313,10 @@ export const generateTableOrderOptionsMember = (
     { name: 'Registration Date: descending', value: '-last_registration_date' },
   ];
 };
+
+export const PAYMENT_PLAN_BACKGROUND_ACTION_ERROR_STATUSES: string[] = [
```
johniak: `RULE_ENGINE_ERROR` and `SEND_TO_PAYMENT_GATEWAY_ERROR` is missing

This could be an enum Type PaymentPlanDetailBackgroundActionStatusEnum[]  instead of string[].

MarekBiczysko (dismissed):

src/frontend/src/components/paymentmodule/PaymentPlanDetails/ExcludeSection/ExcludeSection.tsx
```diff
@@ -102,6 +109,11 @@ function ExcludeSection({
     if (!hasExcludePermission) {
       return t('Permission denied');
     }
+    if (!canRunExclusion) {
```
johniak, line 113: This is also true when you are at terminal error status like `XLSX_EXPORT_ERROR` so it will show message which is confusing

johniak (changes requested):

src/hope/apps/payment/celery_tasks.py
pkujawa: updated the description with:
` - Unexpected failures are no longer retried. The job used to keep retrying for about 3 minutes, while the plan already showed the error status and Save was available again. A retry could then re-apply its own old list of IDs over whatever the user saved in the meantime. The task already treats a failure as final, it sets the error status and writes a message, so retrying underneath that was wrong. Transient failures no longer recover on their own, the user retries from the UI instead.`

johniak (approved):

MarekBiczysko (approved):

PR #6373, opened 2026-09-03
[305854] Fix unclear error and slow batches when finalizing Needs Adjudication tickets

pkujawa:
[AB#305854](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/305854)

## Error handling

If two individuals share a document number and you mark both as distinct, the request failed with a 500. The user saw unlclear error. This happens when closing a single ticket too, not just batches.

And when one ticket in a batch failed, the whole batch was cancelled without saying which ticket caused it.

- The document clash now returns a clear message instead of a 500:
  `Individual IND-26-0000.0157 cannot be marked as distinct: document 1354917433 conflicts with an existing valid document.`
- Every batch error now starts with `Ticket GRV-0000029: `.

## Speed

Finalizing more than one ticket took about 8 seconds locally, for any number of tickets. The tickets were loaded with 18 tables joined in one query, which made Postgres expect a huge result and switch on JIT compilation — 8 seconds to prepare a query returning a handful of rows.

Now the detail tables load in separate small queries. Same data, still no N+1.

Bulk close used the same query, so it is faster too.

### DIFF
--- a/src/hope/models/individual.py
+++ b/src/hope/models/individual.py
@@ -13,6 +13,7 @@ from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 from multiselectfield import MultiSelectField
 from phonenumber_field.modelfields import PhoneNumberField
+from rest_framework.exceptions import ValidationError
 
 from hope.apps.activity_log.utils import create_mapping_dict
 from hope.apps.core.languages import Languages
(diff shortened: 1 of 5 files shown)


src/hope/models/individual.py
```diff
@@ -510,8 +511,10 @@ def mark_as_distinct(self) -> None:
                 doc.save()
             # AB#244721
             except IntegrityError:
-                error_message = f"{self.unicef_id}: Valid Document already exists: {doc.document_number}."
-                raise Exception(error_message)
+                raise ValidationError(
+                    f"Individual {self.unicef_id} cannot be marked as distinct: document "
```
johniak, line 515: The same `IntegrityError` also comes from `unique_for_individual_if_not_removed_and_valid` (two documents of a `unique_for_individual` type on the same person).

johniak (changes requested):

MarekBiczysko (approved):

johniak (approved):

PR #6374, opened 2026-09-03
317775 nullify cw and originating ids on copy

Jeeziorny:
**Context**

https://github.com/unicef/hope/pull/6256#discussion_r3922261006

### DIFF
--- a/tests/unit/apps/program/test_program_utils.py
+++ b/tests/unit/apps/program/test_program_utils.py
@@ -71,6 +71,11 @@ def program2(afghanistan: BusinessArea) -> Program:
 
 
 @pytest.fixture
+def program3(afghanistan: BusinessArea) -> Program:
+    return ProgramFactory(name="Program 3", business_area=afghanistan)
+
+
+@pytest.fixture
 def household_original_already_enrolled(program1: Program) -> Household:
     return HouseholdFactory(
         program=program1,
@@ -621,9 +626,8 @@ def test_copy_program_population_clears_country_workspace_id_and_originating_id(
 
 @pytest.mark.usefixtures("mock_elasticsearch")
 def test_copy_program_population_twice_does_not_violate_unique_constraints(
-    afghanistan: BusinessArea, program1: Program, program2: Program
+    afghanistan: BusinessArea, program1: Program, program2: Program, program3: Program
 ) -> None:
-    program3 = ProgramFactory(name="Program 3", business_area=afghanistan)
     source_individual = IndividualFactory(
         household=None,
         business_area=afghanistan,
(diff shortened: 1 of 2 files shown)


tests/unit/apps/program/test_program_utils.py
```diff
@@ -577,3 +578,121 @@ def test_enroll_households_raises_on_integrity_error(enrollment_test_data: dict)
...
+        originating_id="API#IND-2",
+    )
+    source_household = HouseholdFactory(
+        business_area=afghanistan,
+        program=program1,
+        head_of_household=source_individual,
+        originating_id="API#HH-2",
+    )
+
+    for target_program in (program2, program3):
```
johniak: A `for` loop in the test body. Could you unroll this into two explicit calls (or parametrize) so a failure points at the first or the second copy, and move `program3` into a fixture?

tests/unit/apps/program/test_program_utils.py
```diff
@@ -577,3 +578,121 @@ def test_enroll_households_raises_on_integrity_error(enrollment_test_data: dict)
...
+    assert source_individual.originating_id == "API#IND-1"
+    assert source_household.originating_id == "API#HH-1"
+
+
+@pytest.mark.usefixtures("mock_elasticsearch")
+def test_copy_program_population_twice_does_not_violate_unique_constraints(
+    afghanistan: BusinessArea, program1: Program, program2: Program
+) -> None:
+    program3 = ProgramFactory(name="Program 3", business_area=afghanistan)
+    source_individual = IndividualFactory(
```
johniak: With both `country_workspace_id` and `originating_id` set on the source, this test fails on the base branch on `originating_id_ind_unique_constraint`, not on the new cw constraint (I ran it). Seeding only `country_workspace_id` here would make it actually guard the `(business_area, country_workspace_id)` case from #6256; otherwise the first test already covers both fields.

tests/unit/apps/program/test_program_utils.py
```diff
@@ -577,3 +578,121 @@ def test_enroll_households_raises_on_integrity_error(enrollment_test_data: dict)
                 enrollment_test_data["program2"],
                 str(enrollment_test_data["user"].pk),
             )
+
+
+@pytest.mark.usefixtures("mock_elasticsearch")
+def test_copy_program_population_clears_country_workspace_id_and_originating_id(
+    afghanistan: BusinessArea, program1: Program, program2: Program
+) -> None:
+    source_individual = IndividualFactory(
```
johniak: The same `source_individual` / `source_household` arrangement with cw and originating ids is built inline three times (here, L627, L658). One fixture would shorten the file and matches the fixtures-only convention for test data.

johniak (changes requested):

johniak (approved):

kosior (approved):

PR #6375, opened 2026-09-03
[AB#301335] Add Latin name fields for beneficiaries

johniak:

Continues #5913 (closed) without automatic transliteration. Decision from the Slack discussion (2026-09-02, Stefano): **do not perform any transliteration, leave the fields null, do not apply any processing to these fields** — character-by-character libraries give unusable results for Arabic and Hebrew names.

Beneficiary names can be written in a local script (e.g. Cyrillic). This PR adds `*_latin` fields that store the same names in the Latin alphabet, exactly as provided by the source. The existing name fields are not changed.

## Changes

- **Model**: new optional fields `given/middle/family/full_name_latin` on `Individual` (`ascii_name_validator`: Latin letters, spaces, hyphens, apostrophes), GDPR erase covers them
- **Import**: RDI XLSX/Kobo/People, the RDI push API, Aurora (generic + country services) and the universal update script accept `*_latin` values and store them as provided; missing → stays `NULL`. XLSX/Kobo validators reject non-Latin values with a row error
- **Grievance**: add-individual and individual-data-update tickets accept `*_latin` values, validated with the same rule on the backend and in the form
- **Search**: name search also matches the Latin fields — individual & household lists (DB and ES), office search (household, individual, grievance, payment, payment plan), grievance list, payment `collector_full_name`
- **API**: `*_latin` fields on individual / household member / grievance / needs-adjudication / payment list (snapshot collector) endpoints, core fields + XLSX template columns
- **UI**: Latin name shown in small font under the regular name wherever an individual name is rendered (details, lists, tables, lookups, RDI)

## Compared to #5913

Removed: `transliterate`/`unidecode`, `to_latin` / `set_names_latin`, auto-fill on import, the backfill script, the "transliterate automatically" checkbox and flag in tickets, latin recompute on name change, the `ES_USE_LATIN_NAMES` flag, and the fixture renames those needed.

## Rollout

Deploy only. Without a backfill no existing document carries `*_latin` values; the first record that does gets the field added to the live ES index by dynamic mapping, so search works as soon as data arrives. No reindex step — the next regular `es_reindex --all` picks up the declared mapping (phonetic analyzer / index prefixes).


src/hope/apps/household/migrations/0048_migration.py
```diff
@@ -0,0 +1,90 @@
+# Generated by Django 5.2.15 on 2026-07-01 10:42
```
Jeeziorny, line 1: How about existing individuals, do we want to have data migration to fill the latin variants? I can't see this in the ticket nor PR explicitly excluded or included

tests/e2e/new_selenium/grievance/test_latin_names_data_change.py
```diff
@@ -0,0 +1,73 @@
...
+pytestmark = pytest.mark.django_db()
+
+BUTTON_SUBMIT = 'button[data-cy="button-submit"]'
+INPUT_FULL_NAME_LATIN = 'input[data-cy="input-individualData.fullNameLatin"]'
+
+
+def _select_listbox(browser: HopeTestBrowser, field_name: str, option: str) -> None:
+    # choice options here carry codes (MALE, YES) in data-cy, so pick by visible text
+    browser.click(f'[data-cy="select-{field_name}"]')
+    browser.select_listbox_element(option)
```
Jeeziorny: Can we reuse select_dropdown_option of HopeTestBrowser here?

tests/unit/apps/aurora/test_sri_lanka_registration_service.py
```diff
@@ -221,6 +221,27 @@ def test_import_data_to_datahub(
...
+    registration: Any,
+    user: Any,
+    program: Program,
+    sri_lanka_country: Any,
+    national_id_document_type: Any,
+    sri_lanka_records: list[Record],
+) -> None:
+    service = SriLankaRegistrationService(registration)
+    rdi = service.create_rdi(user, f"sri_lanka rdi latin {datetime.datetime.now()}")
+    service.process_records(rdi.id, [record.id for record in sri_lanka_records])
```
Jeeziorny, line 235: Image

Is it `for` statement in test? Just joking, I think we can leave it!

Jeeziorny (dismissed):
LGTM

src/hope/apps/universal_update_script/universal_individual_update_service/all_updatable_fields.py
```diff
@@ -22,12 +22,17 @@
...
 
 individual_fields: dict[str, tuple[str, Any, Any]] = {
     "full_name": ("full_name", validate_string, handle_simple_field),
     "given_name": ("given_name", validate_string, handle_simple_field),
     "middle_name": ("middle_name", validate_string, handle_simple_field),
     "family_name": ("family_name", validate_string, handle_simple_field),
+    "full_name_latin": ("full_name_latin", ascii_name_validator, handle_simple_field),
+    "given_name_latin": ("given_name_latin", ascii_name_validator, handle_simple_field),
+    "middle_name_latin": ("middle_name_latin", ascii_name_validator, handle_simple_field),
+    "family_name_latin": ("family_name_latin", ascii_name_validator, handle_simple_field),
```
pkujawa: I think these 4 will crash the update on the first row.                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                             
The middle item in the tuple has to be a function taking five arguments that returns an error message, the way `validate_choices` does, because `universal_individual_update_service.py` calls it as:                                                                                      
                                                                                                                                                                                                                                                                                                                             
`error = validator(value, name, Individual, self.business_area, self.program)`
                                                                                                                                                                                                                                                                                                                             
`ascii_name_validator` is a Django RegexValidator - one argument, and it raises instead of returning. So here it will give `TypeError`.   
                                                                                                                                                                                                                                                                                                                             
A wrapper in validator_and_handlers.py fixes it - same shape as `validate_choices` right above it:                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                             
```
      def validate_latin_name(value, name, model_class, business_area, program) -> str | None:                                                                                                                                                                                                                               
          if value is None or value == "":                                                                                                                                                                                                                                                                                   
              return None                                                                                                                                                                                                                                                                                                    
          try:                                                                                                                                                                                                                                                                                                               
              ascii_name_validator(value)                                                                                                                                                                                                                                                                                    
          except ValidationError as e:                                                                                                                                                                                                                                                                                       
              return f"Invalid value {value} for column {name}: {e.message}"                                                                                                                                                                                                                                                 
          return None    
```                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
Could we add a test too that uses a latin column in universal update?

tests/unit/apps/core/test_kobo_template_upload.py
```diff
@@ -101,6 +101,10 @@ def test_upload_invalid_template_returns_expected_errors(
             "Field: relationship_i_c - Choice: OTHER is not present in the file",
             "Field: relationship_i_c - Choice: FOSTER_CHILD is not present in the file",
             "Field: relationship_i_c - Choice: FREE_UNION is not present in the file",
+            "Field: full_name_latin_i_c - Field is missing",
+            "Field: given_name_latin_i_c - Field is missing",
+            "Field: middle_name_latin_i_c - Field is missing",
+            "Field: family_name_latin_i_c - Field is missing",
```
pkujawa: So after this deploys nobody can upload their existing template until these four questions are added to it.
Just sth to keep in mind.

src/hope/contrib/aurora/services/ukraine_flex_registration_service.py
```diff
@@ -50,6 +50,9 @@ class UkraineBaseRegistrationService(BaseRegistrationService):
         "given_name": "given_name_i_c",
         "family_name": "family_name_i_c",
         "middle_name": "patronymic",
+        "given_name_latin": "given_name_latin_i_c",
```
pkujawa, line 53: Ukraine, Czech and Nigeria map only the part fields, never `full_name_latin` - but ukraine and czech already compose `full_name` from the parts and don't do the same for the latin ones.        
        
The lists and the search only ever use the full name - that's true for the standard names too, `IndividualListSerializer` and the search clauses don't touch `given_name`/`family_name` either. It works for `full_name` because it always gets composed; `full_name_latin` never does, so for these three services the latin name lands in the DB and then shows nowhere and matches nothing. It only appears on the detail page and in office search, which read all four.                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                             
Looking at the ticket, we were asked not to process it, so either these services should map `full_name_latin_i_c` too, or the list serializers and the search need to fall back to the part fields when `full_name_latin` is empty.

pkujawa (changes requested):

tests/unit/apps/aurora/test_sri_lanka_registration_service.py
johniak: :D

src/hope/apps/household/migrations/0048_migration.py
johniak: According to Stefano slack comment, we are not doing this

tests/unit/apps/core/test_kobo_template_upload.py
johniak: This is a real issue.

The problem is that this works the same way for every HOPE field used in Kobo. required: False does not decide whether the field has to be present in the form/JSON – it always has to be there. It only decides whether the field can come in empty.

I can change this behaviour, but keep in mind it will affect every non-required field. That is fine with me – I think it should have worked like this from the beginning.

@domdinicola @ntrncic

ntrncic: @srugano add these fields to Kobo template

johniak: On Daily there was decision made to allow non required fields to be absent in form/JSON

pkujawa (approved):

PR #6376, opened 2026-09-03
[333810] Move choices endpoints off business-area routes

kosior:
[AB#333810](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/333810)

Choices/enum endpoints were nested under `/api/rest/business-areas/{slug}/…` even though their payloads don't depend on the business area. They now live on flat routes under `/api/rest/choices/`, so the frontend fetches reference data once per session instead of once per business area.

**Two important changes:**

- Permissions: the moved endpoints drop their per-business-area permission gates and become IsAuthenticated - a business-area-free URL has nothing to check against. This matches every pre-existing `/api/rest/choices/*` endpoint.

- Financial institutions moved to their own business-area-scoped endpoint, so their list is now actually filtered by business area (the filter previously never ran, since individuals/choices never received the business area in serializer context).


src/hope/apps/account/api/views.py
```diff
@@ -195,6 +187,25 @@ def partner_for_grievance_choices(self, request: "Request", *args: Any, **kwargs
         return Response(to_choice_object(choices_data))
 
 
+class PartnerViewSet(BusinessAreaMixin, SerializerActionMixin, PermissionActionMixin, BaseViewSet):
+    """Serve the partner choices scoped to a business area."""
+
+    queryset = Partner.objects.all()
+    permission_classes_by_action = {
+        "choices": [IsAuthenticated],
```
johniak: Any authenticated user can now list the partners of any business area by changing the slug; there is no BA-membership check. `partner_for_grievance_choices` above serves the same partner set and still requires `USER_MANAGEMENT_VIEW_LIST` or the grievance permissions in that BA. Same for `DataCollectingTypeViewSet` and `FinancialInstitutionViewSet`. Is dropping the BA gate intended? If not, keeping `BaseRestPermission` with a widened `permissions_by_action["choices"]` would match the rest of the router.

src/frontend/src/components/grievances/EditIndividualDataChange/EditIndividualDataChange.tsx
```diff
@@ -80,31 +81,28 @@ function EditIndividualDataChange({
...
+    useDocumentTypeChoices();
+
+  const { data: individualChoicesData, isLoading: individualChoicesLoading } =
+    useQuery({
+      queryKey: restQueryKey(RestService.restChoicesIndividualsRetrieve),
+      queryFn: () => RestService.restChoicesIndividualsRetrieve(),
+      enabled: Boolean(businessArea),
+    });
+
+  const { data: financialInstitutionChoices } = useQuery({
```
johniak: This query isn't part of the loading gate below, unlike the other three here and unlike `RequestedIndividualDataChangeTable` in this PR, which does wait for `financialInstitutionChoicesLoading`. `EditAccountRow` gets `financialInstitutionChoices || []`, so an existing account's select briefly holds a value with no matching option. Same in `EditPeopleDataChange` and `IndividualAccounts` (shows the raw id until loaded). Worth picking one pattern.

tests/unit/apps/account/test_views_partner_choices.py
```diff
@@ -0,0 +1,96 @@
...
+    return reverse("api:accounts:partners-choices", kwargs={"business_area_slug": business_area_slug})
+
+
+def test_get_choices_returns_the_partners_of_the_business_area(
+    authenticated_client: Any,
+    afghanistan: BusinessArea,
+    partner: Partner,
+) -> None:
+    partner.allowed_business_areas.add(afghanistan)
+    unicef_hq = PartnerFactory(name="UNICEF HQ")
```
johniak: Factories called directly in the test body; the old test had a `unicef_partners` fixture for exactly this. Same pattern in `test_grievance_choices.py` and `test_household_global_views.py` (`api_client(UserFactory(partner=PartnerFactory(...)))`).

src/hope/apps/core/management/commands/generate_openapi.py
```diff
@@ -50,6 +50,10 @@ def _collect_choices(self) -> dict[str, Any]:
         for action in ChoicesViewSet.get_extra_actions():
             if action.detail or "get" not in action.mapping:
                 continue
+            if not action.kwargs.get("enum_source", True):
+                # Either a bundle (object-shaped, not a flat choice list) or a
+                # database-backed list that must not be frozen into the generated code.
```
johniak: `countries` and `currencies` are DB-backed too and still get frozen, so "must not be frozen" isn't a rule the code follows. Better to say what the flag does.

src/hope/apps/core/api/views.py
```diff
@@ -239,3 +296,71 @@ def countries(self, request: Request) -> Response:
...
+    def survey_categories(self, request: Request) -> Response:
+        """Return the categories a survey can belong to."""
+        resp = ChoiceSerializer(to_choice_object(Survey.CATEGORY_CHOICES), many=True).data
+        return Response(resp)
+
+    @extend_schema(responses={200: ChoiceSerializer(many=True)})
+    @action(detail=False, methods=["get"], url_path="document-types", enum_source=False)
+    def document_types(self, request: Request) -> Response:
+        """Return the document types a person can hold."""
+        from hope.models import DocumentType
```
johniak: `hope.models` is already imported at the top of this module; `DocumentType` can go there with the others.

johniak (changes requested):

src/frontend/src/containers/forms/ProgramForm.tsx
```diff
@@ -36,11 +36,19 @@ const ProgramForm = ({
...
-    queryKey: restQueryKey(RestService.restBusinessAreasProgramsChoicesRetrieve, {
-      businessAreaSlug: businessArea,
-    }),
+    queryKey: restQueryKey(RestService.restChoicesProgramsRetrieve),
+    queryFn: () => RestService.restChoicesProgramsRetrieve(),
+    staleTime: 1000 * 60 * 10,
+    gcTime: 1000 * 60 * 30,
+  });
+
+  const { data: dataCollectingTypeChoices } = useQuery({
```
johniak, line 45: Nothing gates on this query. The parent pages wait on `choicesLoading` and `userPartnerChoicesLoading` only, and `if (!data || !beneficiaryGroupsData) return null` below doesn't cover it. On Edit and Duplicate Program the required Data Collecting Type select mounts with `choices={[]}` while `values.dataCollectingTypeCode` is already set, so MUI logs an out-of-range value and the field renders empty until the fetch lands.

johniak (dismissed):

johniak (dismissed):

pkujawa (dismissed):

domdinicola:
@kosior pls resolve conflicts

johniak (approved):

PR #6378, opened 2026-09-04
TS enhancements

mmaciekk:
Modernizes the frontend `tsconfig.json` for TS 5.9 / Vite 8, turns on the strict options the codebase can already support, and fixes what they flagged.

`strict` and `strictNullChecks` stay off. No behaviour changes except the bugs listed below.

**One commit per compiler option — review commit by commit, not the combined diff.**

| Commit | What it does |
|---|---|
| Modernize tsconfig | `module: preserve`, ES2022, `jsx: react-jsx`, drop dead options and the unused `tsconfig.node.json` |
| `noUnusedLocals` / `noUnusedParameters` | 52 unused imports and params removed |
| `useUnknownInCatchVariables` | 9 `catch` blocks narrowed via 3 new helpers in `utils/errors.ts` |
| `strictFunctionTypes` | 31 callbacks declaring narrower params than their caller passes |
| `verbatimModuleSyntax` | `import` → `import type` in 741 files. Mechanical, done with `eslint --fix`. **This is ~95% of the diff.** |
| Guard null dereferences (x2) | 307 places that could throw at runtime. Now 0. |

Bugs fixed along the way:

- Autocomplete option matching compared `.id`/`.code` against raw input text, so it never matched before an option was picked.
- `approvalNumberRequired - 1 === count` was `NaN` when the number was absent, so approval counts never matched.
- `EditGrievancePage` and others read query data behind only an `isLoading` check — an errored or disabled query settles with `data` undefined and crashed the page.
- `SplitIntoPaymentLists` showed `[object Object]` instead of the error message.
- `NewOfflineTemplatePage` posted `values.name` without seeding it in `initialValues`.

Checks at the final commit: `tsc --noEmit` error set identical to before the series, lint clean, tests 84 passed / 7 pre-existing snapshot failures, `vite build` green.

Not in this PR: turning on `strictNullChecks` itself. The crash sites are cleared, but the remaining reports still need work first.


src/frontend/src/components/paymentmodule/PaymentPlanDetails/PaymentPlanDetailsHeader/SplitIntoPaymentLists.tsx
```diff
@@ -116,7 +122,7 @@ export const SplitIntoPaymentLists = ({
       setDialogOpen(false);
       showMessage(t('Split was successful!'));
     } catch (e) {
-      showMessage(e);
+      showMessage(getErrorMessage(e));
```
johniak: This still hides the reason. `RestService` throws `ApiError`, whose `message` is the status text, so the user now sees "Bad Request" instead of the backend's "Payment plan must be accepted to make a split". The other dialogs in this folder use `showApiErrorMessages(e, showMessage, t('...'))`, which reads `body`.

src/frontend/src/containers/pages/grievances/EditGrievancePage.tsx
```diff
@@ -262,6 +262,11 @@ const EditGrievancePage = (): ReactElement => {
     return <LoadingComponent />;
   if (isPermissionDeniedError(error))
     return <PermissionDenied permission={PERMISSIONS.GRIEVANCES_UPDATE} />;
+  // The loading flags going false does not guarantee data: a query that errored
+  // or is disabled settles with `data` undefined, and everything below
+  // dereferences these three unconditionally.
+  if (!ticketData || !currentUserData || !choicesData)
```
johniak: After the loading flags settle, a missing `ticketData` means the query errored (404, 500), and this renders a spinner forever. `GrievancesDetailsPage` returns `null` at the same point, and `error` is already destructured here. Same pattern in `Messages.tsx`, `Notes.tsx` and `LookUpLinkedTicketsTable.tsx`.

johniak (changes requested):

src/frontend/src/components/paymentmodule/PaymentPlanDetails/PaymentPlanDetailsHeader/SplitIntoPaymentLists.tsx
mmaciekk: Good catch — switched to `showApiErrorMessages(toApiError(e), showMessage, t('Failed to split the Payment Plan'))`, matching `DeletePaymentPlan` and the other dialogs in this folder, so the DRF body reaches the user instead of the status text. (3a5c69a → pushed as part of 11d6902.)

src/frontend/src/containers/pages/grievances/EditGrievancePage.tsx
mmaciekk: Right — returning `null` now instead of `<LoadingComponent />`, same as `GrievancesDetailsPage` at the equivalent point. Checked the other three you named: `Messages.tsx`, `Notes.tsx` and `LookUpLinkedTicketsTable.tsx` already `return null` on their `|| !data` guards, and a sweep of the diff found no other `return <LoadingComponent />` added behind a settled-data check — this was the only site.

johniak (approved):

MarekBiczysko (approved):

PR #6382, opened 2026-09-07
AB#340074 Harden celery task locks and add admin lock removal

johniak:
[AB#340074](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/340074)

Celery workers die on OOM and deploys and leave locks behind for hours (2h, some 24h), so retried jobs fail or silently succeed without doing the work.

- `celery_lock(task, *parts)` in `hope/apps/core/celery_lock.py` replaces every `cache.lock` / `locked_cache` in celery tasks: python-redis-lock with auto-renewal (5 min TTL, renewed every ~200 s), so a dead worker frees its lock within minutes.
- Key convention `celery_lock_<task>:<params>`, applied to all 19 lock sites, including the two hand-rolled `cache.get/set` guards in `prepare_payment_plan` (10h) and `enroll_households_to_program` (24h).
- 12 tasks that mutate the same rows or an external system but had no lock (Payment Gateway send, exclude beneficiaries, Steficon rule/selection, flat amount, custom exchange rate, the five XLSX imports, copy_program) now take one too.
- Held lock ⇒ wait up to 6 min, then `AlreadyRunningError` (non-retriable): the job fails with the lock name in `job.errors` instead of reporting success. Global / per-programme locks (PDU merge, document dedup, token numbers) wait 30 min instead, since a held lock there means queued work.
- Admin: `admin/celery-locks/` page grouped per task, plus "Remove task locks" button on RDI, PaymentPlan, PaymentPlanGroup, Program, WU report, PDUOnlineEdit, UniversalUpdate and aurora Registration. Root only, requires typing `I confirm`, every removal logged in `LogEntry`.
- Tests keep `LocMemCache` (per process, xdist-safe) with a lock class mirroring the redis-lock API.

`Celery Locks,celery-locks/` was added to the default `QUICK_LINKS`; existing environments keep their stored constance value and need the link added by hand.

## Locked tasks (31)

| Area | Tasks |
|------|-------|
| Payment plan | prepare, full rebuild, rebuild stats, exclude beneficiaries, Steficon rule, Steficon HH selection, flat amount, custom exchange rate, send to Payment Gateway, reconciliation overdue e-mail |
| Payment plan XLSX | payment list import, delivery import, FSP extra fields import, follow-up instruction reconciliation import, group delivery import, group delivery export (+ nested token/order numbers per programme) |
| Western Union | report e-mail notifications |
| RDI | XLSX import, Kobo import, programme population import, merge, document deduplication (global), CW arrival hook (classify findings), generic import |
| Other | PDU online edit merge (global), universal update run + template, enroll households to programme, copy programme, Aurora automatic RDI creation |

Everything else (periodic syncs and cleanups, e-mails, file exports, idempotent recalculations, `@app.task` queue wrappers) stays unlocked on purpose: a duplicate run is harmless there and a lock would only add a new failure mode.

Docs: `guide-dev/celery-locks.md` (lock contract, key inventory, testing) and the *Celery locks* section in `guide-adm/maintenance.md`.


johniak:
Nikola Mentioned on 10.09 We leave it in draft for now

PR #6383, opened 2026-09-07
docs: document Needs Adjudication management

mmaciekk:
Adds the user-guide page for Needs Adjudication management. The documentation was written back in August alongside the feature (PR #6235) but the commit never left a local branch, so it was lost when `needs-adjudication-management` and `long-term/grievances-revamp` were merged and deleted — the feature shipped, the docs did not.

## Changes

- New `docs/guide-user/needs-adjudication-management.md` (227 lines) plus 7 screenshots.
- Grievance guides are now grouped into a single **Grievance** nav section rather than sibling top-level entries:

  ```
  - Grievance:
    - Overview: grievance.md
    - Biometric Photo Error Tickets: biometric-photo-error.md
    - Needs Adjudication Management: needs-adjudication-management.md
  ```

- This also registers `biometric-photo-error.md` in `.pages`. That file has been in the repo for a while but was never added to the nav, so it did not render in the User Manual at all.

## Note

The Needs Adjudication feature itself is on `develop`, not yet on `master`, so this page documents functionality that master has not shipped. Retarget to `develop` if the docs should land with the code instead.


docs/guide-user/needs-adjudication-management.md
```diff
@@ -0,0 +1,227 @@
...
+
+| Permission | Grants |
+|---|---|
+| `GRIEVANCES_APPROVE_FLAG_AND_DEDUPE` | Access to the workspace and the ability to finalize |
+| `GRIEVANCES_APPROVE_FLAG_AND_DEDUPE_AS_CREATOR` | The same, restricted to tickets the user created |
+| `GRIEVANCES_APPROVE_FLAG_AND_DEDUPE_AS_OWNER` | The same, restricted to tickets the user owns |
+| `GRIEVANCES_VIEW_BIOMETRIC_RESULTS` | The deduplication engine's similarity score on biometric tickets |
+| `POPULATION_VIEW_INDIVIDUAL_DELIVERY_MECHANISMS_SECTION` | The **Account** row in the comparison table |
+
+Any one of the first three is enough to see the **NA Tickets Management** button and open the screen.
```
pkujawa: This is not true - for NA management we do not check `AS_CREATOR` or `AS_OWNER` permissions - noone can be owner since we skip the assignment step and noone can be creator as these tickets are system generated.
Permissions checked are and `GRIEVANCES_APPROVE_FLAG_AND_DEDUPE` and `GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK` - both are needed to access the new section.

(+ `GRIEVANCES_VIEW_BIOMETRIC_RESULTS` and `POPULATION_VIEW_INDIVIDUAL_DELIVERY_MECHANISMS_SECTION` for the additional info listed in the table)

docs/guide-user/needs-adjudication-management.md
```diff
@@ -0,0 +1,227 @@
...
+Clicking it raises a confirmation:
+
+> You are about to finalize *N* ticket(s). They will be closed and this cannot be undone.
+
+![The Finalize confirmation dialog, with Cancel and Continue](./_screenshots/needs-adjudication/NAFinalize.png)
+
+**CONTINUE** commits the batch, **CANCEL** returns to the workspace with every decision intact. On success a snackbar confirms *N ticket(s) finalized*, the session decisions are cleared, and the counter returns to zero.
+
+!!! warning "Finalize is all or nothing"
+    The whole batch is committed in a single transaction. If any ticket in it fails validation — it has been closed in the meantime, it is no longer a Needs Adjudication ticket, an individual appears in both the duplicate and the distinct list — **nothing at all is written**, and every decision stays on screen so it can be corrected and resubmitted.
```
pkujawa: `Closed in the meantime `is the one case that doesn't fail the batch - those tickets are skipped, the rest still commits, and the message names them (`N ticket(s) finalized. Closed by someone else in the meantime and skipped: ...`).

And `it is no longer a Needs Adjudication ticket` isn't some possible situation.

Maybe we can split:

```
"Tickets someone else closed first are skipped"
If a ticket in the batch was closed by another user between the decision and **Finalize**, it is skipped and the rest of the batch still goes through. The confirmation message names the skipped tickets.

!!! warning "Errors stop the whole batch"
Apart from that, the batch goes through as one — all of it or none of it. If any ticket cannot be closed — an individual on it has been withdrawn somewhere else in the meantime, the same individual is withdrawn on one ticket in the batch and kept as distinct on another - nothing is saved, and every decision stays on screen so it can be corrected and resubmitted.
```

docs/guide-user/needs-adjudication-management.md
```diff
@@ -0,0 +1,227 @@
...
+
+```
+/{business area}/programs/{programme code}/grievance/na-tickets-management
+```
+
+The page header holds a **Grievance Tickets List** breadcrumb and a back arrow, both of which return to the ticket list.
+
+!!! info "Work inside a programme, not All Programmes"
+    The workspace is available in the All Programmes scope, where it adds a **Programme** filter. However, the role-reassignment lookup resolves candidate individuals against the *selected* programme, so adjudicating from within a specific programme is the reliable way to work.
+
```
pkujawa: This isn't right, the feature works for both scenarios - program-scoped and All Programs.

The reassignment lookup isn't scoped to the selected programme — `NaReassignRoleModal` passes the programme of the `individual` on the ticket, and the list, count, detail and finalize calls all have explicit All Programmes branches.

Let's add info that the feature works in both views.

docs/guide-user/needs-adjudication-management.md
```diff
@@ -0,0 +1,227 @@
...
+![The Reassign Role dialog: individual filters, the lookup table, and Save disabled until an individual is picked and Identity Verified is ticked](./_screenshots/needs-adjudication/NAreassignRoleIndTable.png)
+
+!!! info "The dialog does not save to the server"
+    Unlike the reassignment dialog on the single-ticket flow, this one only records the choice in the session alongside the rest of the decision. The handover is performed as part of **Finalize**.
+
+---
+
+## Finalizing
+
+**Finalize** is disabled while there is nothing to finalize, while any managed ticket still needs a reassignment, and while a finalize is already in flight. In the second case, hovering over the button explains why — *N ticket(s) need a role reassignment before finalizing*.
```
pkujawa: Let's add info about ticket number cap: it is disabled also while more than 50 tickets are managed

docs/guide-user/needs-adjudication-management.md
```diff
@@ -0,0 +1,227 @@
...
+
+Each row identifies a ticket by **Ticket ID**, its issue type, its creation date, and a coloured urgency badge. Selecting a row loads it into the comparison panel; the first ticket on the page is selected automatically.
+
+Once a decision has been recorded, the row also carries a status label in its bottom-right corner:
+
+| Label | Meaning |
+|---|---|
+| ***Ticket managed*** | A complete decision has been recorded and this ticket is ready to be finalized |
+| ***Reassignment required*** | A decision exists, but a withdrawn individual still holds a role that has to be handed over first — shown in amber with a warning icon |
+
```
pkujawa, line 86: There's a third label — `Decision incomplete` when a multi-duplicate ticket has some pairs undecided, and it takes precedence over "Reassignment required". 

`| ***Decision incomplete*** | The ticket has more than one possible duplicate and at least one pair has not been decided yet — flagged with a warning icon |`

docs/guide-user/needs-adjudication-management.md
```diff
@@ -0,0 +1,227 @@
...
+The header shows a running counter, **Tickets managed: N**, and the **Finalize** button. The list is always restricted to the Needs Adjudication category and ordered newest first.
+
+![The workspace with the ticket list on the left and the comparison panel on the right](./_screenshots/needs-adjudication/NATicketsManagementPage.png)
+
+---
+
+## Filtering the queue
+
+Open the filters with **SHOW FILTERS**; **HIDE FILTERS** collapses them again. **APPLY** commits the selection and returns the list to page 1, **CLEAR** resets everything. Applied filters are written into the page URL, so a narrowed queue can be bookmarked or shared with a colleague.
```
pkujawa: `APPLY` and `CLEAR` also pop a confirmation when you have decisions waiting: filters only change what you see, but Finalize still closes everything you decided on, including tickets the filter hid.

Maybe worth including

docs/guide-user/needs-adjudication-management.md
```diff
@@ -0,0 +1,227 @@
...
+
+## The workspace
+
+The screen is laid out in three parts:
+
+- a **filters bar** across the top, collapsed by default;
+- the **ticket list** down the left-hand side, roughly 40% of the width;
+- the **comparison panel** filling the rest.
+
+The header shows a running counter, **Tickets managed: N**, and the **Finalize** button. The list is always restricted to the Needs Adjudication category and ordered newest first.
```
pkujawa: The list is also restricted to `active` tickets - worth adding here

pkujawa (changes requested):

docs/guide-user/needs-adjudication-management.md
mmaciekk: Right — fixed in 9a21ac2267. The Permissions section now says both `GRIEVANCES_APPROVE_FLAG_AND_DEDUPE` and `GRIEVANCES_CLOSE_TICKET_EXCLUDING_FEEDBACK` are required (`canManageNeedsAdjudication` uses `.every`, and `bulk_resolve_needs_adjudication` re-checks both per ticket), the AS_CREATOR/AS_OWNER rows are gone, and the two additional-info permissions are split into their own table.

docs/guide-user/needs-adjudication-management.md
mmaciekk: Fixed in 9a21ac2267 — used your split almost verbatim: an info block for the skipped-closed case quoting the real message, and a separate "Errors stop the whole batch" warning. Dropped "no longer a Needs Adjudication ticket".

docs/guide-user/needs-adjudication-management.md
mmaciekk: Fixed in 9a21ac2267 — the admonition is now "Works in both views" and says the list, count, comparison panel and finalize all have All Programmes handling, and that the reassignment lookup resolves against the programme of the individual on the ticket.

docs/guide-user/needs-adjudication-management.md
mmaciekk: Fixed in 9a21ac2267 — the Finalizing section now lists all the disabled conditions with a tooltip table, including the 50-ticket cap (also enforced server-side by `MAX_NEEDS_ADJUDICATION_BATCH`) and the undecided-pair case.

docs/guide-user/needs-adjudication-management.md
mmaciekk: Fixed in 9a21ac2267 — added the `Decision incomplete` row with your wording, plus a line noting it takes precedence over "Reassignment required" (`NaTicketListItem.tsx` checks `incomplete` first).

docs/guide-user/needs-adjudication-management.md
mmaciekk: Good catch — added in 9a21ac2267, with the "Change filters" confirmation text quoted and a note that a filter only changes what is visible while Finalize still closes hidden decided tickets.

docs/guide-user/needs-adjudication-management.md
mmaciekk: Added in 9a21ac2267 — the list is now described as restricted to *active* Needs Adjudication tickets (`grievanceStatus: GrievanceStatuses.Active`).

pkujawa (approved):

PR #6393, opened 2026-09-10
AB#340074 Add clear_celery_locks management command for deploys

johniak:
[AB#340074](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/340074)

Celery workers replaced on deploy leave their task locks in Redis for hours, so redelivered jobs fail or skip their work. This adds `python manage.py clear_celery_locks`, to be run once per deploy after the old worker pods are stopped and before the new ones start.

- Command holds an explicit list of lock key prefixes, one per lock site (`cache.lock`, `cache.add`, `locked_cache`, hand-rolled `cache.get`/`cache.set` guards), and deletes only keys starting with one of them. The rest of the cache is untouched.
- The `prepare_payment_plan` and `enroll_households_to_program` guards used a bare SHA256 of the params as the key, so nothing could tell them apart from other cache entries. Their keys now start with the task name; the hashing itself is unchanged.
- Test `LocMemCache.delete_pattern` re-prefixed the key on delete and never removed anything; fixed so the command can be tested.
- Docs: "Celery locks" section in `guide-adm/maintenance.md` with the deployment order.

### DIFF
--- a/src/hope/apps/core/management/commands/clear_celery_locks.py
+++ b/src/hope/apps/core/management/commands/clear_celery_locks.py
@@ -1,0 +1,42 @@
+from typing import Any
+
+from django.core.cache import cache
+from django.core.management import BaseCommand
+
+# One prefix per lock site in celery tasks, whatever the mechanism (cache.lock, cache.add, locked_cache, get/set guard).
+# When you add a lock to a task, add its key prefix here.
+LOCK_KEY_PREFIXES = (
+    # cache.lock
+    "export_payment_plan_group_delivery_xlsx_",
+    "payment_plan_generate_token_and_order_numbers_",
+    "payment_plan_rebuild_stats_",
+    "payment_plan_full_rebuild_",
+    "send_western_union_report_email_notifications_",
+    "send_payment_plan_reconciliation_overdue_email_",
+    "pdu_online_edit_merge",
+    "lock:run_universal_individual_update_async_task:",
+    "lock:generate_universal_individual_update_template_async_task:",
+    # cache.add
+    "dash_report_task_running_",
+    # locked_cache, which is cache.get_or_set
+    "registration_xlsx_import_async_task-",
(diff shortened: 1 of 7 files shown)


src/hope/apps/core/management/commands/clear_celery_locks.py
```diff
@@ -0,0 +1,41 @@
...
+    "lock:run_universal_individual_update_async_task:",
+    "lock:generate_universal_individual_update_template_async_task:",
+    # cache.add
+    "dash_report_task_running_",
+    # locked_cache, which is cache.get_or_set
+    "registration_xlsx_import_async_task-",
+    "registration_program_population_import_async_task-",
+    "registration_kobo_import_async_task-",
+    "merge_registration_data_import_async_task-",
+    "classify_findings_and_schedule_merge_async_task-",
```
MarekBiczysko: this one is stale (classify_findings_and_schedule_merge_async_task-), use: fetch_findings_and_merge_rdi-program-

MarekBiczysko (changes requested):

src/hope/apps/core/management/commands/clear_celery_locks.py
```diff
@@ -0,0 +1,41 @@
+from typing import Any
+
+from django.core.cache import cache
+from django.core.management import BaseCommand
+
+# One prefix per lock site in celery tasks, whatever the mechanism (cache.lock, cache.add, locked_cache, get/set guard).
+# When you add a lock to a task, add its key prefix here.
+LOCK_KEY_PREFIXES = (
```
Jeeziorny: I feel like those needs to be placed in some constant values in celery tasks and reimported here. Without it we have two sources of truth and `LOCK_KEY_PREFIXES` can get stale after some time

PR #6396, opened 2026-09-11
[336238] GRV notification and My Tasks

pkujawa:
[AB#336238](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/336238)
Replaces the per-ticket grievance emails with four daily emails, and adds a My Tasks page for them to link into.

Before, an operator got an email on every event — ticket created, assigned, changed, plus one reminder per overdue ticket. 1 email per ticket.

## Emails

Each recipient now gets at most four a day, and only when there is something to report:

- **Tickets Needing Assignment** — to people who can assign, for tickets with no assignee
- **Tickets Assigned to You** — tickets assigned to you that day
- **Overdue Tickets** — your tickets past their due date
- **Updated Tickets** — your tickets changed that day by someone else

Every email shows a count for Sensitive and a count for Other, each linking to the matching list. No email names a ticket, so a sensitive ticket's ID never reaches an inbox.

Sent once a day at 06:00 in the recipient's own timezone.

## My Tasks page

A new page with two tabs — Needs Assignment and Assigned To Me — plus a sensitivity filter (All / Sensitive / Other) and an overdue filter. The emails link into it with the tab and filters already applied.

## Settings

Three new Constance entries: the overdue thresholds for sensitive (1 day) and other (30 days) tickets, and the hour the emails go out (06:00).

## Docs

Two new user guide pages: My Tasks, and Grievance Daily Emails.


src/hope/apps/grievance/models.py
```diff
@@ -557,6 +557,9 @@ class Meta:
                 name="idx_gt_ba_updated_not_ign",
             ),
             models.Index(fields=["assigned_at"], name="idx_gt_assigned_at"),
+            models.Index(fields=["assigned_to", "status"], name="idx_gt_assigned_to_status"),
+            models.Index(fields=["assigned_to", "category", "status"], name="idx_gt_assigned_cat_status"),
+            models.Index(fields=["assigned_to", "created_at", "status"], name="idx_gt_assigned_created_status"),
```
nerjan: How about using indexes bound with business_area, as it is almost always as it is done, not indexing status but rather filtering out closed statuses as it is used only this way
```
            models.Index(
                fields=["business_area", "category", "created_at"],
                condition=~models.Q(status=6),  # STATUS_CLOSED
                name="idx_gt_ba_cat_created_open",
            ),
            models.Index(
                fields=["assigned_to", "business_area", "category"],
                condition=~models.Q(status=6),  # STATUS_CLOSED
                name="idx_gt_assigned_ba_cat_open",
            ),
            # _updated_tickets(): one business area, a one-day user_modified window.
            models.Index(fields=["business_area", "user_modified"], name="idx_gt_ba_user_modified"),
            # these 2 are for hourly tasks
            models.Index(fields=["assigned_to", "business_area"], name="idx_gt_assigned_to_ba"),
            models.Index(fields=["created_by", "business_area"], name="idx_gt_created_by_ba"),
```

src/hope/apps/grievance/services/daily_digest_service.py
```diff
@@ -92,103 +172,314 @@ def __init__(
...
+
+        users = User.objects.in_bulk(set(sensitive_totals) | set(other_totals))
+        return (
+            {user: sensitive_totals[user_id] for user_id, user in users.items() if sensitive_totals[user_id]},
+            {user: other_totals[user_id] for user_id, user in users.items() if other_totals[user_id]},
+        )
+
+    def _needs_assignment_counts(self) -> tuple[dict[User, int], dict[User, int]]:
+        """Count unassigned tickets per user who can assign them, sensitive and other separately."""
+        category_by_ticket_id = dict(self._unassigned_tickets().values_list("id", "category"))
```
nerjan: This dict might be really big so it might be better use queries here instead of python objects
```
self._unassigned_tickets().aggregate(
    sensitive=Count("id", filter=Q(category=SENSITIVE) & scope_q(sensitive_programs)),
    other=Count("id", filter=~Q(category=SENSITIVE) & scope_q(other_programs)),
)
```
with `scope_q(ids) = Exists(through.objects.filter(grievanceticket=OuterRef("pk"), program_id__in=ids)) | without_program_q()`

src/hope/apps/grievance/services/daily_digest_service.py
```diff
@@ -92,103 +172,314 @@ def __init__(
...
-                    f"Failed to send the {self.digest_date.isoformat()} grievance digest to user {digest.user.pk}"
-                )
-                failed += 1
+        for user, specs in self._emails_by_recipient().items():
+            for spec, sections in specs:
+                email_key = self.delivery_key(user.pk, spec)
+                if email_key in skip_email_keys:
+                    continue
+                try:
+                    self._build_email(spec, user, sections).send_email()
```
nerjan, line 215: the per-email retry machinery cannot observe a real send failure as it only queues celery task.

nerjan (changes requested):

src/frontend/src/containers/pages/grievances/MyTasksPage.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  // The tab's own axis is derived from the tab, never from the filter bar, so it stays out of the
+  // URL. Sensitivity is the exception: it is a filter, so it travels in appliedFilter, and only a
+  // permission-pinned value is forced here.
+  const extraQueryParams = useMemo(
+    () => ({
+      ...(activeTab === 'needs-assignment'
+        ? // Every unassigned closed ticket is unassigned for good, so the tab that exists to get
+          // tickets assigned never shows them - matching the count the email was built from.
+          { unassigned: true, grievanceStatus: GrievanceStatuses.Active }
```
MarekBiczysko, line 177: Shouldn't we show only the tickets actually "assignable" by the user?
In the All Programs view:
This filter returns all unassigned tickets the user can view, including tickets they cannot assign. For example, if a user can view programmes A and B but can assign tickets only in programme A, they will still see tickets from programme B.

MarekBiczysko: This requires a backend change, I put this comment here just to mark the entry point. (daily emails implements this properly using users_with_permissions_by_program)

src/frontend/src/containers/pages/grievances/MyTasksPage.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+// The tab values travel in the URL as ?tab= and are what the daily digest emails deep-link to;
+// they must stay in step with hope.apps.grievance.constants (PRESET_*). Sensitivity is not a tab:
+// every email counts sensitive and other tickets separately and links each count with ?sensitive=,
+// which lands on the filter below.
+export const MY_TASKS_TABS = [
+  {
+    value: 'needs-assignment',
+    label: 'NEEDS ASSIGNMENT',
+    permissions: [PERMISSIONS.GRIEVANCE_ASSIGN],
```
MarekBiczysko: GRIEVANCE_ASSIGN alone is not enough: loading the tickets requires a list-view permission, while bulk assignment requires GRIEVANCES_UPDATE. For example, a user with only GRIEVANCE_ASSIGN can open this tab but gets a permission error.

GRIEVANCES_VIEW_LIST_PERMISSIONS + PERMISSIONS.GRIEVANCE_ASSIGN + PERMISSIONS.GRIEVANCES_UPDATE ?

The same should be probably applied to bulk assignment button.

Backend needs to be changed accordingly

MarekBiczysko (changes requested):

src/hope/apps/grievance/services/daily_digest_service.py
MarekBiczysko: This loads every unassigned ticket into Python and then copies ticket ID sets for each eligible user:

```
  category_by_ticket_id = dict(
      self._unassigned_tickets().values_list("id", "category")
  )

  visible[user].update(program_ticket_ids - sensitive_ids)
```

  For example, 50,000 unassigned tickets and 200 users with business-area-wide access could create millions of in-memory set entries. This work is also repeated for each recipient timezone.

  Consider calculating the totals in the database using the intersection of programmes where the user can assign and view tickets:

```
  sensitive_programs = assigned_programs & sensitive_view_programs
  other_programs = assigned_programs & other_view_programs

  counts = self._unassigned_tickets().aggregate(
      sensitive=Count(
          "id",
          filter=Q(category=SENSITIVE)
          & scope_q(sensitive_programs),
      ),
      other=Count(
          "id",
          filter=~Q(category=SENSITIVE)
          & scope_q(other_programs),
      ),
  )

```
  The scope could use an Exists subquery so tickets linked to multiple permitted programmes are counted only once:

```
  def scope_q(program_ids: set) -> Q:
      through = GrievanceTicket.programs.through
      return Q(
          Exists(
              through.objects.filter(
                  grievanceticket=OuterRef("pk"),
                  program_id__in=program_ids,
              )
          )
      )
```

  Users with identical programme scopes should share the same aggregate query to avoid replacing the memory issue with one query per user. Programme-less tickets also need to be included conditionally, only when
  the user has both the relevant view and assignment permissions.

src/hope/apps/grievance/services/daily_digest_service.py
```diff
@@ -92,103 +172,314 @@ def __init__(
...
+        `last_notification_sent` keeps the existing repeat spacing, so a stale backlog does not mail
+        every morning.
+        """
+        tickets = self._overdue_tickets(sensitive=sensitive)
+        due_assignee_ids = set(
+            tickets.filter(self._reminder_due_q(sensitive=sensitive)).values_list("assigned_to_id", flat=True)
+        )
+        return {user: total for user, total in self._counts_by_assignee(tickets).items() if user.pk in due_assignee_ids}
+
+    def _reminder_due_q(self, *, sensitive: bool) -> Q:
```
MarekBiczysko, line 462: this should also get notification_time as an argument

src/hope/apps/grievance/services/daily_digest_service.py
```diff
@@ -92,103 +172,314 @@ def __init__(
...
+            .filter(
                 self._recipient_timezone_filter("assigned_to"),
-                Q(assigned_by__isnull=True) | ~Q(assigned_by=F("assigned_to")),
-                Q(assigned_to__is_active=True, assigned_to__email__gt=""),
-                business_area=self.business_area,
-                business_area__enable_email_notification=True,
-                assigned_at__gte=self.start,
-                assigned_at__lt=self.end,
-                assigned_to__isnull=False,
+                overdue_q(),
```
MarekBiczysko: overdue_q(now=self.notification_time)

src/hope/apps/grievance/celery_tasks.py
MarekBiczysko: Not a major issue, but something to consider (please check also other review comments with self.notification_time)

The job stores which day it represents, but overdue tickets are calculated using the current time.

 For example, a job scheduled for 06:00 may run at 09:00. A ticket that became overdue at 08:00 would be included, even though it was not overdue when the job was scheduled.
This ensures the job always produces the same results, whether it runs on time or is retried later.
```

notification_date, notification_time = latest_local_schedule_time(...)
and then save it in the job:
  config={
      # ...
      "notification_time": notification_time.isoformat(),
  }

daily_grievance_digest_async_task_action -> pass this to DailyDigestService (self.notification_time)
```

MarekBiczysko (changes requested):

src/frontend/src/containers/pages/grievances/MyTasksPage.tsx
pkujawa: Agree on including the GRIEVANCES_VIEW_LIST_PERMISSIONS but the GRIEVANCES_UPDATE is just needed for the bulk-assign button. Users without this perm should be able to view the list and assign just from the ticket detail

nerjan (approved):

MarekBiczysko (approved):

PR #6397, opened 2026-09-12
Add TargetPopulation proxy model and admin enhancements

domdinicola:
TargetPopulation proxy,
Payment Plan is an evolution TargetPopulation, it would be useful to create a proxy model to help.

Feedback: nullable fk to program
Message: programme not null
Survey: programme not null


src/hope/admin/target_population.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def get_form(self, request: HttpRequest, obj: Any | None = None, change: bool = False, **kwargs: Any) -> Any:
+        request._payment_plan_obj = obj
+        return super().get_form(request, obj, change, **kwargs)
+
+    def save_model(self, request: HttpRequest, obj: TargetPopulation, form: Any, change: bool) -> None:
+        old_payment_plan = copy_model_object(PaymentPlan.objects.get(pk=obj.pk)) if change and obj.pk else None
+        super().save_model(request, obj, form, change)
+        # skip a no-op log when an edit touched only fields outside ACTIVITY_LOG_MAPPING
+        if old_payment_plan is not None and not create_diff(old_payment_plan, obj, PaymentPlan.ACTIVITY_LOG_MAPPING):
```
Jeeziorny, line 109: This seems to be self-explanatory

tests/unit/admin/test_view_on_ui.py
```diff
@@ -302,12 +273,16 @@ def test_view_on_ui_without_original():
...
 
 @pytest.mark.django_db
-def test_view_on_ui_without_frontend_url(program, cycle):
-    plan = PaymentPlanFactory(program_cycle=cycle(program()))
-    payment = PaymentFactory(parent=plan)
-    btn = _button(payment)
-    _handler(Payment).func(site._registry[Payment], btn)
-    assert btn.href is None
+def test_view_on_ui_without_frontend_url(program):
+    ma = site._registry[Program]
```
Jeeziorny: what's `ma`?

domdinicola: model admin :)

tests/extras/test_utils/factories/payment.py
```diff
@@ -155,6 +163,7 @@ class Meta:
     status_date = factory.LazyFunction(timezone.now)
     currency = factory.SubFactory(CurrencyFactory)
     business_area = factory.SelfAttribute("parent.business_area")
+    program = factory.SubFactory(ProgramFactory, business_area=factory.SelfAttribute("..business_area"))
```
pkujawa: This creates a new Program per payment, so payment.program is different from program of its PP.
`program = factory.SelfAttribute("parent.program")` - like line above for BA

src/hope/admin/target_population.py
```diff
@@ -0,0 +1,124 @@
...
+from django.http import HttpRequest
+
+from hope.admin.utils import HOPEModelAdminBase, ViewOnUiMixin
+from hope.apps.activity_log.utils import copy_model_object, create_diff
+from hope.apps.utils.security import is_root
+from hope.models import PaymentPlan, TargetPopulation, log_create
+
+
+@admin.register(TargetPopulation)
+class TargetPopulationAdmin(ViewOnUiMixin, HOPEModelAdminBase):
```
pkujawa: This is `PaymentPlanAdmin` copied, but without some logic. `plan_type` is readonly on `PaymentPlanAdmin` but editable here.
`payment_plan_purposes` lost `filter_horizontal` and the `formfield_for_manytomany` that filters them to the program's purposes, so it shows from every program. 
But there is `get_form` here, from the `PaymentPlanAdmin`, but `request._payment_plan_obj` was used only for the method that is missing here now. 

Need to be fixed here or we could have a `BasePaymentPlanAdmin` (without the buttons and inlines that we have on PP) and inherit in both places and override here `list_display`, `list_filter` and `frontend_url`

```
@admin.register(TargetPopulation)
class TargetPopulationAdmin(BasePaymentPlanAdmin):
    list_display = (...)
    list_filter = (...)

    def frontend_url(self, obj: TargetPopulation) -> str:
        return f"/{obj.business_area.slug}/programs/{obj.program.code}/target-population/{obj.id}"
```

src/hope/apps/payment/api/serializers.py
```diff
@@ -1549,13 +1549,17 @@ def get_fsp_name(self, obj: Payment) -> str:
...
-        ):
+        if "can_view_fsp_auth_code" in self.context:
+            can_view = self.context["can_view_fsp_auth_code"]
+        else:
+            if "request" not in self.context:
+                return ""
+            user = self.context["request"].user
+            can_view = user.has_perm(
+                Permissions.PM_VIEW_FSP_AUTH_CODE.value,
+                obj.program or obj.business_area,
```
pkujawa: just `obj.program` - it's not nullable now

src/hope/apps/accountability/migrations/0010_migration.py
```diff
@@ -0,0 +1,42 @@
...
+            model_name="survey",
+            name="program",
+            field=models.ForeignKey(
+                blank=True,
+                null=True,
+                on_delete=models.deletion.SET_NULL,
+                related_name="surveys",
+                to="program.Program",
+            ),
+        ),
```
pkujawa: this must be some leftover after manual changes, as Survey model was not edited and this is the current logic there?
Also, the non-nullable program will not be applied in this PR? (If no, please remove this point from description)

src/hope/apps/household/api/views.py
```diff
@@ -207,12 +207,19 @@ def payments(self, request: Any, *args: Any, **kwargs: Any) -> Any:
         )
 
         payments = self.filter_queryset(payments)
+        context = {
+            **self.get_serializer_context(),
+            "can_view_fsp_auth_code": request.user.has_perm(
+                Permissions.PM_VIEW_FSP_AUTH_CODE.value,
+                hh.program or hh.business_area,
```
pkujawa: just `hh.program`

pkujawa (changes requested):

tests/extras/test_utils/factories/payment.py
domdinicola: good catch, I'll fix also elsewhere

src/hope/models/message.py
Jeeziorny, line 71: PR description:
```
Message: programme not null
```
Actual change
```
when program is deleted, set message.program to null
```

Am I missing something?

src/hope/apps/accountability/migrations/0010_migration.py
```diff
@@ -0,0 +1,42 @@
...
+            name="program",
+            field=models.ForeignKey(
+                blank=True,
+                null=True,
+                on_delete=models.deletion.SET_NULL,
+                related_name="messages",
+                to="program.Program",
+            ),
+        ),
+        migrations.AlterField(
```
Jeeziorny: Same goes for survey. PR description says something different that it's implemented here

src/hope/admin/target_population.py
Jeeziorny: This looks copied from `PaymentPlanAdmin`. `request._payment_plan_obj = obj` is set here, but its only reader is `PaymentPlanAdmin.formfield_for_manytomany` (payment_plan.py:287), which this class doesn't inherit

Jeeziorny (changes requested):
I feel like some part of this PR is taken from other places in HOPE and violates DRY rule. I'm pausing reviewing this PR as long as it has symptoms of blind copy paste

src/hope/apps/payment/api/serializers.py
```diff
@@ -1549,13 +1549,17 @@ def get_fsp_name(self, obj: Payment) -> str:
...
 
     def get_fsp_auth_code(self, obj: Payment) -> str:
-        if "request" not in self.context:
-            return ""
-        user = self.context["request"].user
-        if not user.has_perm(
-            Permissions.PM_VIEW_FSP_AUTH_CODE.value,
-            obj.program or obj.business_area,
-        ):
+        if "can_view_fsp_auth_code" in self.context:
```
johniak: Only two of the four callers pass `can_view_fsp_auth_code`; `PaymentViewSet.list` and `PaymentVerificationRecordViewSet.list` still take the per-row `has_perm` branch. Caching the answer per `program_id` on the serializer instance fixes every caller and removes the duplicated context block from both views.

tests/unit/admin/test_view_on_ui.py
```diff
@@ -302,12 +273,16 @@ def test_view_on_ui_without_original():
 
 
 @pytest.mark.django_db
-def test_view_on_ui_without_frontend_url(program, cycle):
-    plan = PaymentPlanFactory(program_cycle=cycle(program()))
-    payment = PaymentFactory(parent=plan)
-    btn = _button(payment)
-    _handler(Payment).func(site._registry[Payment], btn)
-    assert btn.href is None
+def test_view_on_ui_without_frontend_url(program):
```
johniak: `test_frontend_url_survey_without_program` and `test_frontend_url_message_without_program` were removed, but `Survey.program` / `Message.program` are still nullable and the `if not obj.program` guards in `admin/survey.py` and `admin/message.py` are live. `program=None` still works with the new factory defaults, so both tests can stay.

tests/unit/admin/test_view_on_ui.py
```diff
@@ -302,12 +273,16 @@ def test_view_on_ui_without_original():
...
+    programme_admin = site._registry[Program]
+    original = programme_admin.frontend_url
+    programme_admin.frontend_url = lambda obj: None  # type: ignore[method-assign]
```
johniak: `monkeypatch.setattr(programme_admin, "frontend_url", lambda obj: None)` replaces the try/finally and the `type: ignore`.

src/hope/apps/payment/api/serializers.py
```diff
@@ -1549,13 +1549,17 @@ def get_fsp_name(self, obj: Payment) -> str:
...
+            can_view = self.context["can_view_fsp_auth_code"]
+        else:
```
johniak: `else:` followed by a bare `if` can be flattened to `elif`.

johniak (changes requested):

src/hope/apps/registration_data/migrations/0018_migration.py
```diff
@@ -31,4 +33,9 @@ class Migration(migrations.Migration):
             model_name="registrationdataimport",
             name="deduplication_engine_status",
         ),
+        migrations.AlterField(
```
johniak, line 36: This migration is already on `develop` and `origin/staging` (merged 2026-09-17 in #6256), so it's recorded as applied there. The new `AlterField` will never run on those environments: the column stays nullable while the model says NOT NULL, and only fresh installs / prod get the constraint. This needs to be a new `0019_migration.py`.

johniak (changes requested):

domdinicola:
@Jeeziorny can you pls clarify your comment?

johniak (dismissed):

Jeeziorny:
@domdinicola sure, DRY means don't repeat yourself

domdinicola:
@kamilswiec I know what dry means :) I meant to which part of the PR you're referring

johniak:
> @kamilswiec I know what dry means :) I meant to which part of the PR you're referring

Kamil Siwiec is not working in hope for about 4 years :D

domdinicola:
@Jeeziorny I know what dry means :) I meant to which part of the PR you're referring

@johniak thx :D

PR #6398, opened 2026-09-14
301335: Latin fields - testing fixes

Jeeziorny:

This PR addresses bug / bad behaviour caught by Habiba during tests on staging.

### Latin-name validation too strict

The *_latin regex rejected leading/trailing whitespace, multiple spaces between words - so names like " John  O'Brien  " failed on import and other places. 

#### Fix

Let's normalise those strings before validation by removing white chars.

#### Places where I applied normalise before validate
1. RDI XLSX upload,
2. RDI Kobo import,
3. Universal update script,
4. Grievance add/update individual,
5. Grievance FE (validateGrievance.ts).


src/hope/models/individual.py
```diff
@@ -64,7 +64,7 @@
 )
 
 ascii_name_validator = RegexValidator(
-    regex=r"^[A-Za-z]+(?:[ '-][A-Za-z]+)*$",
+    regex=r"^\s*[A-Za-z]+(?:[\s'-]+[A-Za-z]+)*'?\s*$",
```
kosior: Isn't the regex too broad now? Previously the following cases would fail:
```
"Anna\nKovalska"
"Anna\tKovalska"
"Anna\xa0Kovalska"
"Anna--Kovalska"
"Anna'''Kovalska"
"Anna - ' - Kovalska"
```

Also, with the new regex, would the values be saved with trailing/leading spaces in the db?

Maybe a better option would be to normalize the string before the validation. It would also result in cleaner db values. What do you think?

mmaciekk (approved):

src/hope/models/individual.py
Jeeziorny: I like your idea. Let mi fix that.

kosior (approved):

PR #6399, opened 2026-09-14
Add async index for individual program_registration_id

domdinicola:
This index is used for retrieving households and individuals for the beneficiary portal.


src/hope/models/individual.py
```diff
@@ -450,6 +450,7 @@ class Individual(
         max_length=100,
         blank=True,
         null=True,
+        db_index=True,
```
kosior, line 453: First commit added index to `Meta.indexes` which I think was better option and it's suggested by Django docs. Additionally I think there will be a discrepancy between test database and non-test databases because we run tests with `--no-migrations`. 

What's the reason for `db_index=True` change?

kosior (changes requested):
For a thorough review I'll need one more information: 

Which query in the beneficiary portal is supposed to use this index?

src/hope/models/individual.py
domdinicola: consistency... we use db_index=True in the rest of the project..
maybe we can discuss about refactoring to use Meta.indexes

kosior: Ok, I think we'd plan the refactor this would be a good place to start - less work later on.

The difference between `test db` and `non-test db` will be there if we'll stick with the `db_index=True`. I think the test database will have an extra `_like` index. SQL generated when only db_index is used:
```
BEGIN;
--
-- Alter field program_registration_id on individual
--
CREATE INDEX "household_individual_program_registration_id_598b7aa7" ON "household_individual" ("program_registration_id");
CREATE INDEX "household_individual_program_registration_id_598b7aa7_like" ON "household_individual" ("program_registration_id" varchar_pattern_ops);
COMMIT;
```

kosior: Which query exactly are we trying to optimize?

domdinicola: I think best is to have all the change in a separate PR.
we need to be able to get records using that field, beneficiary portal will use individual_program_registration_id  to access the record and we want to do that in an efficient way

kosior: Ok, got it. Would you like me to do some measurements for the query? Like before/after index?

domdinicola: no it's not needed!

kosior (approved):

pkujawa (approved):

saxix (dismissed):

PR #6400, opened 2026-09-15
[WIP][FEATURE]340166: Add structure file blob paths for all FileFIelds

nerjan:
[AB#340166](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/340166)


domdinicola:
@nerjan  pls recreate migrations

src/hope/apps/core/upload_paths.py
kosior: Existing files stay flat in the storage root, only new uploads use the new structure. Is migrating the old blobs out of scope, or should it be a follow-up ticket?

src/hope/apps/core/upload_paths.py
```diff
@@ -0,0 +1,152 @@
...
+    `<upload year>/<business area>/_unassigned/` and `<upload year>/_global/` when `instance`
+    reaches no programme or no business area. Set as the `upload_to` of every file field, and
+    called the same way before `default_storage.save` wherever a file is written directly.
+    """
+    program = get_program(instance)
+    if program is None:
+        return build_upload_path(filename, business_area_slug=get_business_area_slug(instance))
+    return build_upload_path(
+        filename,
+        year=program.start_date.year,
```
kosior: `start_date` is editable. If it moves across a year boundary, a programme's files end up split between two <year>/ folders. Is it OK? Do you know if this has been considered a problem?

src/hope/apps/core/upload_paths.py
```diff
@@ -0,0 +1,152 @@
...
+        return instance
+    if owner_path := _registered_path(OWNER_PATHS, type(instance)):
+        try:
+            owner = nested_getattr(instance, owner_path, None)
+        except (ObjectDoesNotExist, ValidationError, ValueError, TypeError):
+            return None
+        return get_program(owner)
+    program = nested_getattr(instance, _registered_path(PROGRAM_PATHS, type(instance)) or DEFAULT_PROGRAM_PATH, None)
+    if isinstance(program, Manager):
+        return program.order_by("id").first()
```
kosior: A grievance ticket can have more than one programme. `get_program` then takes `programs.order_by("id").first()`, so the choice is arbitrary. Is it intended? Just curious what was the general purpose for this whole change and if we need to worry about any future cleanups or archivisation.

src/hope/apps/core/upload_paths.py
nerjan: it would cost to much to move all these files because of how it is handled. I am not sure what kind of blob storage we are using but the basic example is basically copying the file to new location and then deleting the old ones. With the amount of files we have this idea was dropped.

src/hope/apps/core/upload_paths.py
nerjan: It might have more than 1 program only in db schema, it is not possible from UI perspective.

src/hope/apps/core/upload_paths.py
nerjan: as discussed on daily its not an issue

PR #6401, opened 2026-09-15
[AB#341491]- Household data change tickets: flex fields and other/unknown sex group counts

mmaciekk:
[AB#341491](https://unicef.visualstudio.com/ICTD-HCT-MIS/_workitems/edit/341491)
## Summary

Household data change grievance tickets did not work properly with household flex fields, and silently dropped the *Gender group with unknown sex* / *Gender group with other sex* core fields.

### Flex fields in household data change tickets
- `RequestedHouseholdDataChange` and `GrievanceDetailsToolbar` now read `household_data.flex_fields` (the REST client does not camelize `householdData`, so `flexFields` was always empty): flex fields are counted, selectable and approvable.
- Roles are included in the "select all" count.
- `RequestedHouseholdDataChangeTable` no longer passes a fresh `[]` to `useArrayToDict` on every render.
- `initdemo` seeds household flex attributes (incl. *HH Total Eligible Individuals*) for the Afghanistan Test Program so the *Additional Registration Information* section is non-empty locally.
- Test: closing a ticket applies only approved flex fields and casts string values.

### Other/unknown sex group counts dropped from the ticket
`HouseholdUpdateDataSerializer` is an explicit allowlist and did not declare `other_sex_group_count` / `unknown_sex_group_count`, even though both are offered in the field picker (`Scope.HOUSEHOLD_UPDATE`). DRF discarded them from `validated_data`, so selecting e.g. *Gender group with unknown sex* + *HH Total Eligible Individuals* showed only the flex field in the requested data change table. Both fields are now accepted (create and edit paths share the serializer); regression test added.

### Still not accepted by the serializers (follow-up)
Fields offered in the pickers but not declared in the update serializers, each needing service changes beyond the allowlist:
- household: `consent_sign` (image), `facility`, `facility_admin_area` (FK lookups)
- individual/people: `ind_identification_key`, `hh_identification_key` (core-field name differs from the model attribute)

## Test plan
- [x] `tests/unit/apps/grievance/test_grievance_create_data_change.py`, `test_grievance_ticket_update.py`, `services/data_change/test_household_data_update_service.py`
- [x] `tests/unit/apps/core/test_demo_data_flex_fields.py`, `RequestedHouseholdDataChange.test.tsx`
- [ ] Manually: create a household data change ticket selecting *Gender group with unknown sex* + *HH Total Eligible Individuals*; both rows appear, can be approved, and closing the ticket updates the household


Jeeziorny:
PR Description summary does not mention the issue reported in the ticket. I'd like to understand what was causing the issue reported in the ticket and how this PR solved that. Please explain me in the comment if summary is valid and me is who needs special treatment :D

src/frontend/src/components/grievances/GrievancesPhotoModals/GrievanceFlexFieldPhotoModal.tsx
```diff
@@ -38,14 +38,20 @@ export function GrievanceFlexFieldPhotoModal({
     return null;
   }
 
+  // individualData is camelized by the REST client, householdData is not
```
Jeeziorny: There are a few occurences of information, that householdData is not camelized by REST client. Is this redundancy on purpose?

src/hope/apps/core/management/commands/demo_data/flex_fields.py
```diff
@@ -0,0 +1,55 @@
...
+    ("living_situation_h_f", FlexibleAttribute.SELECT_ONE, "Living Situation"),
+)
+LIVING_SITUATION_CHOICES = (
+    ("owner", "Owner"),
+    ("renter", "Renter"),
+    ("hosted", "Hosted"),
+)
+
+
+def generate_household_flex_fields() -> None:
```
Jeeziorny, line 20: Good job adding that.

Jeeziorny:
> HouseholdUpdateDataSerializer is an explicit allowlist and did not declare other_sex_group_count / unknown_sex_group_count, even though both are offered in the field picker 


Should they be really offered in the field picker? Code from `HouseholdDataUpdateService`
```python
    Household.objects.filter(id=household.id).update(flex_fields=merged_flex_fields, **only_approved_data)

    invalidate_household_list_cache(household.program_id)
    updated_household = Household.objects.get(id=household.id)

    if admin_area_title.get("value") is not None and is_approved(admin_area_title):
        area = Area.objects.filter(p_code=admin_area_title.get("value")).first()
        updated_household.set_admin_areas(area)
    # update Roles
    for role in roles_data:
        # update only approved roles
        if role.get("approve_status") is True:
            individual_id = role["individual_id"]
            individual = get_object_or_404(Individual, id=individual_id)
            handle_role(household, individual, role.get("value"))

    new_household = Household.objects.select_for_update().get(id=household.id)
    recalculate_data(new_household)
```

That mean:
1. Update HH with approved data (in this case those `*_count` values)
2. `recalculate_data`. 

Should we give a user ability to change fields that are recalculated by the system?

src/hope/apps/grievance/services/data_change/household_data_update_service.py
Jeeziorny, line 144: I see the pattern in repo, where right after `verify_flex_fields` `save_images` is being called which isn't the case in `hh_data_update_service`. Probably not something to be addressed in this PR, but shouldn't we call `save_images` here as well? @nerjan @johniak  Janek & Janek developers

src/frontend/src/components/grievances/GrievancesPhotoModals/GrievanceFlexFieldPhotoModal.tsx
```diff
@@ -38,14 +38,20 @@ export function GrievanceFlexFieldPhotoModal({
...
+    ? data.ticketDetails?.individualData
+    : data.ticketDetails?.householdData;
   const flexFields = isIndividual
-    ? data?.ticketDetails?.individualData?.flex_fields
-    : data.ticketDetails?.householdDataUpdateTicketDetails?.householdData
-        ?.flex_fields;
+    ? ticketData?.flexFields
+    : ticketData?.flex_fields;
+  // core IMAGE fields (photo, consent_sign) sit next to flex_fields, not inside them
+  const change = flexFields?.[field.name] ?? ticketData?.[field.name];
```
johniak: On the `isIndividual` branch `ticketData` is `individualData`, which `deepCamelize` does recurse into (only `household_data` is in `notCamelizedKeys`), so the key is `consentSign` and `ticketData?.['consent_sign']` always misses. The consent signature renders as `-` on every people ticket; `photo` only works because it is one word. `ticketData?.[isIndividual ? camelCase(field.name) : field.name]` fixes it.

src/hope/apps/grievance/services/data_change/utils.py
```diff
@@ -434,6 +434,17 @@ def handle_photo(photo: InMemoryUploadedFile | str | None, photoraw: str | None)
     return None
 
 
+def handle_image_field(data: dict, field_name: str) -> None:
```
johniak, line 437: `handle_image_field` is the same logic as `_handle_photo_field` in `individual_data_update_service.py`, and the two now run on adjacent lines (`:113-114`, `:199-200`). Drop the private one and call `handle_image_field(individual_data, "photo")`.

src/frontend/src/components/grievances/GrievancesPhotoModals/GrievanceFlexFieldPhotoModal.tsx
```diff
@@ -38,14 +38,20 @@ export function GrievanceFlexFieldPhotoModal({
     return null;
   }
 
+  // individualData is camelized by the REST client, householdData is not
+  const ticketData = isIndividual
+    ? data.ticketDetails?.individualData
+    : data.ticketDetails?.householdData;
```
johniak: The editable twin, `GrievanceFlexFieldPhotoModalEditable.tsx:47-51`, still reads `ticketDetails?.householdDataUpdateTicketDetails?.householdData?.flexFields` — the dead path removed here — and then indexes it without a guard, so it throws instead of rendering. Worth fixing in the same PR while the shape is fresh.

johniak (changes requested):

mmaciekk:
> PR Description summary does not mention the issue reported in the ticket. I'd like to understand what was causing the issue reported in the ticket and how this PR solved that. Please explain me in the comment if summary is valid and me is who needs special treatment :D

I updated the PR description. It now lists HH Total Eligible Individuals specifically.

src/frontend/src/components/grievances/GrievancesPhotoModals/GrievanceFlexFieldPhotoModal.tsx
mmaciekk: Not on purpose - it was the same sentence copy-pasted to four places, which is a good sign the knowledge belongs somewhere else. It now lives once in `grievances/utils/ticketData`, next to the `notCamelizedKeys` rule it describes, and the readers call helpers instead of restating it.

That also caught three more places where the asymmetry was simply wrong - see the two replies below and the new section in the PR description.

src/frontend/src/components/grievances/GrievancesPhotoModals/GrievanceFlexFieldPhotoModal.tsx
mmaciekk: Correct, thank you - fixed. `EditPeopleDataChange/CurrentValue.tsx` sends every IMAGE field through this modal with `isIndividual`, so `consent_sign` looked up `individualData['consent_sign']` while the client had stored it as `consentSign`, and the signature rendered as a dash on every people ticket. `photo` survived only because camelCase leaves it alone.

The lookup now goes through `getTicketFieldChange`, which camelizes core field names on the individual side and leaves flex field names alone. `GrievanceFlexFieldPhotoModal.test.tsx` gained the two individual core field cases that were missing - that gap is why it slipped through.

src/frontend/src/components/grievances/GrievancesPhotoModals/GrievanceFlexFieldPhotoModal.tsx
mmaciekk: Agreed, done in this PR. It was worse than rendering nothing: `flexFields` came back `undefined` and line 55 indexed it, so editing a ticket with an image field threw instead of rendering. Accepting `consent_sign` made that reachable far more often.

It now reads `ticketDetails.householdData` / `ticketDetails.individualData` through the same helper, and a field the ticket does not change falls through to the file input instead of throwing. Those were the last two occurrences of the `*DataUpdateTicketDetails` path left in the codebase. New `GrievanceFlexFieldPhotoModalEditable.test.tsx` covers household flex, individual core and the empty case.

While checking the other readers of the same shape, two more were wrong:

- `editGrievanceUtils.ts` read `householdData.flexFields`, so **editing** a household data change ticket dropped every flex field row and added an empty `flex_fields` row instead.
- `countApprovedAndUnapproved` in `GrievanceDetailsToolbar` only recognised `approve_status`, so individual and people tickets counted 0 approved and 0 unapproved and the close dialog showed no warning at all. The counting moved into `countTicketChanges` in the helper, where it is unit-tested for both shapes.

johniak (approved):

nerjan:
> > HouseholdUpdateDataSerializer is an explicit allowlist and did not declare other_sex_group_count / unknown_sex_group_count, even though both are offered in the field picker
> 
> Should they be really offered in the field picker? Code from `HouseholdDataUpdateService`
> 
> ```python
>     Household.objects.filter(id=household.id).update(flex_fields=merged_flex_fields, **only_approved_data)
> 
>     invalidate_household_list_cache(household.program_id)
>     updated_household = Household.objects.get(id=household.id)
> 
>     if admin_area_title.get("value") is not None and is_approved(admin_area_title):
>         area = Area.objects.filter(p_code=admin_area_title.get("value")).first()
>         updated_household.set_admin_areas(area)
>     # update Roles
>     for role in roles_data:
>         # update only approved roles
>         if role.get("approve_status") is True:
>             individual_id = role["individual_id"]
>             individual = get_object_or_404(Individual, id=individual_id)
>             handle_role(household, individual, role.get("value"))
> 
>     new_household = Household.objects.select_for_update().get(id=household.id)
>     recalculate_data(new_household)
> ```
> 
> That mean:
> 
>     1. Update HH with approved data (in this case those `*_count` values)
> 
>     2. `recalculate_data`.
> 
> 
> Should we give a user ability to change fields that are recalculated by the system?

recalculate_data overwrites them only when the data collecting type has recalculate_composition on. So for other cases it is needed.

src/hope/apps/grievance/services/data_change/utils.py
nerjan: True, fixed

Jeeziorny (approved):
LGTM

src/hope/apps/grievance/services/data_change/household_data_update_service.py
nerjan: I added save_images even tho it is not reachable yet, but its small price for the future.

PR #6402, opened 2026-09-15
339485 payment plan not eligible status

MarekBiczysko:
[AB#339485](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/339485)


mmaciekk (dismissed):

domdinicola (dismissed):

src/hope/apps/payment/api/views.py
```diff
@@ -2655,20 +2666,54 @@ class PaymentViewSet(
...
             id=self.kwargs["payment_id"],
             parent__program_cycle__program=self.program,
         )
 
     def get_queryset(self) -> QuerySet:
         parent = self.payment_plan
+        if self.action in ("not_eligible", "not_eligible_count"):
+            if not self.request.user.is_superuser:
+                raise PermissionDenied("Only superusers can view Not Eligible payments.")
+            return with_payment_related_data(parent.payment_items.filter(status=Payment.STATUS_NOT_ELIGIBLE))
```
pkujawa: get_object above hides Not Eligible payments from non-superusers, but this viewset doesn't, and it's what is used in country search and the grievance payment lookup. So user sees a "Not Eligible" row and gets a 404 when they click it — and in the grievance lookup they can attach that payment to a ticket first. The household payments action already handles this by calling .eligible(); can this one do the same, or  mirror the superuser check?

    def get_queryset(self) -> QuerySet:
        queryset = super().get_queryset()
        if not self.request.user.is_superuser:
            queryset = queryset.exclude(status=Payment.STATUS_NOT_ELIGIBLE)
        return with_payment_related_data(queryset).order_by("-created_at")

tests/unit/apps/payment/test_payment_view.py
```diff
@@ -22,6 +22,7 @@
 )
 from hope.apps.account.permissions import Permissions
 from hope.apps.household.const import ROLE_ALTERNATE
+from hope.apps.payment.api.views import PaymentViewSet
```
pkujawa: Seems that importing `PaymentViewSet` here at the module level makes this file fail to collect on its own, because of a circular import.

monkeypatch for this is just used in 1 test and can be dropped and we can use the format like in all other tests here

```
    response = not_eligible_payment_context["client"].get(url)
    results = response.json()["results"]
```

src/hope/apps/payment/api/views.py
pkujawa, line 2748: PaymentGlobalViewSet and ProgramViewSet.payments both still list Not Eligible payments to anyone with `PM_VIEW_DETAILS`. Country search goes through PaymentGlobalViewSet; the grievance payment lookup goes through either, depending on whether a program is selected.
They should also have the superuser check

src/hope/apps/payment/api/views.py
```diff
@@ -2647,6 +2650,14 @@ class PaymentViewSet(
...
             Permissions.PAYMENT_VERIFICATION_VIEW_DETAILS,
         ],
+        "not_eligible": [
+            Permissions.PM_VIEW_DETAILS,
+            Permissions.PAYMENT_VERIFICATION_VIEW_DETAILS,
+        ],
+        "not_eligible_count": [
+            Permissions.PM_VIEW_DETAILS,
+            Permissions.PAYMENT_VERIFICATION_VIEW_DETAILS,
+        ],
```
pkujawa, line 2661: So these two actions are only allowed for superusers, and are checked inside `get_queryset`. 
Non-superusers without these perms will get 403 listing them, but they will not grant them access. While the superusers pass the permission check always.
But we cannot put an empty list here either since it would be False then. We could use a different permission class with superuser check for these two, but I guess it isn't worth it.
Let's maybe at least drop a comment here that the real gate is being a superuser.

pkujawa (changes requested):

src/hope/apps/payment/api/views.py
MarekBiczysko: It seems like all of these should show only .eligible() payments (they we're visible as PENDING payments before and this was also wrong)

mmaciekk (approved):

pkujawa (approved):

PR #6405, opened 2026-09-15
328305_ezCash_vision_integration_flow_fix

MarekBiczysko:
Implemented the two-callback Vision flow:

  - First callback with blank status and fc_num sets PP_CREATED, stores vision_payplanSno, and returns HTTP 200.
  - Second callback requires status="SUCCESS" and a valid fc_num.
  - PP_CREATED supports later callbacks and Django admin FC recovery.
  - Missing Vision ID after PP_CREATED is treated as callback failure.
  - Frontend continues polling every 60 seconds while PP_CREATED.
  - Admin recognizes PP_CREATED as confirmed Vision receipt.


domdinicola (dismissed):

src/hope/contrib/vision/services.py
```diff
@@ -166,33 +167,46 @@ def process_callback(
...
+        acknowledgement_can_update_status = payment_plan.vision_status in {
+            VisionStatus.SEND_FAILED.value,
+            VisionStatus.WAITING_FOR_CALLBACK.value,
+            VisionStatus.PP_CREATED.value,
+        }
+        fc_assignment_failed = False
+        if is_payment_plan_created_acknowledgement:
+            # A repeated creation acknowledgement must not replace a later FC result or failure.
+            if acknowledgement_can_update_status:
+                vision_data["vision_id"] = vision_payment_plan_id
```
Jeeziorny, line 171: Shouldn't we set 
```python
vision_data["sent"] = True
```
here?

src/hope/contrib/vision/services.py
```diff
@@ -166,33 +167,46 @@ def process_callback(
...
 
-        vision_data["vision_id"] = vision_payment_plan_id
-        if fc_num:
-            vision_data["fc_num"] = fc_num
+        is_payment_plan_created_acknowledgement = not vision_result and not fc_num
+        acknowledgement_can_update_status = payment_plan.vision_status in {
+            VisionStatus.SEND_FAILED.value,
+            VisionStatus.WAITING_FOR_CALLBACK.value,
+            VisionStatus.PP_CREATED.value,
+        }
```
pkujawa: This is RECOVERABLE_STATUSES minus the three FC failures, and it's checked even on the else case.
Can we put it in choices.py and do the check inside the if?

pkujawa: This is RECOVERABLE_STATUSES minus the three FC failures, and it's checked even on the else case.
Can we put it in choices.py and do the check inside the if?

pkujawa (approved):

Jeeziorny (approved):

PR #6407, opened 2026-09-16
[341640] Count only active beneficiaries in population totals

pkujawa:
[AB#341640](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/341640)

- Payment plan totals use the same rule as household size, so the two cannot disagree.
- Withdrawing a household now recalculates its composition and KAB counters.  Mass Withdraw in admin and the withdraw household grievance both go through the same service, so both are fixed. Withdrawing one individual already worked.
- Program household and individual counts skip withdrawn people. The recount already ran on withdrawal, it just counted the withdrawn ones too, so the number never moved.
- active_individuals no longer counts non beneficiaries, so it matches size. This also means a household left with only non beneficiary records now withdraws itself, which it did not before.
- Marking an individual duplicate or distinct in Needs Adjudication now recalculates the household.
- Removed POST /households/{id}/withdraw. Nothing used it, it seems like a leftover after GraphQL refactor.
- Needs Adjudication recalculates the household once at the end of closing the ticket, after the roles are reassigned. Before it ran once per individual, and before the reassignment, so child_hoh was worked out from the head that was being retired.
- Marking somebody distinct now recounts the programme. So does a relationship change through a data update ticket, the XLSX update or the universal update script.
- Editing an individual in admin recalculates now too.
- Household.withdrawn is read only in admin. Ticking it only half withdrew the household, the members stayed active. The Withdraw button on the same page does it properly.
- A collision merge now recalculates the household it merged into. It withdraws the members that are not in the new record, and its counters were overwritten with the incoming numbers and never recalculated.
- New recount_program_size command. The stored counters change meaning here and nothing else refreshes them, so a quiet programme would keep the old number.

Frontend
- The grievance household lookup now sends withdrawn=false, the same as the individual lookup already did. Withdrawn households no longer show up when picking a household for a ticket.
- The Status filter is hidden in that lookup, so a withdrawn household cannot be picked on purpose either.

## To run after deploy

- python manage.py recount_program_size, once per environment. Every existing programme keeps its old total until it runs, and until then the number of households filter on the programme list will not match the column next to it.


src/hope/apps/grievance/services/needs_adjudication_ticket_services.py
```diff
@@ -421,6 +422,7 @@ def mark_as_duplicate_individual(
         household.refresh_from_db()
         if household.active_individuals.count() == 0:
             household.withdraw()
+        recalculate_data(household)
```
kosior, line 425: I'm just thinking about the placement of the household data recalculation. The flow looks smth like this:

- `close_needs_adjudication_new_ticket`
    - `mark_as_duplicate_individual`
        - `recalculate_data`  
    - `reassign_roles_on_marking_as_duplicate_individual_service`

Would it make sense to move `recalculate_data` after the `reassign_roles_on_marking_as_duplicate_individual_service`? Does the `reassign_roles_on_marking_as_duplicate_individual_service` update the data somehow that would affect the `recalculate_data`?

Also by moving the `recalculate_data` potentially we could remove double recalculation for a household.

MarekBiczysko (approved):

src/hope/models/program.py
```diff
@@ -323,8 +324,10 @@ def get_total_number_of_households_from_payments(
         )
 
     def adjust_program_size(self) -> None:
-        self.household_count = self.households.count()
-        self.individual_count = self.individuals.count()
+        self.household_count = self.households.filter(withdrawn=False).count()
+        self.individual_count = (
+            self.individuals.filter(withdrawn=False, duplicate=False).exclude(relationship=NON_BENEFICIARY).count()
+        )
```
kosior, line 330: The counts now depend on withdrawn, duplicate and relationship, so any flow that changes those fields should trigger a recount. Are there new places where we'd want to call adjust_program_size?

Places I'd like to confirm:

1. `individual_marked_as_distinct` - the signal isn't connected, and Household.unwithdraw() doesn't send one
2. `individual_data_update_service.py` - relationship can change to or from NON_BENEFICIARY
3. `individual_xlsx_update.py` - same, it only schedules the household recalculation
4. Universal update script - same, schedule_population_recalculation recalculates households only
5. Django admin - withdrawn, duplicate and relationship are editable on Individual. Probably fine to skip.

src/hope/apps/grievance/services/needs_adjudication_ticket_services.py
pkujawa: Good catch, it does matter. When the duplicate is the head of household the reassignment gives the head role to somebody else, and `recalculate_data` checks `child_hoh` and `fchild_hoh`, so doing it first meant those flags came from the person being changed.
It is now one `recalculate_data` per affected household at the end of `close_needs_adjudication_new_ticket`, after the reassignment, which also removes the double recalculation - before this, a ticket with several duplicates in one household recalculated it once per individual.

src/hope/models/program.py
pkujawa: Agreed, all done.

`individual_marked_as_distinct` connected, `Household.unwithdraw()` still sends nothing on its own, but the only place that calls it outside `HouseholdBulkWithdrawService` is the needs adjudication distinct path, which the signal covers, and the bulk service already recounts.

`individual_data_update_service`, `individual_xlsx_update` and the universal update script recount when relationship changes.

Django admin is done for Individual. For Household I actually made `withdrawn` read-only instead, because setting the flag by hand did it incorrectly anyway - the members stayed active, documents stayed valid and their tickets stayed open. The Withdraw button on the same page goes through `HouseholdBulkWithdrawService` and handles it correctly.

Also added `recount_program_size` command that we need to run once per environment, to update these.

MarekBiczysko (approved):

kosior (approved):

PR #6419, opened 2026-09-18
Fix duplicate list/count requests in REST tables

mmaciekk:
## What was wrong

Opening a list page (e.g. Grievance Tickets → System-Generated) sent the same list and count requests twice. The two requests differed only in `ordering=-created_at`.

The cause was in how tables manage their state. The page component built the query parameters from the filters and fired the request right away. Then `UniversalRestTable` ran an effect that added the sort order, page size and offset and wrote them back into the page's state, which produced a second, different request. Two components owned the same state and kept syncing it with effects.

Checking all 46 tables, roughly 25–30 of them double-fired on load, for one of three reasons:

- the table had a default sort column but the page didn't include it in its first request,
- the table's default page size was 5 while the first request asked for 10,
- the first request had no page size or offset at all, and the second one did.

Filter changes had the same problem, and count requests included the sort order, so every click on a column header refetched the count too.

## What this does

- Adds a `useTableState` hook that is the only owner of page, page size and sort. Everything the API needs (`ordering`, `limit`, `offset`) is derived while rendering, so the page builds its request from the same state in the same render. No more effects writing state back and forth.
- `UniversalRestTable` and `UniversalRestQueryTable` no longer hold state; they take a `tableState` prop. The old `queryVariables` / `setQueryVariables` / `page` / `setPage` / `defaultOrderBy` / `rowsPerPageOptions` props are removed.
- All tables are migrated to the new pattern. Count requests no longer include the sort order.
- Tables that already reset to page 1 when filters change keep doing that (via `resetPageOn`); the others keep their current behaviour.

A few pre-existing issues went away as a side effect: the Users table always requested 10 rows and ignored sorting, the payment record lookup ignored pagination, and the programme cycles table in the payment module refetched manually on every state change.

## Tests

- Unit tests for `useTableState` (sort derivation, direction toggle, page reset on page-size change, page reset on filter change without a stale request).
- `GrievancesTable` test asserting the list is requested exactly once with `ordering=-created_at&limit=10&offset=0` and the count exactly once without sort or pagination. This test fails on `develop`.
- `ProgrammesTable` test asserting a single list request with `limit=5`.


src/frontend/src/containers/tables/targeting/TargetPopulationHouseholdTable/TargetPopulationHouseholdTable.tsx
```diff
@@ -41,19 +41,23 @@ export function TargetPopulationHouseholdTable({
     [variables, businessAreaSlug, programCode, id],
   );
 
-  const [queryVariables, setQueryVariables] = useState(initialQueryVariables);
-  useEffect(() => {
-    setQueryVariables(initialQueryVariables);
-  }, [initialQueryVariables]);
+  const table = useTableState({
+    rowsPerPageOptions: [10, 15, 20],
+    resetPageOn: filterVariables,
```
johniak, line 46: `filterVariables` is rebuilt from the `variables` prop, and `TargetingHouseholds.tsx:13` passes `variables={{ businessArea }}` inline, so every re-render of the details page (it polls every 3s while the TP is building) sends the table back to page 1. Paging to offset 10 and bumping parent state makes the next request go out with offset 0. Memoise `variables` in `TargetingHouseholds`, or deep-compare in `useTableState` (S4).

src/frontend/src/shared/autocompletes/BaseAutocompleteFilterRest.tsx
```diff
@@ -68,7 +68,9 @@ export function BaseAutocompleteFilterRest<TOption = AutocompleteOption>({
     }
   }, [open, debouncedInputText, loadData]);
 
-  if (!data) return null;
+  // Hide only while the first fetch is in flight, so a consumer that defers
+  // fetching until the dropdown opens still renders its (empty) field.
+  if (!data && loading) return null;
```
johniak: With fetching deferred to open, `!data && loading` is true exactly while the first open is in flight, so the whole Autocomplete unmounts: the input disappears, focus is lost, and the popup re-mounts once data lands. MUI already renders the `loading` prop, so this early return can go.

src/frontend/src/shared/autocompletes/AssigneeAutocompleteRestFilter.tsx
```diff
@@ -72,11 +75,12 @@ export function AssigneeAutocompleteRestFilter({
...
   }, [debouncedInputText]);
 
+  // `refetch` ignores `enabled`, so mirror the same guard here.
   const loadData = useCallback(() => {
-    if (businessArea) {
+    if (businessArea && (open || value)) {
       refetch();
     }
-  }, [businessArea, refetch]);
+  }, [businessArea, open, value, refetch]);
```
johniak: Adding `open` and `value` here changes `loadData`'s identity on every open/close/select, and `BaseAutocompleteFilterRest` calls `loadData()` from a `[loadData]` effect as well as the `[open, …]` one. Measured against `develop`: each reopen now fires 2 requests instead of 1, and picking a value fires 1 instead of 0. Same in the other five filters. Since the query already has `enabled` and its key carries `search`, the mount effect in `BaseAutocompleteFilterRest` can be dropped (all six consumers are in this PR), or `loadData` can read `open`/`value` through a ref.

src/frontend/src/hooks/useTableState.ts
```diff
@@ -0,0 +1,111 @@
...
+  const [orderBy, setOrderBy] = useState(defaultOrderBy);
+  const [orderDirection, setOrderDirection] = useState<Order>(
+    defaultOrderDirection,
+  );
+
+  // Reset the page while rendering (not in an effect): React re-renders
+  // immediately with page 0 before committing, so no query goes out with a
+  // stale offset.
+  const [prevResetPageOn, setPrevResetPageOn] = useState(resetPageOn);
+  if (prevResetPageOn !== resetPageOn) {
```
johniak: Identity comparison makes every un-memoised prop feeding a caller's `filterVariables` a page reset on each parent re-render (S1 is one). Before this PR the same churn only re-set query variables to a value-equal key. `isEqual` from lodash here would make the hook robust to that whole class of callers; the object is small.

src/frontend/src/containers/tables/population/HouseholdTable/HouseholdTable.tsx
```diff
@@ -85,38 +82,23 @@ export const HouseholdTable = ({
...
-    filter.orderBy,
-    filter.rdiId,
-  ]);
-  const [queryVariables, setQueryVariables] = useState(initialQueryVariables);
-  useEffect(() => {
-    setQueryVariables(initialQueryVariables);
-  }, [initialQueryVariables]);
+  const table = useTableState({
+    rowsPerPageOptions: [10, 15, 20],
+    defaultOrdering: filter.orderBy,
```
johniak, line 90: `filter.orderBy` moved out of `filterVariables`, so it is no longer part of `resetPageOn`. The removed effect reset the page on `filter.orderBy`; now changing "Sort by" in the filter bar while on page 3 keeps offset 20. Is that intended?

johniak (changes requested):

src/frontend/src/containers/tables/targeting/TargetPopulationHouseholdTable/TargetPopulationHouseholdTable.tsx
mmaciekk: Fixed via the hook: `useTableState` now compares `resetPageOn` with `isEqual`, so the inline `{ businessArea }` from `TargetingHouseholds` no longer resets the page on each poll. Left `TargetingHouseholds` as is since it's value-equal every render. (e0de0f8e6a)

src/frontend/src/shared/autocompletes/BaseAutocompleteFilterRest.tsx
mmaciekk: Removed the early return. The field renders immediately and MUI's `loading` shows the spinner. Only visible difference: a value preselected from the URL shows its id until the label fetch lands, where before the field was hidden for that interval. (e0de0f8e6a)

src/frontend/src/shared/autocompletes/AssigneeAutocompleteRestFilter.tsx
mmaciekk: Went with dropping the mount effect in `BaseAutocompleteFilterRest`; `enabled: open || !!value` covers the URL-preselected case. `loadData` in all six filters is back to its `develop` deps, so it's stable and only called from the `if (open)` effect. Counts now match `develop`: 1 per open, 0 on select, 0 on mount without a value. Added a small test for the base component. (e0de0f8e6a)

src/frontend/src/hooks/useTableState.ts
mmaciekk: Done — `resetPageOn` is now compared with `lodash/isEqual`, and I dropped the 'memoized' wording from the comment. Added a test that a structurally equal new object does not reset the page. (e0de0f8e6a)

src/frontend/src/containers/tables/population/HouseholdTable/HouseholdTable.tsx
mmaciekk: Not intended, thanks. `useTableState` now also resets the page when `defaultOrdering` changes (same rule as `requestSort`), which covers the 'Sort by' dropdown. Test added. (e0de0f8e6a)

src/frontend/src/shared/autocompletes/BaseAutocompleteFilterRest.tsx
mmaciekk: Second thought on this one (de2c1ca30b): dropping the guard entirely meant a value preselected from the URL flashed its raw id in the input until the label fetch landed, which is worse than the old hidden field. So the guard is back as `if (loading && !open) return null`: the field is hidden only while it is closed and resolving that initial label, and it is never hidden while open, so the first open no longer unmounts the input or drops focus. `isLoading` already means "no data yet and fetching" in TanStack v5, so the `!data` part and the `data` prop stay removed. Tests cover both cases.

johniak (approved):

johniak (dismissed):

domdinicola:
@mmaciekk pls fix conflicts

src/frontend/src/components/grievances/GrievancesTable/GrievancesTable.test.tsx
```diff
@@ -127,6 +132,53 @@ describe('GrievancesTable', () => {
...
+    await waitFor(() => {
+      expect(
+        RestService.restBusinessAreasGrievanceTicketsList,
+      ).toHaveBeenCalled();
+      expect(
+        RestService.restBusinessAreasGrievanceTicketsCountRetrieve,
+      ).toHaveBeenCalled();
+    });
+    // Give any stray re-render a chance to fire a second request.
+    await new Promise((resolve) => setTimeout(resolve, 50));
```
nerjan, line 148: if the runner would be slow then 50ms could not be enough and might fail randomly. Would be nice to solve this with some async await for that.

mmaciekk: Good point, fixed. The test now waits until the data has loaded instead of sleeping for 50ms.

nerjan (approved):

johniak (approved):

